PR #22, opened 2016-09-14
Add Operations to label edges in digraphs

markuspf:
This is work in progress to add labels to edges.

My use case will be weighted graphs.

### DIFF
--- a/doc/digraph.xml
+++ b/doc/digraph.xml
@@ -271,6 +271,17 @@ gap> DigraphVertexLabel(gr, 2);
     old ones, inherit their labels from their parents. 
 
     <Example><![CDATA[
+gap> gr := Digraph([[3], [1, 3, 5], [1], [1, 2, 4], [2, 3, 5]]);
+<digraph with 5 vertices, 11 edges>
+gap> DigraphEdgeLabels(gr);
+[ [ 1 ], [ 1, 1, 1 ], [ 1 ], [ 1, 1, 1 ], [ 1, 1, 1 ] ]
+gap> SetDigraphEdgeLabel(gr, 2, 1, "d");
+gap> DigraphVertexLabels(gr);
+[ [ 1 ], [ "d", 1, 1 ], [ 1 ], [ 1, 1, 1 ], [ 1, 1, 1 ] ]
+gap> gr := InducedSubdigraph(gr, [1, 3]);
+<digraph with 2 vertices, 0 edges>
+gap> DigraphVertexLabels(gr);
+[ "a", "c" ]]]
 ]]></Example>
   </Description>
 </ManSection>
@@ -294,7 +305,17 @@ gap> DigraphVertexLabel(gr, 2);
     old ones, inherit their edgelabels from their parents. 
 
     <Example><![CDATA[
-]]></Example>
+gap> gr := Digraph([[3], [1, 3, 5], [1], [1, 2, 4], [2, 3, 5]]);
+<digraph with 5 vertices, 11 edges>
+gap> DigraphEdgeLabel(gr, 3, 1);
+3
+gap> SetDigraphEdgeLabel(gr, 2, 5, [42]);
+gap> DigraphEdgeLabel(gr, 2, 5);
+[ 42 ]
+gap> gr := InducedSubdigraph(gr, [1, 2]);
+<digraph with 2 vertices, 0 edges>
+gap> DigraphVertexLabel(gr, 2);
+"d"]]></Example>
   </Description>
 </ManSection>
 <#/GAPDoc>
--- a/doc/digraph.xml
+++ b/doc/digraph.xml
@@ -252,6 +252,53 @@ gap> DigraphVertexLabel(gr, 2);
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="DigraphEdgeLabels">
+<ManSection>
+  <Oper Name="DigraphEdgeLabels" Arg="digraph"/>
+  <Oper Name="SetDigraphEdgeLabels" Arg="digraph, list"/>
+  <Description>
+    If <A>digraph</A> is a digraph, then <C>DigraphEdgeLabels</C> returns a
+    copy of the labels of the edges in <A>digraph</A>.
+    <C>SetDigraphEdgeLabels</C> can be used to set the labels of the vertices
+    in <A>digraph</A> to the list of lists of arbitrary &GAP; objects
+    <A>list</A>. <P/>
+
+    The label of an edge can be changed an arbitrary number of times. If no
+    label has been set for the edge <A>(i,j)</A>, then the default value is
+    <A>1</A>. <P/>
+
+    Induced subdigraphs, and other operations which create new digraphs from
+    old ones, inherit their labels from their parents. 
+
+    <Example><![CDATA[
+]]></Example>
+  </Description>
+</ManSection>
+<#/GAPDoc>
+
+<#GAPDoc Label="DigraphEdgeLabel">
+<ManSection>
+  <Oper Name = "DigraphEdgeLabel" Arg="digraph, i, j"/>
+  <Oper Name ="SetDigraphEdgeLabel" Arg="digraph, i, j, obj"/>
+  <Description>
+    If <A>digraph</A> is a digraph, then the first operation returns the label
+    of the edge between vertex <A>i</A> and <A>j</A>. The second operation
+    can be used to set the label of the edge between vertex <A>i</A> and <A>j</A>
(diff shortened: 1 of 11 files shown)


markuspf:
This is obviously not ready to merge yet, just basis for discussion.

gap/digraph.gi
```diff
@@ -381,6 +381,81 @@ function(graph)
...
+    else
+        ErrorNoReturn("SetDigraphEdgeLabel: not an edge");
+    fi;
+end);
+
+InstallMethod(SetDigraphEdgeLabels, "for a digraph, and a list",
+[IsDigraph, IsList],
+function(graph, labels)
+    local i;
+    if Length(labels) = Length(graph!.adj) and
```
wilfwilson: `Length(graph!.adj)` should be `DigraphNrVertices(graph)`.

markuspf:
I must have done something daft, I'll fix this later.

james-d-mitchell:
@markuspf looks like a syntax error, which causes Digraphs to not load causing all tests to fail...

markuspf:
Yea, superflous `,`. Let's see what happens now.

markuspf:
Thanks to @wilfwilson for the review. I think I addressed the issues raised, except preserving labels across all operations.

I'll call this merge-ready from my side, but of course subject to comments from @james-d-mitchell or @wilfwilson.

gap/oper.gi
```diff
@@ -1338,6 +1360,7 @@ function(digraph)
   od;
   gr := DigraphNC(new_adj, tot);
   SetDigraphVertexLabels(gr, DigraphVertexLabels(digraph));
+  # Multidigraphs did not have edge labels
```
wilfwilson: It could be the case that `DigraphRemoveAllMultipleEdges` is called on a digraph which has no multiple edges -- in this case we'd probably just want to return the original digraph. But I don't expect you to make that change.

gap/digraph.gi
```diff
@@ -381,6 +381,73 @@ function(graph)
...
+    fi;
+end);
+
+InstallMethod(SetDigraphEdgeLabels, "for a digraph, and a function",
+[IsDigraph, IsFunction],
+function(graph, wtf)
+    local i,j;
+
+    DIGRAPHS_InitEdgeLabels(graph);
+
```
wilfwilson: Over the new three lines, I'd prefer it if you didn't access `graph!.adj` directly, and used `adj := OutNeighbours(graph);` to get access to this information instead.

gap/digraph.gi
```diff
@@ -381,6 +381,73 @@ function(graph)
...
+InstallMethod(DigraphEdgeLabel, "for a digraph, a pos int, and a pos int",
+[IsDigraph, IsPosInt, IsPosInt],
+function(graph, i, j)
+    local p;
+    DIGRAPHS_InitEdgeLabels(graph);
+    p := Position(graph!.adj[i], j);
+    if p <> fail then
+        return ShallowCopy(graph!.edgelabels[i][p]);
+    else
+        ErrorNoReturn("DigraphEdgeLabel: vertex"
```
wilfwilson: The whitespace in this error message is still incorrect - it looks like "DigraphEdgeLabel: vertex[ 3, 4 ]..." (no space between 'vertex' and the last).

doc/digraph.xml
```diff
@@ -252,6 +252,89 @@ gap> DigraphVertexLabel(gr, 2);
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="DigraphEdgeLabels">
```
wilfwilson: This section of documentation should state that you can only read/set the edge labels of a digraph without multiple edges.

doc/digraph.xml
```diff
@@ -252,6 +252,89 @@ gap> DigraphVertexLabel(gr, 2);
...
+gap> DigraphEdgeLabels(gr);
+[ [ 1 ], [ "d", 1 ], [ 1 ] ]
+gap> OutNeighbours(gr);
+[ [ 3 ], [ 1, 3 ], [ 1 ] ]
+]]></Example>
+  </Description>
+</ManSection>
+<#/GAPDoc>
+
+<#GAPDoc Label="DigraphEdgeLabel">
```
wilfwilson: I think this section of documentation should also state that you can only set/read the edge labels of a digraph without multiple edges.

wilfwilson:
@markuspf Just to note, I made a couple of corrections to today's review, so I'd recommend looking at my review in the browser rather than through whatever was sent to your email address.

doc/digraph.xml
```diff
@@ -252,6 +252,89 @@ gap> DigraphVertexLabel(gr, 2);
...
+<#GAPDoc Label="DigraphEdgeLabels">
+<ManSection>
+  <Oper Name="DigraphEdgeLabels" Arg="digraph"/>
+  <Oper Name="SetDigraphEdgeLabels" Arg="digraph, labels"/>
+  <Oper Name="SetDigraphEdgeLabels" Arg="digraph, func"/>
+  <Description>
+    If <A>digraph</A> is a digraph, then <C>DigraphEdgeLabels</C> returns a
+    copy of the labels of the edges in <A>digraph</A> as a list of lists
+    <C>labels</C> such that <C>labels[i][j]</C> is the label on the edge
+    from vertex <A>i</A> to vertex <A>j</A>.
```
markuspf: It seems to me that this is going to lead to problems. Is it maybe desirable to actually use an *edge* (i.e. a pair (i,j)) here, or at least provide the option to do so?

wilfwilson: I'm not sure what change you're proposing.

Which problems will come up, and what do you mean by "using an edge"?

I realise that, with your PR, one is able to set the edge label of `[i,j]` using `SetDigraphEdgeLabel(gr, i, j, value);`, and one can read it by using `DigraphEdgeLabel(gr, i, j);`. I think that's a good idea. But currently, to read/write *all* edge labels at once, you don't do it by explicitly referring to both ends of each edge.

Are you suggesting that `DigraphEdgeLabels` should return a square matrix, so that `labels[i][j]` has the  behaviour you've documentation above? I can't think what else you might mean. One question then is, what value should the matrix have when there is no edge?

markuspf:
There's an awkward problem with the `IsMultiDigraph` split in that one has to spray awkward `if` statements everywhere (for instance in `CopyDigraph` etc), so I'll probably have to move the setting of edge labels into internal functions and only have the front-end error out (or, alternatively, support labels on multi-edges).

wilfwilson:
Alright. It seems difficult to support edge labels for multidigraphs, it's something we thought about a couple of years ago when initially creating the package, and it was tricky - how you distinguish between different edges between the same vertices? It quickly became very fiddly.

I don't mind how you get round the problem of making sure that edge labels aren't accidentally set for multidigraphs. As you suggest, having internal functions to do the setting for `DigraphCopy` (which will fail silently for a multidigraph) seems like a sensible approach.
As long as the user-facing function `SetDigraphEdgeLabels` does have `Errors` when called with a multidigraph.

markuspf:
I think supporting multdigraphs with edge labels would actually require a slightly different graph representation, and ultimately wouldn't give very much advantage.

One can simulate multi-edges by labels (just have a label that represents the different edges).

The representation I would be thinking of is having essentially edges that point to their source and target, and I believe that will also lead to very inefficient algorithms, unless one *also* keeps out-edges for every vertex etc. I don't believe it's worth it.

I'll have a stab at hiding the edge labels stuff in internal functions; that has the advantage taht if one did want to support edge labels at a future point one could try to hack that into the internal functions.

wilfwilson:
We use to have a second internal representation exactly where edges were stored by their source and target. But we scrapped it before release.

Just to be clear, you're not removing the user-facing functions that you've written are you? You're just adding additional internal functions, to be used by methods in the package?

markuspf:
Yeah, I am just adding some functions that silently fail if called on a digraph with multiple edges, not removing any.

markuspf:
I just noticed that my pull request is filed against master, but it seems you're developing on the 0.6 branch, should I try to rebase my work on the 0.6 branch?

james-d-mitchell:
@markuspf Yes, please file your pull request against stable-0.6. 

Also: I agree completely with both of your comments about not supporting edge labels of multidigraphs.

markuspf:
Slightly unrelated question: Do you have a convenient way of getting a display of the test coverage (like codecov.io for example, or just a script to produce @ChrisJefferson's HTML coverage)?

james-d-mitchell:
The latter, I will send you the script on slack

markuspf:
Cheers!

PR #25, opened 2016-10-05
Add IsDirectedTree

le27:
Adds a new function IsDirectedTree to test if a given digraph is a directed tree. (A directed tree is an acyclic digraph with precisely 1 source, such that no two vertices share an out-neighbour.)

### DIFF
--- a/doc/prop.xml
+++ b/doc/prop.xml
@@ -567,3 +567,40 @@ true
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="IsDirectedTree">
+<ManSection>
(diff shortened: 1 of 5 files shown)


mtorpey:
Looks like one of those two test failures is still there, just a digraph being displayed on line 856.

If you try fixing it, go ahead and run `DigraphsTestStandard();` before you commit, to check it's working.

Everything else in this pull request looks good.

le27:
thnx will do :)

mtorpey:
Looks great!  Next time we do this, I'll teach you about "amend" commits so that we don't have a chain of bug-fix commits at the end.  But this looks good for now.  Merging!

Issue #15, opened 2016-02-16
Read DIMACS format


james-d-mitchell:
Originally reported by: **Christopher Jefferson (Bitbucket: [ChrisJefferson](http://bitbucket.org/ChrisJefferson), GitHub: [ChrisJefferson](http://github.com/ChrisJefferson))**

---

Is there any way to read the 'DIMACS' format in Digraphs? I couldn't find it.

---
- Bitbucket: https://bitbucket.org/james-d-mitchell/digraphs/issue/15

mtorpey:
This is merged and done!  I've added a Trello card for dreadnaut format.

PR #26, opened 2016-10-12
Add support for writing and reading DIMACS files

wilfwilson:
The purpose of this PR is to address Issue #15 

I still need to add documentation and tests.

### DIFF
--- a/gap/io.gi
+++ b/gap/io.gi
@@ -1157,7 +1157,7 @@ function(name)
 
   if not nr_edges in [directed_edges, 2 * directed_edges, symmetric_edges] then
     Info(InfoDigraphs, 1,
-         "A different number of edges than expected was found,");
+         "An unexpected number of edges was found,");
   fi;
 
   IO_Close(file);
@@ -1181,7 +1181,7 @@ function(name, digraph)
   file := IO_CompressedFile(UserHomeExpand(name), "w");
   if file = fail then
     ErrorNoReturn("Digraphs: WriteDIMACSDigraph:\n",
-                  "can not open the file ", name, ",");
+                  "cannot open the file ", name, ",");
   fi;
 
   n := DigraphNrVertices(digraph);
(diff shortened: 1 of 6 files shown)


wilfwilson:
I've added documentation and tests now, and finalised the code. It's ready to go as far as I'm concerned.

gap/io.gi
```diff
@@ -1039,6 +1039,201 @@ function(arg)
...
+    first_char := next[1];
+
+    # digraph definition line
+    if first_char = 'p' then
+      if IsBound(vertices) or Length(split) <> 4 or split[2] <> "edge" then
+        malformed_file();
+      fi;
+      nr_vertices     := int_from_string(split[3]);
+      vertices        := [1 .. nr_vertices];
+      vertex_labels   := vertices * 0 + 1;
```
mtorpey: Is this the fastest way of doing this?   Perhaps `ListWithIdenticalEntries(nr_vertices, 1)` would be better.

gap/io.gi
```diff
@@ -1039,6 +1039,201 @@ function(arg)
...
+    next := IO_ReadLine(file);
+  od;
+
+  if not IsBound(vertices) then
+    malformed_file();
+  fi;
+
+  if not nr_edges in [directed_edges, 2 * directed_edges, symmetric_edges] then
+    Info(InfoDigraphs, 1,
+         "A different number of edges than expected was found,");
```
mtorpey: Grammar nazis prefer "different from" to "different than".

gap/io.gi
```diff
@@ -1039,6 +1039,201 @@ function(arg)
...
+
+  if not IsSymmetricDigraph(digraph) then
+    ErrorNoReturn("Digraphs: WriteDIMACSDigraph: usage,\n",
+                  "the digraph <digraph> must be symmetric,");
+  fi;
+
+  file := IO_CompressedFile(UserHomeExpand(name), "w");
+  if file = fail then
+    ErrorNoReturn("Digraphs: WriteDIMACSDigraph:\n",
+                  "can not open the file ", name, ",");
```
mtorpey: `cannot` instead of `can not`

gap/io.gi
wilfwilson: I believe that would be no faster.

james-d-mitchell: Yup, I think Wilf's method is fastest, and more succinct, if a bit cryptic.

PR #28, opened 2016-10-19
CompleteMultipartiteDigraph

stuartburrell:
This short pull request includes:

- Implementation and declaration for CompleteMultibipartiteDigraph
- Tests for CompleteMultibipartiteDigraph
- Documentation for CompleteMultibipartiteDigraph


mtorpey:
Hi Stuart!

I'm gonna look at this and hopefully merge it today (since you were so good about merging my pull request).

But just so you're aware: since this pull request is from `stuartburrell:stable-0.6`, anything you push to that branch will appear on here.  I'm not sure whether you intended the last couple of commits to be included - you certainly don't need to do any merging, since github does that automatically for pull requests.

stuartburrell:
Hi, awesome thanks!

That funny business at the end was me moving my changes to the stable-0.6 branch on my system 
as I had to merge all the changes I was doing on my master into it.

mtorpey:
There's some good stuff in here, but a few things to address before we should merge this:

* The word should be "multipartite", not "multibipartite".  "bi" indicates a split into two parts.  See e.g. [here](https://en.wikipedia.org/wiki/Multipartite_graph).
* Why do we require `n >= 2`?  I see nothing wrong with having a single segment, where `CompleteMultipartiteDigraph([i])` is a graph with `i` vertices and no edges.  Or even for `n=0`, where `CompleteMultipartiteDigraph([])` is a graph with no vertices or edges.
* When you check `p_size >= 1`, you should instead check `IsPosInt(p_size)`.  This ensures that the user has supplied integers and not floats or strings or anything weird.
* We have a standard format for reporting errors which you can see throughout the package's `.gi` files - here's an example which you might try to mimic:
```
ErrorNoReturn("Digraphs: DigraphRemoveEdges: usage,\n",
              "the second argument <edges> must be a list of indices of\n",
              "edges or a list of edges of the first argument <digraph>,");
```
* In line 1514, `[ 2.. Length` should read `[2 .. Length`.  Obviously this is pretty minor.
* Maybe include a couple of extra tests that actually check whether a few edges are present, or print out the OutNeighbours for some of the smaller graphs.  At the moment we're only checking that they have the correct number of edges.
* In the documentation, the args list shows an optional argument called `n`.  Obviously the function at the moment only takes a single argument, so that should be removed.
* The description is a bit clunky.  I'd use a term like "independent sets" instead of "parts", and make it totally clear that a vertex has an edge to *every* vertex outside its set and *no* vertex inside its set.

These are all fairly minor problems, and the function itself seems to be correct!  If you fix those things and push again, I'll be happy to merge.  I'll do this using "squash and merge", so don't worry about having loads of ugly commits.  :smiley:

stuartburrell:
Thanks so much for all the helpful comments!

Think I've covered them with the new commits.

mtorpey:
Looking pretty good!  Just two things left that I can see.

* In the code in line 1500, the error message still says "multibipartite", and of course the tests reflect this.
* In the doc, we still have `Arg = "orders[ ]"`.  Those square brackets aren't needed - just `Arg = "orders"` is fine.

And after that I'll finally be happy to merge!

stuartburrell:
There we go! Thanks

mtorpey:
I've merged it (squashing all the commits into one).

Since I did a squash merge, note that your `stable-0.6` history will differ from the one on the server.  You can fix this by doing
```
git fetch origin
git checkout stable-0.6
git reset --hard origin/stable-0.6
```

PR #35, opened 2016-12-07
Convert ReadPackage single argument calls to two arguments

fingolfin:


wilfwilson:
Thanks for this @fingolfin - we'll be happy to merge this, but we're wondering why this change is necessary?

Issue #23, opened 2016-09-14
Improve clique finding for digraphs with isolated vertices


wilfwilson:
For a digraph with isolated vertices, the automorphism group tends to be large - since the isolated vertices can be permuted at will. Therefore calculations involving the automorphism group of such digraphs may take longer. However, for cliques, an isolated vertex is only ever in a trivial (maximal) clique - so we really don't need to compute with these at all. (And they're probably not relevant to other computations either). Therefore the clique finder should pre-process a digraph to remove its isolated vertices using `ReducedDigraph`.

wilfwilson:
Resolved by PR #37

PR #37, opened 2016-12-07
Fix Issue #23: DigraphMaximalCliques

wilfwilson:
`DigraphMaximalCliques` and `DigraphMaximalCliquesReps` used to be slow when the digraph in question had lots of isolated vertices. This is because the subgroup of the automorphism group which acts on these vertices is the symmetric group, which is really big. Therefore calculations involving the automorphism group were taking ages. All of the time was spent calculating stabilizer chains, etc.

Now, in certain cases when it is safe to do so, we (in essence) remove the isolated vertices (except one), and remove the generators of the automorphism group which act on the isolated vertices. This speeds things up.


wilfwilson:
I'm going to merge this when the tests pass.

PR #44, opened 2017-02-01
Is hamiltonian digraph

le27:
Adds a function called IsHamiltonianDigraph which checks if a given digraph is hamiltonian

### DIFF
--- a/doc/prop.xml
+++ b/doc/prop.xml
@@ -677,3 +677,36 @@ true
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="IsHamiltonianDigraph">
+<ManSection>
+  <Prop Name="IsHamiltonianDigraph" Arg="digraph"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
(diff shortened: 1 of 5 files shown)


james-d-mitchell:
Julius agreed to look at this too, and squash-merge if he's happy.

ChrisJefferson:
I'm just jumping in as a separate, interested reader.

I feel you should mention in the docs that uses `O(DigraphNrVertices(g)!)` space and time.

Also, while any algorithm for hamiltonian digraph will use `O(DigraphNrVertices(g)!)`, you can do quite a bit better than this is you want to -- depending on how hard you want to try. It would be nice at least to switch to only using `O(DigraphNrVertices(g))` space by doing a depth-first search (which could also terminate early as soon as it found a path).

james-d-mitchell:
Thanks @chrisjefferson, it'd be good if we could improve the complexity as Chris indicates.

wilfwilson:
Why was this closed?

james-d-mitchell:
I think because @le27 opened a new pull request with the same functionality.

PR #48, opened 2017-03-03
Introduce methods for undirected spanning trees and forests

wilfwilson:
The package already includes the property `IsUndirectedTree`, which decides whether a digraph is a tree in the usual graph-theoretic sense (symmetric, connected, no cycles).

This PR adds several new features related to undirected trees:

* property`IsUndirectedForest`: an undirected forest is a disjoint union of undirected trees.
* attributes `UndirectedSpanningTree` and `UndirectedSpanningForest`: spanning tree has the usual sense, and a spanning forest is a subdigraph that is a forest and is maximal (with respect to containment of subdigraphs) with this property. This later attribute is the most important one for my purposes. Every digraph has a spanning forest (which may be an empty digraph); not every digraph has a spanning tree.
* operations `IsUndirectedSpanningTree` and `IsUndirectedSpanningForest`.

I will use the attribute `UndirectedSpanningForest` in the Semigroups package. In particular, I'll add a method for `IdempotentGeneratedSubsemigroup` for a Rees 0-matrix semigroup over a group, which will be able to return a semigroup with a much smaller generating set than before. This involves calculating a spanning forest of the `RZMSDigraph`.

### DIFF
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -1211,3 +1211,58 @@ gap> OutNeighbours(sym);
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="UndirectedSpanningTree">
+<ManSection>
+  <Attr Name="UndirectedSpanningTree" Arg="digraph"/>
+  <Attr Name="UndirectedSpanningForest" Arg="digraph"/>
+  <Returns>A digraph, or <K>fail</K>.</Returns>
+  <Description>
+    If <A>digraph</A> is a digraph with at least one vertex, then
+    <C>UndirectedSpanningForest</C> returns an undirected spanning forest of
+    <A>digraph</A>, otherwise this attribute returns <K>fail</K>. See <Ref
+      Oper="IsUndirectedSpanningForest" /> for the definition of an undirected
+    spanning forest. <P/>
+
+    If <A>digraph</A> is a digraph with at least one vertex and whose <Ref
+      Attr="MaximalSymmetricSubdigraph"/> is connected (see <Ref
+      Prop="IsConnectedDigraph" />), then <C>UndirectedSpanningTree</C> returns
+    an undirected spanning tree of <A>digraph</A>, otherwise this attribute
+    returns <K>fail</K>.  See <Ref Oper="IsUndirectedSpanningTree" /> for the
+    definition of an undirected spanning tree. <P/>
+
+    Note that for a digraph that has an undirected spanning tree, the attribute
+    <C>UndirectedSpanningTree</C> returns the same digraph as the attribute
+    <C>UndirectedSpanningForest</C>.
+
+    <Example><![CDATA[
+gap> gr := Digraph([[1, 2, 1, 3], [1], [4], [3, 4, 3]]);
+<multidigraph with 4 vertices, 9 edges>
+gap> UndirectedSpanningTree(gr);
+fail
+gap> forest := UndirectedSpanningForest(gr);
+<digraph with 4 vertices, 4 edges>
+gap> OutNeighbours(forest);
+[ [ 2 ], [ 1 ], [ 4 ], [ 3 ] ]
+gap> IsUndirectedSpanningForest(gr, forest);
+true
+gap> DigraphConnectedComponents(forest).comps;
+[ [ 1, 2 ], [ 3, 4 ] ]
+gap> DigraphConnectedComponents(MaximalSymmetricSubdigraph(gr)).comps;
+[ [ 1, 2 ], [ 3, 4 ] ]
+gap> UndirectedSpanningForest(MaximalSymmetricSubdigraph(gr))
+> = forest;
+true
+gap> gr := CompleteDigraph(4);
+<digraph with 4 vertices, 12 edges>
+gap> tree := UndirectedSpanningTree(gr);
+<digraph with 4 vertices, 6 edges>
+gap> IsUndirectedSpanningTree(gr, tree);
+true
+gap> tree = UndirectedSpanningForest(gr);
+true
+gap> UndirectedSpanningForest(EmptyDigraph(0));
+fail]]></Example>
+  </Description>
+</ManSection>
+<#/GAPDoc>
--- a/gap/attr.gi
+++ b/gap/attr.gi
@@ -1,7 +1,7 @@
 #############################################################################
 ##
 #W  attr.gi
-#Y  Copyright (C) 2014                                   James D. Mitchell
+#Y  Copyright (C) 2014-17                                James D. Mitchell
 ##
 ##  Licensing information can be found in the README file of this package.
 ##
@@ -1347,4 +1347,39 @@ function(gr, loops)
   SetIsSymmetricDigraph(new_gr, true);
   SetDigraphVertexLabels(new_gr, DigraphVertexLabels(gr));
   return new_gr;
+end);
+
+#
+
+InstallMethod(UndirectedSpanningForest,
+"for a digraph",
+[IsDigraph],
+function(gr)
+  local out;
+  if DigraphNrVertices(gr) = 0 then
+    return fail;
+  fi;
+
+  gr := MaximalSymmetricSubdigraph(gr);
+  out := Digraph(DIGRAPH_SYMMETRIC_SPANNING_FOREST(OutNeighbours(gr)));
+  SetIsUndirectedForest(out, true);
+  SetIsMultiDigraph(out, false);
+  SetDigraphHasLoops(out, false);
+  return out;
+end);
+
+#
+
+InstallMethod(UndirectedSpanningTree,
+"for a digraph",
+[IsDigraph],
+function(gr)
+  local out;
+  if DigraphNrVertices(gr) = 0
+      or not IsStronglyConnectedDigraph(MaximalSymmetricSubdigraph(gr)) then
+    return fail;
+  fi;
+  out := UndirectedSpanningForest(gr);
+  SetIsUndirectedTree(out, true);
+  return out;
 end);
--- a/gap/oper.gi
+++ b/gap/oper.gi
@@ -1,7 +1,7 @@
 #############################################################################
 ##
 #W  oper.gi
-#Y  Copyright (C) 2014                                   James D. Mitchell
+#Y  Copyright (C) 2014-17                                James D. Mitchell
 ##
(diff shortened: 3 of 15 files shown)


doc/attr.xml
```diff
@@ -1211,3 +1211,60 @@ gap> OutNeighbours(sym);
...
+    returns <K>fail</K>.  See <Ref Oper="IsUndirectedSpanningTree" /> for the
+    definition of an undirected spanning tree.
+
+    <P/>
+
+    Note that for a digraph that has an undirected spanning tree, the attribute
+    <C>UndirectedSpanningTree</C> returns the same digraph as does the attribute
+    <C>UndirectedSpanningForest</C>.
+
+    <P/>
```
james-d-mitchell: This `<P/>` is superfluous, no?

doc/attr.xml
```diff
@@ -1211,3 +1211,60 @@ gap> OutNeighbours(sym);
...
+  <Attr Name="UndirectedSpanningTree" Arg="digraph"/>
+  <Attr Name="UndirectedSpanningForest" Arg="digraph"/>
+  <Returns>A digraph, or <K>fail</K>.</Returns>
+  <Description>
+    If <A>digraph</A> is a digraph with at least one vertex, then
+    <C>UndirectedSpanningForest</C> returns an undirected spanning forest of
+    <A>digraph</A>, otherwise this attribute returns <K>fail</K>. See <Ref
+      Oper="IsUndirectedSpanningForest" /> for the definition of an undirected
+    spanning forest.
+
```
james-d-mitchell: Also why are there line breaks before all the `<P/>` here?

doc/attr.xml
```diff
@@ -1211,3 +1211,60 @@ gap> OutNeighbours(sym);
...
+      Attr="MaximalSymmetricSubdigraph"/> is connected (see <Ref
+      Prop="IsConnectedDigraph" />), then <C>UndirectedSpanningTree</C> returns
+    an undirected spanning tree of <A>digraph</A>, otherwise this attribute
+    returns <K>fail</K>.  See <Ref Oper="IsUndirectedSpanningTree" /> for the
+    definition of an undirected spanning tree.
+
+    <P/>
+
+    Note that for a digraph that has an undirected spanning tree, the attribute
+    <C>UndirectedSpanningTree</C> returns the same digraph as does the attribute
```
james-d-mitchell: I'd remove the word "does" from this line.

doc/attr.xml
```diff
@@ -1211,3 +1211,60 @@ gap> OutNeighbours(sym);
...
+gap> OutNeighbours(forest);
+[ [ 2 ], [ 1 ], [ 4 ], [ 3 ] ]
+gap> IsUndirectedSpanningForest(gr, forest);
+true
+gap> DigraphConnectedComponents(forest).comps;
+[ [ 1, 2 ], [ 3, 4 ] ]
+gap> DigraphConnectedComponents(MaximalSymmetricSubdigraph(gr)).comps;
+[ [ 1, 2 ], [ 3, 4 ] ]
+gap> UndirectedSpanningTree(gr);
+fail
```
james-d-mitchell: How about adding `UndirectedSpanningForest(gr)` here?

gap/attr.gi
```diff
@@ -1348,3 +1348,38 @@ function(gr, loops)
...
+end);
+
+#
+
+InstallMethod(UndirectedSpanningTree,
+"for a digraph",
+[IsDigraph],
+function(gr)
+  local out;
+  if DigraphNrVertices(gr) = 0 or
```
james-d-mitchell: `or` should be on the next line

gap/oper.gi
```diff
@@ -50,6 +50,42 @@ end);
...
+
+  sym := MaximalSymmetricSubdigraph(super);
+  if not IsUndirectedForest(sub) or not IsSubdigraph(sym, sub) then
+    return false;
+  fi;
+
+  comps1 := List(DigraphConnectedComponents(sym).comps, ShallowCopy);
+  comps2 := List(DigraphConnectedComponents(sub).comps, ShallowCopy);
+
+  if Length(comps1) <> Length(comps2) then
```
james-d-mitchell: Why copy `comps1` and `comps2` just to return `false` in the next line?

gap/oper.gi
```diff
@@ -50,6 +50,42 @@ end);
 
 #
 
+InstallMethod(IsUndirectedSpanningForest, "for a digraph and a digraph",
+[IsDigraph, IsDigraph],
+function(super, sub)
+  local sym, comps1, comps2;
+
+  sym := MaximalSymmetricSubdigraph(super);
```
james-d-mitchell: Shouldn't this be calculated only if `IsUndirectedForest(sub)`?

gap/prop.gi
```diff
@@ -395,6 +395,24 @@ function(gr)
...
 
+InstallMethod(IsUndirectedForest, "for a digraph", [IsDigraph],
+function(gr)
+  local comps, comp;
+
+  if not IsSymmetricDigraph(gr) or DigraphNrVertices(gr) = 0
+      or IsMultiDigraph(gr) then
+    return false;
+  fi;
+  comps := DigraphConnectedComponents(gr);
```
james-d-mitchell: Better do the `comps := DigraphConnectedComponents(gr).comps;`

james-d-mitchell: in this line and not in the `for` loop.

doc/attr.xml
wilfwilson: The way I think of it, the `<P/>` tags marks the boundary between paragraphs. So I like the style of putting it clearly between the preceding and proceeding paragraphs. But I'll happily change it.

doc/attr.xml
wilfwilson: Yes, it's redundant in the sense that it doesn't affect the output. But semantically I put it there since it still marks the end of the preceding paragraph. I'll remove it.

doc/attr.xml
wilfwilson: The purpose of this line is to show that a digraph does not have a spanning tree if its underlying symmetric "graph" is not connected. We've already calculated variable `forest := UndirectedSpanningForest(gr)` earlier in the example. I'll add something similar.

gap/oper.gi
```diff
@@ -50,6 +50,49 @@ end);
...
+  comps1 := DigraphConnectedComponents(sym).comps;
+  comps2 := DigraphConnectedComponents(sub).comps;
+
+  if Length(comps1) <> Length(comps2) then
+    return false;
+  fi;
+
+  comps1 := List(comps1, ShallowCopy);
+  comps2 := List(comps2, ShallowCopy);
+  Perform(comps1, Sort);
```
james-d-mitchell: I think these 6 lines could be two: 

`comps1 := Set(comps1, SortedList);`

but I didn't test this.

src/digraphs.c
```diff
@@ -854,6 +854,85 @@ static Obj FuncDIGRAPH_TOPO_SORT(Obj self, Obj adj) {
...
+          stack[0] = next;
+          stack[1] = 1;
+          // if <next> is a brand new vertex, add it to the spanning tree
+          if (ptr[next] == 0) {
+            // Create the forward edge, j -> next
+            out_j = ELM_PLIST(out, j);
+            len   = LEN_PLIST(out_j);
+            if (len == 0) {
+              RetypeBag(out_j, T_PLIST_CYC + IMMUTABLE);
+              CHANGED_BAG(out_j);
```
james-d-mitchell: I think you need a `CHANGED_BAG(out)` here too.

src/digraphs.c
```diff
@@ -854,6 +854,85 @@ static Obj FuncDIGRAPH_TOPO_SORT(Obj self, Obj adj) {
...
+            }
+            AssPlist(out_j, len + 1, INTOBJ_INT(next));
+            // Create the reverse edge, next -> j
+            // Since <next> is new, new = out[next] is still a T_PLIST_EMPTY
+            new = ELM_PLIST(out, next);
+            RetypeBag(new, T_PLIST_CYC + IMMUTABLE);
+            CHANGED_BAG(new);
+            AssPlist(new, 1, INTOBJ_INT(j));
```
james-d-mitchell: And here another `CHANGED_BAG(out)`

src/digraphs.c
```diff
@@ -854,6 +854,85 @@ static Obj FuncDIGRAPH_TOPO_SORT(Obj self, Obj adj) {
   return out;
 }
 
+// TODO(*) use generic DFS, when we have one.
+
```
james-d-mitchell: Please also add some explanation of what you are doing here, can be a sentence or two.

gap/oper.gi
wilfwilson: Much better solution.

src/digraphs.c
wilfwilson: Going by the other functions in `digraphs.c`, it seems like `CHANGED_BAG(out_j)` isn't  what's needed here (presumably the function `RetypeBag` does everything necessary with respect to `out_j`), but that it should instead *only* be the `CHANGED_BAG(out)` that you suggest. Does this sound right to you? If not, and we need both, then modifications are needed to some of the other functions in `digraphs.c`.

src/digraphs.c
wilfwilson: Going by my previous comment, the `CHANGED_BAG(new)` should be *replaced* by the `CHANGED_BAG(out)`. In this case, is it still necessary to have the `CHANGED_BAG` within the `if` statement a few lines above, since they're so close together?

wilfwilson: (I don't know much about this stuff)

src/digraphs.c
james-d-mitchell: Sounds right

src/digraphs.c
james-d-mitchell: I'd put the `CHANGED_BAG(out)` after the `AssPlist(new, 1, INTOBJ_INT(j))`, and it's not the proximity that matters it's whether or not there can be a garbage collection between the statements.

james-d-mitchell: Also not having `CHANGED_BAG` is bad (if it is required), having it there unnecessarily does no harm.

wilfwilson:
Thanks for reviewing this @james-d-mitchell. I've made all the changes that you suggested. Good job spotting the stuff about the `CHANGED_BAG`s.

Issue #19, opened 2016-06-06
Digraph canonical labelling not respecting colours


james-d-mitchell:
Originally reported by: **Finn Smith (Bitbucket: [flsmith](http://bitbucket.org/flsmith), GitHub: [flsmith](http://github.com/flsmith))**

---

I might have misunderstood the point behind this function, but I'm confused by the example that is given in the manual

```
#!GAP

gap> G := Digraph(10, [1, 1, 3, 4, 4, 5, 8, 8], [6, 3, 3, 9, 10, 9, 4, 10]);
<digraph with 10 vertices, 8 edges>

gap> DigraphCanonicalLabelling(G, [1, 1, 1, 1, 2, 3, 1, 3, 2, 1]);
(1,4,5,7)(3,6,9,8,10)
```

I can't see in what sense this result respects colouring.

In the different encoding of coloured digraphs I have, I get a different result... which also seems not to respect colouring.

---
- Bitbucket: https://bitbucket.org/james-d-mitchell/digraphs/issue/19

wilfwilson:
I've convinced myself, and talked to @flsmith about it, that `DigraphCanonicalLabelling` is working correctly with respect to colours. However, I think that Finn is right to be confused by the documentation. I'll improve the documentation and then close this issue.

wilfwilson:
The documentation essentially gave the impression that `DigraphCanonicalLabelling` for a coloured digraph returns a permutation that fixes the colouring.

wilfwilson:
This was sorted out by #49, and released in `v0.7.0`.

Issue #40, opened 2017-01-23
DigraphGroup doesn't record group Size


ChrisJefferson:
DigraphGroup should be able to extract from bliss the size of the found group, and attach it to the group. This information would be very useful, as I have a group where DigraphGroup took 3 seconds, but Size is still not finished, and has taken over 10 minutes.

james-d-mitchell:
This is a good idea. 

@ChrisJefferson Do you have any idea how to get the size from bliss? On very quick inspection, I can't seem to find a method which returns the size, and in `defs.hh` it seems to suggest that `gmp` is required to do this.

ChrisJefferson:
@saadattieh 's alternative implementation is now complete. The main useful features it will bring are:

1) Canonical image works properly with partitions (fixes #19)
2) DigraphGroup will return a group with the known Size, and also a valid stabilizer chain (if wanted.. I suspect it will be wanted)
3) Faster, and with (optional) parallelisation.

I suggest to start with we plug both systems in (partly for comparision), then later bliss can be removed if the replacement is always better. We could also just have a "big bang" were bliss is replaced.

james-d-mitchell:
This sounds really good. I'm happy with either a big bang or a both systems, the latter might be a bit more complicated than necessary. We can always compare the development version with the release version. Do you want one of us to do this or will you and @saadattieh?

wilfwilson:
@ChrisJefferson What's the status/location of @SaadAttieh's implementation? Do you still recommend integrating it into Digraphs?

james-d-mitchell:
I was wondering about this recently too.

ChrisJefferson:
Hi!

Sorry, going back to an old, old PR.

As this is a useful feature, and hungry seems to have gotten stuck on some subtle bugs we are probably never going to fix, I've decided to just add this to Bliss instead. I have hacked up the feature in bliss, it would be useful to have someone to work with me on the digraphs side of things.

PR #49, opened 2017-03-10
Improve user-facing isomorphism and canonical labelling functions, especially with colourings

wilfwilson:
This PR adds support for isomorphism testing, and isomorphism finding, for *coloured* digraphs.

---

Previously in the Digraphs package... you could ask for the automorphism group of a coloured digraph without multiple edges (i.e. the subgroup of the automorphism of the digraph consisting of those elements that preserve the colouring), and you could ask for the canonical labelling of a coloured digraph without multiple edges.

I've now added similar support for coloured digraphs to the operations `IsIsomorphicDigraph` and `IsomorphismDigraphs`, for testing whether two coloured digraphs without multiple edges are isomorphic as coloured digraphs (i.e. whether there exists an isomorphism between the digraphs that respects the colourings of the vertices).

In addition to these new features, I've made minor improvements to the existing code, such as checking in `IsIsomorphicDigraph` whether the two digraphs are equal before going on to compute a canonical labelling.

The main change of this type is a new global function `DIGRAPHS_ValidateVertexPartition`. Creating this function allows me to replace a large amount of duplicated code which existed to validate whether the "colouring" given as an argument to `AutomorphismGroup` or `DigraphCanonicalLabelling` was indeed a valid colouring of the relevant digraph. This is now also used by `IsIsomorphicDigraph` and `IsomorphismDigraphs`.

Given a non-negative integer `n` and a homogeneous list `partition`, this `DIGRAPHS_ValidateVertexPartition` tests whether `partition` is a valid partition `[1 .. n]`. A valid partition of `[1 .. n]` is either:
1. a list of length `n` consisting of numbers in the range `[0 .. n]` (so that `partition[i]` is the colour of vertex `i`), or
2. a list of non-empty disjoint lists whose union is `[1 .. n]` (so that `partition[i]` is the list of those vertices with colour `i`).

If `partition` *is* a valid partition of `[1 .. n]` then it returns the partition in form 1 (this is the form that bliss likes); otherwise it returns `fail`.

This is both a more and less permissive definition of colouring than existed previously.

It's more permissive, since in the first form, I now allow the colour `0` to be used. This is because:

* bliss works fine with any colour in the range `[0 .. n]` (but not outside of this range), and
* in the Semigroups package, a method for `CanonicalBooleanMat` calculates the canonical labelling of a particular colouring digraph which uses colours 0 and 1. Since colour 0 is not currently allowed in Digraphs, this method uses `DIGRAPH_CANONICAL_LABELING_COLORS`. I think it would be better if this method used the operation `DigraphCanonicalLabelling` for a digraph and a colouring.

It's less permissive, since in the second form, I now require the list to consist of *disjoint* lists. This was currently allowed, but to my mind that was a mistake, as it doesn't make sense for a vertex to have two colours.

In time, this global function should also replace the validation of the colouring given to `HomomorphismDigraphsFinder`.

---

However, the main work on this PR is an improvement to documentation all of the functions `AutomorphismGroup`, `DigraphCanonicalLabelling`, `IsIsomorphicDigraph`, and `IsomorphismDigraphs` in the file `bliss.*`.

The previous documentation for these functions was either incorrect, incomplete, or confusing (this confusion prompted issue #19). It certainly left me scratching my head for a while. I've tried to make the documentation as clear, as thorough, and as helpful as possible.

---

As an aside, since in the `bliss.xml` documentation we talk a lot about colourings, where we simply mean a labelling, I now emphasise in the documentation for `DigraphColouring` and `ChromaticNumber` that these colourings are special in some sense -- they are *proper* colourings.

### DIFF
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -13,17 +13,19 @@
   <Attr Name="ChromaticNumber" Arg="digraph"/>
   <Returns> A non-negative integer.</Returns>
   <Description>
-    A digraph coloring is a labeling of the vertices (using precisely <A>n</A>
-    colors) in such a way that two adjacent vertices can not have the same
-    label.  Alternatively, it can be defined to be a <Ref
-      Oper="DigraphEpimorphism"/> from <A>digraph</A> onto a complete digraph
-    with <A>n</A> vertices.<P/>
+    A <E>proper colouring</E> of a digraph is a labelling of its
+    vertices in such a way that adjacent vertices have different labels.
+    Equivalently, a proper digraph colouring can be defined to be a <Ref
+      Oper="DigraphEpimorphism"/> from a digraph onto a complete digraph. <P/>
 
-    <C>ChromaticNumber</C> returns the least non-negative integer <C>n</C> such
-    that there is a coloring of the loopless digraph <A>digraph</A> with
-    <C>n</C> colors.  In other words, <C>ChromaticNumber</C> returns the least
-    value <C>n</C> such that <C>DigraphColoring(<A>digraph</A>, n)</C> does not
-    return <K>fail</K>.
+    If <A>digraph</A> is a digraph without loops (see <Ref
+      Prop="DigraphHasLoops"/>, then <C>ChromaticNumber</C> returns the least
+    non-negative integer <C>n</C> such that there is a proper colouring of
+    <A>digraph</A> with <C>n</C> colours.  In other words, for a digraph with at
+    least one vertex, <C>ChromaticNumber</C> returns the least number <C>n</C>
+    such that <C>DigraphColouring(<A>digraph</A>, n)</C> does not return
+    <K>fail</K>. See <Ref Oper="DigraphColouring"
+      Label="for a digraph and a number of colours"/>.
 
     <Example><![CDATA[
 gap> ChromaticNumber(NullDigraph(10));
--- a/doc/bliss.xml
+++ b/doc/bliss.xml
@@ -2,6 +2,7 @@
 ##
 #W  bliss.xml
 #Y  Copyright (C) 2014-17                               James D. Mitchell
+##                                                         Wilf A. Wilson
 ##
 ##  Licensing information can be found in the README file of this package.
 ##
@@ -11,83 +12,59 @@
 <#GAPDoc Label="AutomorphismGroupDigraph">
 <ManSection>
   <Attr Name="AutomorphismGroup" Label="for a digraph" Arg="digraph"/>
-  <Oper Name="AutomorphismGroup" Label="for a digraph and a homogeneous list"
-    Arg="digraph, colors"/>
   <Returns>A permutation group.</Returns>
   <Description>
-    If <A>digraph</A> is a digraph without multiple edges, then this function
-    returns the automorphism group of <A>digraph</A>, as a group of
-    permutations on the vertices of <A>digraph</A>. <P/>
+    If <A>digraph</A> is a digraph, then this attribute contains the group of
+    automorphisms of <A>digraph</A>.  An <E>automorphism</E> of <A>digraph</A>
+    is an isomorphism from <A>digraph</A> to itself. See <Ref
+      Oper="IsomorphismDigraphs" Label="for digraphs" /> for more information
+    about isomorphisms of digraphs. <P/>
 
-    If the <A>colors</A> argument is specified, then the group will consist of
-    only those automorphisms which respect the given colouring.  The colouring
-    <A>colors</A> can be in one of two forms: <P/>
+    The form in which the automorphism group is returned depends on whether
+    <A>digraph</A> has multiple edges; see <Ref Prop="IsMultiDigraph" />. <P/>
 
     <List>
+      <Mark>for a digraph without multiple edges</Mark>
       <Item>
-        A list of positive integers of size the number of vertices of
-        <A>digraph</A>, where <A>colors</A><C>[i]</C> is the colour of vertex
-        <C>i</C>.
+        If <A>digraph</A> has no multiple edges, then the automorphism group is
+        returned as a group of permutations on the vertices of <A>digraph</A>.
       </Item>
+      <Mark>for a multidigraph</Mark>
       <Item>
-        A list of lists, such that <A>colors</A><C>[i]</C> is a list of all
-        vertices with colour <C>i</C>.
+        If <A>digraph</A> is a multidigraph, then the automorphism group is
+        a group of permutations on the vertices and edges of <A>digraph</A>.
+        <P/>
+
+        For convenience, the group is returned as the direct product <C>G</C> of
+        the group of automorphisms of the vertices of <A>digraph</A> with the
+        stabiliser of the vertices in the automorphism group of the edges. These
+        two groups can be accessed using the operation <Ref Oper="Projection"
+          Label="for a domain and a positive integer" BookName="ref"/>, with the
+        second argument being <C>1</C> or <C>2</C>, respectively. <P/>
+
+        The permutations in the group <C>Projection(G, 1)</C> act on the
+        vertices of <A>digraph</A>, and the permutations in the group
+        <C>Projection(G, 2)</C> act on the indices of
+        <C>DigraphEdges(<A>digraph</A>)</C>.
       </Item>
     </List>
 
-    The automorphism group is found using <URL
-      Text="bliss">http://www.tcs.tkk.fi/Software/bliss/index.html</URL> by
-    Tommi Junttila and Petteri Kaski.
+    The automorphism group is found using &bliss; by Tommi Junttila and Petteri
+    Kaski.
 
     <Example><![CDATA[
 gap> johnson := DigraphFromGraph6String("E}lw");
 <digraph with 6 vertices, 24 edges>
-gap> AutomorphismGroup(johnson);
+gap> G := AutomorphismGroup(johnson);
 Group([ (3,4), (2,3)(4,5), (1,2)(5,6) ])
-gap> Size(last);
-48
+gap> StructureDescription(G);
+"C2 x S4"
 gap> cycle := CycleDigraph(9);
 <digraph with 9 vertices, 9 edges>
-gap> a := AutomorphismGroup(cycle);;
-gap> StructureDescription(a);
+gap> G := AutomorphismGroup(cycle);
+Group([ (1,2,3,4,5,6,7,8,9) ])
+gap> StructureDescription(G);
 "C9"
-gap> a := AutomorphismGroup(cycle, [1, 2, 3, 1, 2, 3, 1, 2, 3]);;
-gap> StructureDescription(a);
-"C3"
-gap> a := AutomorphismGroup(cycle, [[1, 4, 7], [2, 5, 8], [3, 6, 9]]);;
-gap> StructureDescription(a);
-"C3"
-]]></Example>
-    </Description>
-  </ManSection>
-<#/GAPDoc>
-
-<#GAPDoc Label="AutomorphismGroupMultiDigraph">
-  <ManSection>
-    <Attr Name="AutomorphismGroup" Label="for a multidigraph" Arg="digraph"/>
-    <Returns>A direct product of permutation groups.</Returns>
-    <Description>
-    If <A>digraph</A> is a multidigraph, then this function returns the
-    automorphism group of <A>digraph</A>, as a group of permutations on the
-    vertices and edges of <A>digraph</A>.<P/>
-
-    For convenience, the returned group is the direct product of the group of
-    automorphisms of the vertices of <A>digraph</A> with the stabiliser of the
-    vertices in the automorphism group of the edges. These two groups can be
-    accessed using the <Ref Oper="Projection"
-      Label="for a domain and a positive integer" BookName="ref"/> with second
-    argument <C>1</C> and <C>2</C>,
-    respectively.
-    <P/>
-
-    The permutations in the automorphism group of the edges act on the indices
-    of the edges of <A>digraph</A>.<P/>
-
-    The automorphism group is found using <URL
-      Text="bliss">http://www.tcs.tkk.fi/Software/bliss/index.html</URL> by
-    Tommi Junttila and Petteri Kaski.
-
-    <Example><![CDATA[
 gap> gr := DigraphEdgeUnion(CycleDigraph(3), CycleDigraph(3));
 <multidigraph with 3 vertices, 6 edges>
 gap> G := AutomorphismGroup(gr);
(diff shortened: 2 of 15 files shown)


wilfwilson:
Note: this PR makes no changes to anything at the C or C++ level, nor does any of the new code directly interface with any C functions. In particular, the new methods for `IsIsomorphicDigraph` and `IsomorphismDigraphs` call the already-existing `DigraphCanonicalLabelling` methods. Therefore this PR is entirely independent of any changes that we may or may not be making the software underpinning this package (see the discussion on issue #40).

doc/attr.xml
```diff
@@ -13,14 +13,14 @@
   <Attr Name="ChromaticNumber" Arg="digraph"/>
   <Returns> A non-negative integer.</Returns>
   <Description>
-    A digraph coloring is a labeling of the vertices (using precisely <A>n</A>
```
james-d-mitchell: This is weird, what is `n` here? This is not really related to your pull request, but since you are changing this man section anyway, why not remove the statement in brackets here? I guess this was copy pasted from another manual entry where the number of colours was specified by an argument `n`.

gap/bliss.gi
```diff
@@ -12,141 +13,247 @@
...
-                    "the second arg must be a list of length ",
-                    DigraphNrVertices(graph), " of integers in [1 .. ",
-                    DigraphNrVertices(graph), "],");
-    fi;
-    colors := list;
+  n := DigraphNrVertices(digraph);
+  colours := DIGRAPHS_ValidateVertexPartition(n, list);
+  if colours = fail then
+    ErrorNoReturn("Digraphs: DigraphCanonicalLabelling: usage,\n",
+                  "the argument <list> does not define a valid colouring of ",
```
james-d-mitchell: "Colouring" or "coloring"? There seem to be both in this file. I'm fine with either but we should probably be consistent.

gap/bliss.gi
```diff
@@ -12,141 +13,247 @@
...
+                  "the vertices of\n<digraph>. Either <list> should be a list ",
+                  "of length <n> consisting of integers\nin the range ",
+                  "[0 .. <n>], or <list> should be a list of non-empty ",
+                  "disjoint\nlists whose union is the vertices of <digraph>,");
```
james-d-mitchell: I think this error message could be more helpful. What is `n` for example? I think it would be better to add this error message to `DIGRAPHS_ValidateVertexPartition`and be more specific about what is wrong (as was previously attempted). I haven't yet looked at `DIGRAPHS_ValidateVertexPartition`, will comment more there.

gap/bliss.gi
```diff
@@ -12,141 +13,247 @@
...
   fi;
-
-  return DIGRAPH_CANONICAL_LABELING_COLORS(graph, colors);
+  return DIGRAPH_CANONICAL_LABELING_COLORS(digraph, colours);
```
james-d-mitchell: Shouldn't this be `LABELLING' for consistency?

gap/bliss.gi
```diff
@@ -12,141 +13,247 @@
...
+
+#
+
+InstallMethod(IsIsomorphicDigraph, "for digraphs and homogeneous lists",
+[IsDigraph, IsDigraph, IsHomogeneousList, IsHomogeneousList],
+function(gr1, gr2, c1, c2)
+  local m, colour1, n, colour2, class_sizes, i;
+
+  if IsMultiDigraph(gr1) or IsMultiDigraph(gr2) then
+    ErrorNoReturn("Digraphs: IsIsomorphicDigraph: usage,\n",
```
james-d-mitchell: Why is this not supported? On the face of it, isn't this just the same as the non-coloured case?

gap/bliss.gi
```diff
@@ -12,141 +13,247 @@
...
+  fi;
+
+  class_sizes := ListWithIdenticalEntries(n, 0);
+  for i in DigraphVertices(gr1) do
+    class_sizes[colour1[i]] := class_sizes[colour1[i]] + 1;
+    class_sizes[colour2[i]] := class_sizes[colour2[i]] - 1;
+  od;
+  if not ForAll(class_sizes, x -> x = 0) then
+    return false;
+  elif gr1 = gr2 and colour1 = colour2 then
```
james-d-mitchell: Do we have any evidence that equality checking is faster than checking isomorphism? I don't really have any preconceptions, except that finding a canonical labelling is usually very fast, so I'm just checking.

gap/bliss.gi
```diff
@@ -12,141 +13,247 @@
...
-  if not IsIsomorphicDigraph(g1, g2) then
+  if not IsIsomorphicDigraph(gr1, gr2) then
     return fail;
   fi;
 
-  if IsMultiDigraph(g1) then
-    label1 := DigraphCanonicalLabelling(g1);
-    label2 := DigraphCanonicalLabelling(g2);
+  if IsMultiDigraph(gr1) then
+    if gr1 = gr2 then
```
james-d-mitchell: Same comment as above, is this really better?

gap/bliss.gi
```diff
@@ -12,141 +13,247 @@
...
+
+#
+
+InstallMethod(IsomorphismDigraphs, "for digraphs and homogeneous lists",
+[IsDigraph, IsDigraph, IsHomogeneousList, IsHomogeneousList],
+function(gr1, gr2, c1, c2)
+  local m, colour1, n, colour2, class_sizes, label1, label2, i;
+
+  if IsMultiDigraph(gr1) or IsMultiDigraph(gr2) then
+    ErrorNoReturn("Digraphs: IsomorphismDigraphs: usage,\n",
```
james-d-mitchell: Again, why?

gap/bliss.gi
```diff
@@ -12,141 +13,247 @@
...
+  class_sizes := ListWithIdenticalEntries(n, 0);
+  for i in DigraphVertices(gr1) do
+    class_sizes[colour1[i]] := class_sizes[colour1[i]] + 1;
+    class_sizes[colour2[i]] := class_sizes[colour2[i]] - 1;
+  od;
+  if not ForAll(class_sizes, x -> x = 0) then
+    return fail;
+  fi;
+
+  if gr1 = gr2 and colour1 = colour2 then
```
james-d-mitchell: `elif` here as above, please.

gap/utils.gi
```diff
@@ -350,6 +350,64 @@ function(graph)
...
+# of the list [1 .. n]. A valid partition of [1 .. n] is either:
+#
+# 1. A list of length <n> consisting of numbers in the range [0 .. n]
+# 2. A list of non-empty disjoint lists whose union is [1 .. n]
+#
+# If <partition> is a valid partition of [1 .. n] then this global function
+# returns the partition, in form 1 (converting it to this form if necessary).
+# If <partition> is invalid, then the function returns <fail>
+
+InstallGlobalFunction(DIGRAPHS_ValidateVertexPartition,
```
james-d-mitchell: Shouldn't this be in `bliss.gi`?

gap/utils.gi
```diff
@@ -350,6 +350,64 @@ function(graph)
...
+function(n, partition)
+  local colours, vertices, i, vertex;
```
james-d-mitchell: Why not take third argument which is a string containing the name of the function calling `DIGRAPHS_ValidateVertexPartition`? Then you could have specific error messages in this function rather than the same repeated error message in the functions calling this one.

gap/utils.gi
```diff
@@ -350,6 +350,64 @@ function(graph)
...
+# Given a non-negative integer <n> and a homogeneous list <partition>,
+# this global function tests whether <partition> is a valid partition
+# of the list [1 .. n]. A valid partition of [1 .. n] is either:
+#
+# 1. A list of length <n> consisting of numbers in the range [0 .. n]
+# 2. A list of non-empty disjoint lists whose union is [1 .. n]
+#
+# If <partition> is a valid partition of [1 .. n] then this global function
+# returns the partition, in form 1 (converting it to this form if necessary).
+# If <partition> is invalid, then the function returns <fail>
```
james-d-mitchell: Missing full stop

gap/utils.gi
```diff
@@ -350,6 +350,64 @@ function(graph)
...
+    return fail;
+  fi;
+
+  # is <partition> a list of colours or a list of lists of vertices?
+
+  if IsPosInt(partition[1]) and Length(partition) = n then
+    colours := [];
+    for i in partition do
+      if not IsInt(i) or i < 0 or n < i then
+        return fail;
```
james-d-mitchell: Error message

gap/utils.gi
```diff
@@ -350,6 +350,64 @@ function(graph)
...
+                  "the first argument <n> must be a non-negative integer, and ",
+                  "the second argument\n<partition> must be a homogeneous ",
+                  "list,");
+  elif n = 0 then
+    if IsEmpty(partition) then
+      return partition;
+    fi;
+    return fail;
+  elif IsEmpty(partition) then
+    return fail;
```
james-d-mitchell: error

doc/bliss.xml
```diff
@@ -11,83 +12,58 @@
 <#GAPDoc Label="AutomorphismGroupDigraph">
 <ManSection>
   <Attr Name="AutomorphismGroup" Label="for a digraph" Arg="digraph"/>
-  <Oper Name="AutomorphismGroup" Label="for a digraph and a homogeneous list"
-    Arg="digraph, colors"/>
-  <Returns>A permutation group.</Returns>
+  <Returns>A permutation group, or a direct product of permutation
```
james-d-mitchell: Isn't a direct product of perm groups a perm group?

doc/bliss.xml
```diff
@@ -11,83 +12,58 @@
...
+    groups.</Returns>
   <Description>
-    If <A>digraph</A> is a digraph without multiple edges, then this function
-    returns the automorphism group of <A>digraph</A>, as a group of
-    permutations on the vertices of <A>digraph</A>. <P/>
+    If <A>digraph</A> is a digraph, then this attribute returns the group of
```
james-d-mitchell: `returns` -> `stores` or `contains` maybe?

doc/bliss.xml
```diff
@@ -11,83 +12,58 @@
...
+        If <A>digraph</A> is a multidigraph, then the automorphism group is
+        a group of permutations on the vertices and edges of <A>digraph</A>.
+        <P/>
+
+        For convenience, the group is returned as the direct product of the
+        group of automorphisms of the vertices of <A>digraph</A> with the
+        stabiliser of the vertices in the automorphism group of the edges. These
+        two groups can be accessed using the operation <Ref Oper="Projection"
+          Label="for a domain and a positive integer" BookName="ref"/>, with the
+        second argument being <C>1</C> or <C>2</C>, respectively. <P/>
```
james-d-mitchell: This is really good, can we say which projection corresponds to which action, and actually give an example? This would be most helpful.

doc/bliss.xml
```diff
@@ -98,72 +74,143 @@ gap> Range(Projection(G, 2));
...
+
+      This operation computes the automorphism group of a coloured digraph.
+      Currently, it is only possible to compute the automorphism group of a
+      coloured digraph in the &Digraphs; package if the digraph has no multiple
+      edges; see <Ref Prop="IsMultiDigraph" />. <P/>
+
+      A coloured digraph can be specified by its underlying digraph
+      <A>digraph</A> and its colouring <A>colours</A>. Let <C>n</C> be the
+      number of vertices of <A>digraph</A>. The colouring <A>colours</A> may
+      have one of the following two forms: <P/>
```
james-d-mitchell: I think the `<P/>` here is superfluous.

doc/attr.xml
wilfwilson: Yes this must have been copied from the documentation for `DigraphColouring` when the documentation for `ChromaticNumber` was add. I'll reword it.

gap/bliss.gi
wilfwilson: I favour colouring. I'll aim for consistency (including local variable names).

gap/bliss.gi
wilfwilson: Yes, I'll make this change. The only reason I didn't change the names of the C functions is because the Semigroups package called one of these directly and I didn't want to have to fiddle with that. However I'll make a PR to Semigroups to sort this out.

gap/utils.gi
wilfwilson: I envisage `DIGRAPHS_ValidateVertexPartition` being used by the methods in `grahom.gi` as well, so I thought this would be the best place for it.

gap/bliss.gi
wilfwilson: In a worst-case complexity sense, surely isomorphism testing is harder. For equality, it's a bit worse than `O(m + n)` since the out-neighbours of a vertex aren't required to be sorted. For each vertex of the two digraphs, you sort the respective out-neighbour list, and then compare, which doesn't seem that bad. It's implemented in `C` so it's fairly fast.

Here's some experimental evidence which I'd say shows that it is worth doing equality checking, although it's not dramatic.

```
gap> gr1 := CycleDigraph(50000);;
gap> gr2 := CycleDigraph(50000);;
gap> DigraphCanonicalLabelling(gr1);; 2 * time;
468
gap> gr1 = gr2;; time;
3
```

```
gap> gr1 := RandomDigraph(5000, 0.05);;
gap> gr2 := DigraphCopy(gr1);;
gap> DigraphCanonicalLabelling(gr1);; 2 * time;
198
gap> gr1 = gr2;; time;
4
```

```
gap> gr1 := RandomDigraph(5000, 0.2);;
gap> gr2 := DigraphCopy(gr1);;
gap> DigraphCanonicalLabelling(gr1);; 2 * time;
608
gap> gr1 = gr2;; time;
16
```

```
gap> gr1 := RandomDigraph(10000, 0.25);;
gap> gr2 := DigraphCopy(gr1);;
gap> DigraphCanonicalLabelling(gr1);; 2 * time;
3334
gap> gr1 = gr2;; time;
81
```

wilfwilson: (I do `2 * time;` for the canonical labelling since you've got to calculate it twice.)

doc/bliss.xml
wilfwilson: Yes. `AutomorphismGroup` for a multidigraph previously had its own manual section, but I combined it with the section for a digraph without multiple edges (but without colours). The returns bit for the multidigraph section said "A direct product of permutation groups", so I just kept that bit.

doc/bliss.xml
wilfwilson: You probably like it because you wrote it :) I'll do what I can to improve it.

wilfwilson:
I will modify `DIGRAPHS_ValidateVertexPartition` as you describe, to display a more relevant error message, rather than returning `fail` and the calling function displaying a less-relevant error message.

The functions with colourings that I wrote do not support multidigraphs since the two-argument `DigraphCanonicalLabelling` does not support multidigraphs, it just returns `fail` (although this behaviour is undocumented in the released package). I'm not sure why coloured multidigraph canonical labelling was not added. I'll see if I can work out whether it is easy to add.

I'm happy to disallow the colour 0.

Another question: for a colouring of the form `[[1, 4], [3], [2, 5]]`, we disallow empty sublists (since then it's not a homogeneous list), and this implies that the colours used have to be precisely some range `[0 .. m]`. However, for a colouring of the form `[1, 3, 3, 1]`, we do not currently require that the used colours form a range, so for instance in this last example the colours used are `[1, 3]`. How do you think we should resolve this inconsistency? Either the first form allows empty lists, or the second doesn't allow colours to be skipped? Or we leave it as-is. I suggest we leave it as it is.

wilfwilson:
I've answered my own question. This example uses the unique colour `1` for the first digraph, and the unique colour `2` for the second digraph. These aren't isomorphic as coloured digraphs since the colours don't match up! It's necessary for two coloured digraphs to have the same number of vertices of each particular colour in order to be isomorphic.

But the following code returns true rather than the (correct) false:
```
gap> a1 := Digraph([[2],[]]);;
gap> a2 := Digraph([[],[1]]);;
gap> q1 := DigraphCanonicalLabelling(a1, [1, 1]);;
gap> q2 := DigraphCanonicalLabelling(a2, [2, 2]);;
gap> OnDigraphs(a1, q1) = OnDigraphs(a2, q2);
true
```
(Note: this is hence a bug in the released version of the package).

Therefore a colouring of the form `[2, 2]` (not using the colour `1`) should be disallowed.

wilfwilson:
I've revisted what I had written before. However, for consistency's sake, I still think it's right that a colouring of the form `[1, 3, 2, 3, 1]` shouldn't miss out colours.

wilfwilson:
I think I've sorted all of this out now.

The main change I've made since your review was to add support for passing coloured multidigraphs to bliss. This mostly involved creating a function for constructing coloured multidigraphs for bliss, and a slightly different hook function. Coloured non-multidigraphs used one trick to encode their colours and directed edges, whilst non-coloured multidigraphs used another trick to encode their multiple edges. I just combined these things together.

That means we can ask about the automorphism group and canonical labelling (and hence isomorphism) of a coloured multidigraph. It seems to work fine.

james-d-mitchell:
@wilfwilson I agree about not missing colours, I think this is the only sensible option.

gap/utils.gi
james-d-mitchell: Right but `utils.g*` only contains things relating to the administration of the package, and not its digraphs functionality. I suggest moving this to `bliss.g*`

doc/bliss.xml
james-d-mitchell: Fair enough

doc/bliss.xml
james-d-mitchell: Right :1st_place_medal:

doc/z-chap6.xml
```diff
@@ -42,7 +45,7 @@ from} $E_a$ \emph{to} $E_b$. In this case we say that $E_a$ and $E_b$ are
     <#Include Label="DigraphEpimorphism">
     <#Include Label="EpimorphismsDigraphs">
     <#Include Label="GeneratorsOfEndomorphismMonoid">
-    <#Include Label="DigraphColoring">
+    <#Include Label="DigraphColouring">
```
james-d-mitchell: I think we'll need to keep `DigraphColoring` too for backwards compatibility.

gap/utils.gi
```diff
@@ -350,6 +350,113 @@ function(graph)
...
+                      "the argument <partition> does not define a ",
+                      "colouring of the vertices [1 .. ", n, "],\nsince ",
+                      "it does not assign a colour to the vertex ", i, ",");
+      fi;
+      return colours;
+    fi;
+  fi;
+  ErrorNoReturn("Digraphs: ", method, ": usage,\n",
+                "the argument <partition> does not define a ",
+                "colouring of the vertices [1 .. ", n, "].\nTo do so, the ",
```
james-d-mitchell: Remove the `To do so`

doc/z-chap6.xml
wilfwilson: The synonym already exists in `grahom.gd`, and there are tests using both `DigraphColoring` and `DigraphColouring`. I just changed the name of the mansection's label.

wilfwilson:
@james-d-mitchell All sorted.

PR #50, opened 2017-03-14
Hamiltonian cycles

le27:
Implements a new property `IsHamiltonianDigraph` which checks if a given digraph is hamiltonian, and a new attribute `HamiltonianPath` which finds a hamiltonian path if there is one.

### DIFF
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -1313,3 +1313,42 @@ fail]]></Example>
   </Description>
 </ManSection>
 <#/GAPDoc>
+  
+<#GAPDoc Label="HamiltonianPath">
+<ManSection>
+  <Attr Name="HamiltonianPath" Arg="digraph"/>
+  <Returns>A list or <K>fail</K>.</Returns>
+  <Description>
+    Returns a Hamiltonian path if one exists, <K>fail</K> if not.<P/>
+
+    A <E>Hamiltonian path</E> of a digraph with <C>n</C> vertices is directed
+    cycle of length <C>n</C>. If <A>digraph</A> is a digraph that contains a
+    Hamiltonian path, then this function returns one, described in the form
+    used by <Ref Attr="DigraphAllSimpleCircuits"/>. Note if <A>digraph</A> has
+    0 or 1 vertices, then <C>HamiltonianPath</C> returns <C>[]</C> or
+    <C>[1]</C>, respectively.<P/>
+
+    The method used in this attribute has the same worst case complexity as
+    <Ref Oper="DigraphMonomorphism"/>.
+
+    <Example><![CDATA[
+gap> g := Digraph([[]]);
+<digraph with 1 vertex, 0 edges>
+gap> HamiltonianPath(g);
+[ 1 ]
+gap> g := Digraph([[2], [1]]);
+<digraph with 2 vertices, 2 edges>
+gap> HamiltonianPath(g);
+[ 1, 2 ]
+gap> g := Digraph([[3], [], [2]]);
+<digraph with 3 vertices, 2 edges>
+gap> HamiltonianPath(g);
+fail
+gap> g := Digraph([[2], [3], [1]]);
+<digraph with 3 vertices, 3 edges>
+gap> HamiltonianPath(g);
+[ 1, 2, 3 ]
+]]></Example>
+  </Description>
+</ManSection>
+<#/GAPDoc>
--- a/doc/prop.xml
+++ b/doc/prop.xml
@@ -850,3 +850,41 @@ true
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="IsHamiltonianDigraph">
+<ManSection>
+  <Prop Name="IsHamiltonianDigraph" Arg="digraph"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    If <A>digraph</A> is Hamiltonian, then this property returns
+    <K>true</K>, and <K>false</K> if it is not. <P/>
+
+    A digraph with <C>n</C> vertices is <E>Hamiltonian</E> if it has a
+    directed cycle of length <C>n</C>. See Section <Ref Subsect="Definitions"
+     Style="Number" /> for the definition of a directed cycle.
+    Note the empty digraphs on 0 and 1 vertices are considered to be
+    Hamiltonian.<P/>
+
+    The method used in this operation has the worst case complexity as
+    <Ref Oper="DigraphMonomorphism"/>.
+
+    <Example><![CDATA[
+gap> g := Digraph([[]]);
+<digraph with 1 vertex, 0 edges>
+gap> IsHamiltonianDigraph(g);
+true
+gap> g := Digraph([[2], [1]]);
+<digraph with 2 vertices, 2 edges>
+gap> IsHamiltonianDigraph(g);
+true
+gap> g := Digraph([[3], [], [2]]);
+<digraph with 3 vertices, 2 edges>
+gap> IsHamiltonianDigraph(g);
+false
+gap> g := Digraph([[2], [3], [1]]);
+<digraph with 3 vertices, 3 edges>
+gap> IsHamiltonianDigraph(g);
+true
+]]></Example>
+  </Description>
+</ManSection>
+<#/GAPDoc>
--- a/doc/z-chap5.xml
+++ b/doc/z-chap5.xml
@@ -33,6 +33,7 @@
     <#Include Label="IsDirectedTree">
     <#Include Label="IsUndirectedTree">
     <#Include Label="IsEulerianDigraph">
+    <#Include Label="IsHamiltonianDigraph">
     <#Include Label="IsCycleDigraph">
   </Section>
   
--- a/gap/attr.gd
+++ b/gap/attr.gd
@@ -69,6 +69,7 @@ DeclareOperation("DIGRAPHS_MaximalSymmetricSubdigraph", [IsDigraph, IsBool]);
 
 DeclareAttribute("UndirectedSpanningTree", IsDigraph);
 DeclareAttribute("UndirectedSpanningForest", IsDigraph);
+DeclareAttribute("HamiltonianPath", IsDigraph);
 
 # AsGraph must be mutable for grape to function properly
 DeclareAttribute("AsGraph", IsDigraph, "mutable");
--- a/gap/prop.gi
+++ b/gap/prop.gi
@@ -567,3 +567,111 @@ function(gr)
   return true;
 
 end);
+
+# Meyniel's Theorem: a strongly connected digraph with n vertices, in which
+# any two non-adjacent vertices have full degree sum at least 2n − 1, is
+# Hamiltonian.
+# This function uses theorems 4.1 and 4.2 from the following paper:
+# Sufficient conditions for a digraph to be Hamiltonian
+# http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.99.4560&rep=rep1
+# &type=pdf
+# A vertex z dominates a pair of vertices {x, y} if z->x and z->y
+# A pair of vertices {x, y} dominates a vertex z if x->z and y->z
+# Theorem 4.1: a strongly connected digraph with n vertices in which every pair
+# of non-adjacent dominated vertices {x,y} satifies either of:
+# 1.(full degree of x) ≥ n and (full degree of y) ≥ n - 1
+# 2.(full degree of y) ≥ n and (full degree of x) ≥ n - 1
+# Is Hamiltonian.
+# Theorem 4.2: a strongly connected digraph with n vertices in which every pair
+# of non-adjacent vertices {x,y} which is dominated or dominating satifies:
+# 1. (out degree of x) + (in degree of y) ≥ n
+# 2. (out degree of y) + (in degree of x) ≥ n
+# Is Hamiltonian.
+InstallMethod(IsHamiltonianDigraph, "for a digraph",
+[IsDigraph],
+function(gr)
+  local out, indegs, fulldegs, outdegs, n, checkMT, check41, check42, iter,
+        dominatedcheck, dominatingcheck, adjmatrix, i, j, k, tempblist;
+
+  if DigraphNrVertices(gr) <= 1 and IsEmptyDigraph(gr) then
+    return true;
+  elif not IsStronglyConnectedDigraph(gr) then
+    return false;
+  fi;
+
+  if IsMultiDigraph(gr) then
+    gr := DigraphRemoveAllMultipleEdges(gr);
+  fi;
+  if DigraphHasLoops(gr) then
+    gr := DigraphRemoveLoops(gr);
+  fi;
+
+  n := DigraphNrVertices(gr);
+
+  if n <= 512 then
+    out := OutNeighbours(gr);
+    indegs := InDegrees(gr);
+    outdegs := OutDegrees(gr);
+    fulldegs := indegs + outdegs;
+    adjmatrix := BooleanAdjacencyMatrix(gr);
+    #checks if Meyniel's theorem, Theorem 4.1 or Theorem 4.2 are applicable.
+    checkMT := true;
+    check41 := true;
+    check42 := true;
+    for i in [1 .. n] do
+       for j in [1 .. n] do
+          if i <> j and not adjmatrix[j][i] and not adjmatrix[i][j] then
+              #Meyniel's theorem
+              if checkMT and fulldegs[i] + fulldegs[j] < 2 * n - 1 then
+                  checkMT := false;
+              fi;
+
+              if check41 or check42 then
+                dominatedcheck := false;
+                dominatingcheck := false;
+                for k in [1 .. n] do
+                  if adjmatrix[k][i] and adjmatrix[k][j] then
+                     dominatedcheck := true;
+                     break;
+                  fi;
+                od;
+                tempblist := adjmatrix[i];
+                IntersectBlist(tempblist, adjmatrix[j]);
+                dominatingcheck := true in tempblist;
+              fi;
+
+              #Theorem 4.1
+              if check41 and dominatedcheck then
+                 if fulldegs[i] < n - 1 or fulldegs[j] < n - 1 or
+                    (fulldegs[i] = n - 1 and fulldegs[j] = n - 1) then
+                   check41 := false;
+                 fi;
+              fi;
+
+              #Theorem 4.2
+              if check42 and (dominatingcheck or dominatedcheck) then
+                 if (indegs[i] + outdegs[j]) < n or
+                    (indegs[j] + outdegs[i]) < n then
+                   check42 := false;
+                 fi;
+              fi;
+          fi;
+          if not (checkMT or check41 or check42) then
+            break;
+          fi;
+      od;
+      if not (checkMT or check41 or check42) then
+        break;
+      fi;
+    od;
+    if checkMT or check41 or check42 then
+      return true;
+    fi;
+  fi;
+
+  return HamiltonianPath(gr) <> fail;
+end);
+
+InstallMethod(IsHamiltonianDigraph, "for a digraph with hamiltonian path",
+[IsDigraph and HasHamiltonianPath],
+x -> HamiltonianPath(x) <> fail);
--- a/doc/z-chap4.xml
+++ b/doc/z-chap4.xml
@@ -58,6 +58,7 @@
     <#Include Label="DigraphLayers">
     <#Include Label="DigraphDegeneracy">
     <#Include Label="DigraphDegeneracyOrdering">
+    <#Include Label="HamiltonianPath">
   </Section>
   
   <Section><Heading>Cayley graphs of groups</Heading>
--- a/gap/attr.gi
+++ b/gap/attr.gi
@@ -1544,3 +1544,42 @@ function(gr)
   SetIsUndirectedTree(out, true);
   return out;
 end);
+
+#
+
+InstallMethod(HamiltonianPath,
+"for a digraph",
+[IsDigraph],
+function(gr)
+  local path, out, iter, n;
+
+  if DigraphNrVertices(gr) <= 1 and IsEmptyDigraph(gr) then
+    if DigraphNrVertices(gr) = 0 then
+      return [];
+    else
+      return [1];
+    fi;
+  elif not IsStronglyConnectedDigraph(gr) then
+    return fail;
+  fi;
+
+  out := OutNeighbours(gr);
+
+  if DigraphNrVertices(gr) < 256 then
+    path := DigraphMonomorphism(CycleDigraph(DigraphNrVertices(gr)), gr);
+    if path = fail then
+      return fail;
+    fi;
+    return ImageListOfTransformation(path, DigraphNrVertices(gr));
+  fi;
+
+  iter := IteratorOfPaths(gr, 1, 1);
+  n := DigraphNrVertices(gr) + 1;
+  while not IsDoneIterator(iter) do
+    path := NextIterator(iter)[1];
+    if Length(path) = n then
+      return path;
+    fi;
+  od;
+  return fail;
+end);
--- a/gap/prop.gd
+++ b/gap/prop.gd
@@ -33,6 +33,7 @@ DeclareProperty("IsDirectedTree", IsDigraph);
 DeclareProperty("IsUndirectedTree", IsDigraph);
 DeclareProperty("IsUndirectedForest", IsDigraph);
 DeclareProperty("IsEulerianDigraph", IsDigraph);
+DeclareProperty("IsHamiltonianDigraph", IsDigraph);
 DeclareProperty("IsMeetSemilatticeDigraph", IsDigraph);
 DeclareProperty("IsJoinSemilatticeDigraph", IsDigraph);
 DeclareProperty("IsCycleDigraph", IsDigraph);
--- a/tst/standard/attr.tst
+++ b/tst/standard/attr.tst
@@ -1504,6 +1504,159 @@ gap> ArticulationPoints(Digraph([[1], [2]]));
 gap> ArticulationPoints(Digraph([[2], [1]]));
 [  ]
 
+#T# HamiltonianPath
+gap> g := Digraph([]);
+<digraph with 0 vertices, 0 edges>
+gap> HamiltonianPath(g);
+[  ]
+gap> g := Digraph([[]]);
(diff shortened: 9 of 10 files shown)


wilfwilson:
@le27 Please rebase this on top of the most recent commit in the `master` branch - doing so will require you to resolve one or more conflicts. This way you'll get the most up-to-date Travis setup, so that your pull request will be tested properly.

james-d-mitchell:
@le27 are you going to update this at some point?

le27:
sorry I thought I had fixed this a while ago I'll fix it now.

james-d-mitchell:
So, your second commit says that you are implementing a more efficient version of this, but in the test files, only very small graphs are tested. Can you add some tests (probably in the `extreme` directory) that actually test the performance of your implementation?

doc/prop.xml
```diff
@@ -788,3 +788,38 @@ true
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="IsHamiltonianDigraph">
+<ManSection>
+  <Prop Name="IsHamiltonianDigraph" Arg="digraph"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    If <A>digraph</A> is Hamiltonian, this property returns
+    <K>true</K>.<P/>
```
james-d-mitchell: So what does it return if it is not Hamiltonian?

doc/prop.xml
```diff
@@ -788,3 +788,38 @@ true
...
+
+    A Hamiltonian digraph is a digraph which has a circuit visiting 
+    each vertex exactly once.
+    Note the empty digraphs on 0 and 1 vertices are considered
```
james-d-mitchell: Do you mean digraphs with no edges? Or do you just mean "any digraph with 0 or 1 vertices is Hamiltonian."? I note in the code you really be a digraphs with 0 or 1 vertices and no edges, so a digraph with an 1 vertex and a loop is not Hamiltonian?

doc/prop.xml
```diff
@@ -788,3 +788,38 @@ true
...
 
+<#GAPDoc Label="IsHamiltonianDigraph">
+<ManSection>
+  <Prop Name="IsHamiltonianDigraph" Arg="digraph"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    If <A>digraph</A> is Hamiltonian, this property returns
+    <K>true</K>.<P/>
+
+    A Hamiltonian digraph is a digraph which has a circuit visiting
```
james-d-mitchell: "<E>Hamiltonian</E>" and wouldn't it be better to write, something like "A digraph is <E>Hamiltonian</E>..."?

james-d-mitchell: Probably you should also elaborate the meaning of circuit in this context too.

doc/prop.xml
james-d-mitchell: You should also say something here about the complexity of the implemented algorithm.

gap/prop.gi
```diff
@@ -555,3 +555,108 @@ function(gr)
...
+InstallMethod(IsHamiltonianDigraph, "for a digraph",
+[IsDigraph],
+function(g)
+  local g2, out, Indegs, Outdegs, Degrees, n, check,
+  dominatedcheck, dominatingordominatedcheck, i, j, k, dp;
+
+  if DigraphNrVertices(g) <= 1 and IsEmptyDigraph(g) then
+     return true;
+  fi;
+  if not IsStronglyConnectedDigraph(g) then
```
james-d-mitchell: `if` -> `elif` and remove the previous line

gap/prop.gi
```diff
@@ -555,3 +555,108 @@ function(gr)
   return true;
 
 end);
+
+#
+
```
james-d-mitchell: There are no comments in your code here. You should add either a description of the algorithm you are using, or a link/url/reference for the underlying algorithm. Without that this will be harder to maintain this code.

gap/prop.gi
```diff
@@ -555,3 +555,108 @@ function(gr)
...
+  local g2, out, Indegs, Outdegs, Degrees, n, check,
+  dominatedcheck, dominatingordominatedcheck, i, j, k, dp;
+
+  if DigraphNrVertices(g) <= 1 and IsEmptyDigraph(g) then
+     return true;
+  fi;
+  if not IsStronglyConnectedDigraph(g) then
+     return false;
+  fi;
+  g2 := DigraphRemoveLoops(g);
```
james-d-mitchell: This is unnecessarily wasteful, since the digraph `g` is copied here even if it has no loops anyway. Better do `if DigraphHasLoops(g) then ...` and otherwise set `g2 := g`

wilfwilson:
@james-d-mitchell is this the sort of thing you were on about, for the default DFS?

```
is_hamiltonian := function(gr)
  local iter, n;
  iter := IteratorOfPaths(gr, 1, 1);
  n := DigraphNrVertices(gr) + 1;
  while not IsDoneIterator(iter) do
    if Length(NextIterator(iter)[1]) = n then
      return true;
    fi;
  od;
  return false;
end;
```
We iterate over paths from the first vertex to itself and return `true` if we find one that contains every vertex.

wilfwilson:
@james-d-mitchell Here's an example where the above method works

```
gap> gr := DigraphFromGraph6String("SXqk_CiF?c?O?WACAwNIGGgpijPyB_LM?");
```

james-d-mitchell:
@wilfwilson yes something like this.

james-d-mitchell:
You can also use `DigraphMonomorphism(CycleDigraph(NrVertices(gr)), gr) <> fail;`. I just discussed this with Luke and he is going to experiment to decide which method to use in which situation.

doc/oper.xml
```diff
@@ -1503,3 +1503,41 @@ true
   </Description>
 </ManSection>
 <#/GAPDoc>
+  
+<#GAPDoc Label="HamiltonianPath">
+<ManSection>
+  <Prop Name="HamiltonianPath" Arg="digraph"/>
+  <Returns>A List or <K>fail</K>.</Returns>
```
james-d-mitchell: "List" -> "list" (no capital)

doc/oper.xml
```diff
@@ -1503,3 +1503,41 @@ true
...
+  <Description>
+    Returns a <E>Hamiltonian</E> path if one exists, <K>fail</K> if not.
```
james-d-mitchell: Only emphasis words when you are defining them, here you are not defining this, so lose the emphasis.

doc/oper.xml
```diff
@@ -1503,3 +1503,41 @@ true
...
+
+    A <E>Hamiltonian</E> path is a directed circuit visiting
+    each vertex exactly once. See Section <Ref Subsect="Definitions"
+     Style="Number" /> for the definition of a directed circuit.
+    Note the empty digraphs on 0 and 1 vertices are considered
+    Hamiltonian.<P/>
+
+    The method used in this operation has worst case complexity <M>O(n!)</M>
```
james-d-mitchell: I am not sure that this is correct, since you no longer use the naive DFS approach. Maybe remove this sentence altogether.

james-d-mitchell: Or say something like "This method uses DigraphMonomorphism, and hence has worst case complexity the same as DigraphMonomorphism."

doc/prop.xml
```diff
@@ -19,7 +19,7 @@
       Oper="CycleDigraph"/>.<P/>
 
     A digraph is a <E>cycle</E> if and only if it is strongly connected and has
-    the same number of edges as vertices. 
+    the same number of edges as vertices.
```
james-d-mitchell: This change is not related to this PR, so please drop it.

doc/prop.xml
```diff
@@ -46,7 +46,7 @@ true]]></Example>
   <Returns><K>true</K> or <K>false</K>.</Returns>
   <Description>
     A connected digraph is <E>biconnected</E> if it is still connected (in the
-    sense of <Ref Prop="IsConnectedDigraph"/>) when any vertex is removed. 
+    sense of <Ref Prop="IsConnectedDigraph"/>) when any vertex is removed.
     <C>IsBiconnectedDigraph</C> returns <K>true</K> if the digraph
```
james-d-mitchell: Same again, please drop this unrelated change.

doc/prop.xml
```diff
@@ -79,7 +79,7 @@ false]]></Example>
     <Ref Prop="IsAntisymmetricDigraph"/> and <Ref Prop="IsTransitiveDigraph"/>.
     For a partial order digraph <A>digraph</A>, the corresponding partial order
     is the relation <M>\leq</M>, defined by <M>x \leq y</M> if and only if
-    <C>[x, y]</C> is an edge of <A>digraph</A>.    
+    <C>[x, y]</C> is an edge of <A>digraph</A>.
```
james-d-mitchell: Same again, please drop this unrelated change.

doc/prop.xml
```diff
@@ -106,7 +106,7 @@ true
   <Prop Name="IsLatticeDigraph" Arg="digraph"/>
   <Returns><K>true</K> or <K>false</K>.</Returns>
   <Description>
-    
+
     <C>IsMeetSemilatticeDigraph</C> returns <K>true</K> if the digraph
```
james-d-mitchell: Same again, please drop this unrelated change.

doc/prop.xml
```diff
@@ -850,3 +850,41 @@ true
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="IsHamiltonianDigraph">
+<ManSection>
+  <Prop Name="IsHamiltonianDigraph" Arg="digraph"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    If <A>digraph</A> is <E>Hamiltonian</E>, this property returns
```
james-d-mitchell: No emphasis.

doc/prop.xml
```diff
@@ -850,3 +850,41 @@ true
...
+    <K>true</K>, and <K>false</K> if it is not. <P/>
+
+    A digraph is <E>Hamiltonian</E> if it has a directed circuit visiting
+    each vertex exactly once. See Section <Ref Subsect="Definitions"
+     Style="Number" /> for the definition of a directed circuit.
+    Note the empty digraphs on 0 and 1 vertices are considered
+    Hamiltonian.<P/>
+
+    The method used in this operation has worst case complexity <M>O(n!)</M>
```
james-d-mitchell: Same comment as before.

doc/z-chap5.xml
```diff
@@ -33,7 +33,13 @@
     <#Include Label="IsDirectedTree">
     <#Include Label="IsUndirectedTree">
     <#Include Label="IsEulerianDigraph">
+    <#Include Label="IsHamiltonianDigraph">
+    <#Include Label="HamiltonianPath">
```
james-d-mitchell: This is in the wrong part of the manual, it should be where the other attributes are, not in the section about properties.

doc/z-chap5.xml
```diff
@@ -33,7 +33,13 @@
...
     <#Include Label="IsCycleDigraph">
   </Section>
-  
+
+  <Section><Heading>Partial Orders</Heading>
+    <#Include Label="IsPartialOrderDigraph">
+    <#Include Label="IsMeetSemilatticeDigraph">
+  </Section>
```
james-d-mitchell: This section was deleted, you need to rebase on the current HEAD of the master branch.

gap/oper.gd
```diff
@@ -98,6 +98,8 @@ DeclareOperation("PartialOrderDigraphMeetOfVertices",
 [IsDigraph, IsPosInt, IsPosInt]);
 DeclareOperation("DigraphClosure", [IsDigraph, IsPosInt]);
 
+DeclareOperation("HamiltonianPath", [IsDigraph]);
```
james-d-mitchell: This should be declared as an attribute not an operation, so that it is not recalculated every time this is called. It should therefore go in the `attr.g*` files.

gap/oper.gi
```diff
@@ -1841,3 +1841,38 @@ function(gr, edges)
...
+      return [];
+    else
+      return [1];
+    fi;
+  elif not IsStronglyConnectedDigraph(gr) then
+    return fail;
+  fi;
+
+  out := OutNeighbours(gr);
+  obviouspath := true;
```
james-d-mitchell: Is it worth doing this? I think it probably isn't. Since `DigraphMonomorphism` will do this anyway.

gap/oper.gi
```diff
@@ -1841,3 +1841,38 @@ function(gr, edges)
...
+    if not i + 1 in out[i] then
+      obviouspath := false;
+      break;
+    fi;
+  od;
+  if obviouspath and 1 in out[DigraphNrVertices(gr)] then
+    return [1 .. DigraphNrVertices(gr)] + 0;
+  fi;
+
+  map := DigraphMonomorphism(CycleDigraph(DigraphNrVertices(gr)), gr);
```
james-d-mitchell: This only works if the number of vertices in the digraph is at most 512, so you should check for this and use the method suggested by Wilf if the number of vertices is greater than 512.

gap/prop.gi
```diff
@@ -567,3 +567,109 @@ function(gr)
...
+  fi;
+
+  if IsMultiDigraph(gr) then
+    gr := DigraphRemoveAllMultipleEdges(gr);
+  fi;
+  if DigraphHasLoops(gr) then
+    gr := DigraphRemoveLoops(gr);
+  fi;
+
+  if IsCycleDigraph(gr) then
```
james-d-mitchell: I don't know if this check is worth it either.

tst/standard/oper.tst
```diff
@@ -1993,6 +1993,123 @@ gap> edges := [[1, 1], [2, 3]];
 gap> IsMaximalMatching(gr, edges);
 true
```
james-d-mitchell: These tests are in the wrong file.

doc/oper.xml
```diff
@@ -1503,3 +1503,41 @@ true
   </Description>
 </ManSection>
 <#/GAPDoc>
+  
+<#GAPDoc Label="HamiltonianPath">
+<ManSection>
+  <Prop Name="HamiltonianPath" Arg="digraph"/>
+  <Returns>A list or <K>fail</K>.</Returns>
+  <Description>
+    Returns a <E>Hamiltonian</E> path if one exists, <K>fail</K> if not.
```
james-d-mitchell: No emphasis here

doc/oper.xml
```diff
@@ -1503,3 +1503,41 @@ true
...
+
+    A Hamiltonian path is a directed circuit visiting
```
james-d-mitchell: Emphasis here.

doc/oper.xml
```diff
@@ -1503,3 +1503,41 @@ true
...
+    each vertex exactly once. See Section <Ref Subsect="Definitions"
+     Style="Number" /> for the definition of a directed circuit.
+    Note the empty digraphs on 0 and 1 vertices are considered
+    Hamiltonian.<P/>
+
+    The method used in this operation has the worst case complexity as
+    DigraphMonomorphism.
```
james-d-mitchell: \<Ref Oper="DigraphMonomorphism"/\>

doc/prop.xml
```diff
@@ -850,3 +850,41 @@ true
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="IsHamiltonianDigraph">
+<ManSection>
+  <Prop Name="IsHamiltonianDigraph" Arg="digraph"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    If digraph is <E>Hamiltonian</E>, this property returns
```
james-d-mitchell: No emphasis.

doc/prop.xml
```diff
@@ -850,3 +850,41 @@ true
...
+    <K>true</K>, and <K>false</K> if it is not. <P/>
+
+    A digraph is <E>Hamiltonian</E> if it has a directed circuit visiting
+    each vertex exactly once. See Section <Ref Subsect="Definitions"
+     Style="Number" /> for the definition of a directed circuit.
+    Note the empty digraphs on 0 and 1 vertices are considered
+    Hamiltonian.<P/>
+
+    The method used in this operation has the worst case complexity as
+    DigraphMonomorphism.
```
james-d-mitchell: Reference

gap/attr.gd
```diff
@@ -69,6 +69,7 @@ DeclareOperation("DIGRAPHS_MaximalSymmetricSubdigraph", [IsDigraph, IsBool]);
 
 DeclareAttribute("UndirectedSpanningTree", IsDigraph);
 DeclareAttribute("UndirectedSpanningForest", IsDigraph);
+DeclareOperation("HamiltonianPath", [IsDigraph]);
```
wilfwilson: This should be an attribute in my opinion. Since its value can't change.

doc/attr.xml
```diff
@@ -1310,3 +1310,42 @@ fail]]></Example>
   </Description>
 </ManSection>
 <#/GAPDoc>
+  
+<#GAPDoc Label="HamiltonianPath">
+<ManSection>
+  <Prop Name="HamiltonianPath" Arg="digraph"/>
```
wilfwilson: This shouldn't be a `Prop` - the type should agree with how `HamiltonianPath` (currently it's an Operation, and so it should say `Oper` here, although further down I suggest that `HamiltonianPath` should instead be declared as an attribute, and so this should be `Attr`).

doc/attr.xml
```diff
@@ -1310,3 +1310,42 @@ fail]]></Example>
...
+  <Returns>A list or <K>fail</K>.</Returns>
+  <Description>
+    Returns a Hamiltonian path if one exists, <K>fail</K> if not.
+
+    A <E>Hamiltonian path</E> is a directed circuit visiting
+    each vertex exactly once. See Section <Ref Subsect="Definitions"
+     Style="Number" /> for the definition of a directed circuit.
```
wilfwilson: You should actually read these definitions, since currently this documentation is not quite true. We define a directed circuit to be a particular kind sequence of vertices and edges. Your method just returns a sequence of vertices, so it's *not* returning a directed circuit, as you state. Please look at the documentation for `DigraphAllSimpleCircuits` to see an example of how to be more precise about this. Also - you should say `cycle` rather than `circuit` - again, look at the definition.

You could say something like `A <E>Hamiltonian path</E> of a digraph with <C>n</C> vertices is directed cycle of length <C>n</C>. If <A>digraph</A> is a digraph that contains a Hamiltonian path, then this function returns one, described in the form used by <Ref Attr="DigraphAllSimpleCircuits"/>...` and whatever else you need to make this bit of documentation correct and consistent.

doc/attr.xml
```diff
@@ -1310,3 +1310,42 @@ fail]]></Example>
...
+    Note the empty digraphs on 0 and 1 vertices are considered
+    Hamiltonian.<P/>
+
+    The method used in this operation has the worst case complexity as
```
wilfwilson: It's not currently an `operation` - so make sure this agrees with however it is declared.

doc/attr.xml
```diff
@@ -1310,3 +1310,42 @@ fail]]></Example>
...
+  <Prop Name="HamiltonianPath" Arg="digraph"/>
+  <Returns>A list or <K>fail</K>.</Returns>
+  <Description>
+    Returns a Hamiltonian path if one exists, <K>fail</K> if not.
+
+    A <E>Hamiltonian path</E> is a directed circuit visiting
+    each vertex exactly once. See Section <Ref Subsect="Definitions"
+     Style="Number" /> for the definition of a directed circuit.
+    Note the empty digraphs on 0 and 1 vertices are considered
+    Hamiltonian.<P/>
```
wilfwilson: You've not defined what Hamiltonian is so this doesn't make sense on its own. Instead, you could say something like `If <A>digraph</A> has 0 or 1 vertices, then <C>HamiltonianPath</C> returns [] or [1], respectively.`

doc/attr.xml
```diff
@@ -1310,3 +1310,42 @@ fail]]></Example>
...
+
+    The method used in this operation has the worst case complexity as
+    DigraphMonomorphism. See <Ref Oper="DigraphMonomorphism"/> for information
```
wilfwilson: just have `...complexity as <Ref Oper="DigraphMonomorphism"/>`.

doc/prop.xml
```diff
@@ -852,3 +852,42 @@ true
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="IsHamiltonianDigraph">
+<ManSection>
+  <Prop Name="IsHamiltonianDigraph" Arg="digraph"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    If digraph is Hamiltonian, this property returns
```
wilfwilson: `A` tags around `digraph`, since it's the argument. Also have the word `then` after the comma: `, then this...`

doc/prop.xml
```diff
@@ -852,3 +852,42 @@ true
...
+    <K>true</K>, and <K>false</K> if it is not. <P/>
+
+    A digraph is <E>Hamiltonian</E> if it has a directed circuit visiting
+    each vertex exactly once. See Section <Ref Subsect="Definitions"
+     Style="Number" /> for the definition of a directed circuit.
+    Note the empty digraphs on 0 and 1 vertices are considered
```
wilfwilson: Add "to be", i.e. `considered to be Hamiltonian`

doc/prop.xml
```diff
@@ -852,3 +852,42 @@ true
...
+<ManSection>
+  <Prop Name="IsHamiltonianDigraph" Arg="digraph"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    If digraph is Hamiltonian, this property returns
+    <K>true</K>, and <K>false</K> if it is not. <P/>
+
+    A digraph is <E>Hamiltonian</E> if it has a directed circuit visiting
+    each vertex exactly once. See Section <Ref Subsect="Definitions"
+     Style="Number" /> for the definition of a directed circuit.
```
wilfwilson: directed circuit -> directed cycle

doc/prop.xml
```diff
@@ -852,3 +852,42 @@ true
...
+<#GAPDoc Label="IsHamiltonianDigraph">
+<ManSection>
+  <Prop Name="IsHamiltonianDigraph" Arg="digraph"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    If digraph is Hamiltonian, this property returns
+    <K>true</K>, and <K>false</K> if it is not. <P/>
+
+    A digraph is <E>Hamiltonian</E> if it has a directed circuit visiting
+    each vertex exactly once. See Section <Ref Subsect="Definitions"
```
wilfwilson: Change it to `if it has a directed cycle of length <C>n</C>.`

doc/prop.xml
```diff
@@ -852,3 +852,42 @@ true
...
 
+<#GAPDoc Label="IsHamiltonianDigraph">
+<ManSection>
+  <Prop Name="IsHamiltonianDigraph" Arg="digraph"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    If digraph is Hamiltonian, this property returns
+    <K>true</K>, and <K>false</K> if it is not. <P/>
+
+    A digraph is <E>Hamiltonian</E> if it has a directed circuit visiting
```
wilfwilson: Change it to `A digraph with <C>n</C> vertices is...`

doc/prop.xml
```diff
@@ -852,3 +852,42 @@ true
...
+    each vertex exactly once. See Section <Ref Subsect="Definitions"
+     Style="Number" /> for the definition of a directed circuit.
+    Note the empty digraphs on 0 and 1 vertices are considered
+    Hamiltonian.<P/>
+
+    The method used in this operation has the worst case complexity as
+    DigraphMonomorphism. See <Ref Oper="DigraphMonomorphism"/> for information
```
wilfwilson: Again, just have `...complexity as <Ref Oper="DigraphMonomorphism"/>.`

gap/attr.gd
```diff
@@ -69,6 +69,7 @@ DeclareOperation("DIGRAPHS_MaximalSymmetricSubdigraph", [IsDigraph, IsBool]);
 
 DeclareAttribute("UndirectedSpanningTree", IsDigraph);
 DeclareAttribute("UndirectedSpanningForest", IsDigraph);
+DeclareAttribute("HamiltonianPath", [IsDigraph]);
```
wilfwilson: Get rid of the square brackets here.

wilfwilson:
You should check why this is happening:
```
gap> HamiltonianPath(CycleDigraph(512));
fail
```
It works fine for 511 and 513.

wilfwilson:
@le27 what do you think the status of this is? It seems like you've addressed all of the comments that I've made so far. It seems like I've only reviewed the documentation so far, however, so if you think it's ready then I'll now review the code.

le27:
As far as I can tell the code should be ready as it is now. Please check it for problems as it is.

wilfwilson:
Will do!

gap/prop.gi
```diff
@@ -567,3 +567,118 @@ function(gr)
...
+          fi;
+      od;
+      if not (checkMT or check41 or check42) then
+        break;
+      fi;
+    od;
+    if checkMT or check41 or check42 then
+      return true;
+    fi;
+
```
wilfwilson: From this point on, I think you should replace the following by:
```
  fi;

  return HamiltonianPath(gr) <> fail;
end);
```

gap/prop.gi
```diff
@@ -567,3 +567,118 @@ function(gr)
...
+  iter := IteratorOfPaths(gr, 1, 1);
+  n := n + 1;
+  while not IsDoneIterator(iter) do
+    if Length(NextIterator(iter)[1]) = n then
+      return true;
+    fi;
+  od;
+  return false;
+end);
```
wilfwilson: After this, you should add the following method:
```
InstallMethod(IsHamiltonianDigraph, "for a digraph with hamiltonian path",
[IsDigraph and HasHamiltonianPath],
x -> HamiltonianPath(x) <> fail);
```
Because, if we've already calculated a path, we already know the answer.

wilfwilson: And also, add a test for this case.

wilfwilson:
Thanks @le27. Although @james-d-mitchell hasn't clicked "Approve" to the changes that he requested, it looks like you've addressed his concerns, so I'm happy to merge.

PR #51, opened 2017-03-20
Add 4 properties relating to partial orders

james-d-mitchell:
This is really @ChristopherRussell PR, but anyway. In this PR we add functionality to Digraphs for checking if a digraph is a partial order (as a binary relation), a meet or join semilattice, and a lattice. 

There is documentation and tests for everything.


james-d-mitchell:
This appears to be exposing a seg fault somewhere.

james-d-mitchell:
The tests are failing due to a bug in `IsAsymmetricDigraph` detailed in Issue #52.

gap/prop.gi
```diff
@@ -8,7 +8,47 @@
...
+[IsDigraph],
+function(digraph)
+  local out_nbrs, i, j, intr;
+  if IsPartialOrderDigraph(digraph) then
+    out_nbrs := OutNeighbours(digraph);
+    for i in [1 .. Size(out_nbrs)] do
+      for j in [i + 1 .. Size(out_nbrs)] do
+        # Is there a better way of doing this? JDM
+        intr := Intersection(out_nbrs[i], out_nbrs[j]);
+        if not intr in out_nbrs then
```
wilfwilson: I'm pretty sure that this doesn't work, since `intr` is guaranteed to be sorted (it's a set, since it is produced by `Intersection`), however the entries of `out_nbrs` are not guaranteed to be sorted. So although `int` might equal `[1, 3, 5]` and `out[5]` might equal `[5, 3, 1]`, this line would return `false` even though I think you would want it to return `true`.

doc/prop.xml
```diff
@@ -8,6 +8,106 @@
...
 ##
 
+<#GAPDoc Label="IsPartialOrderDigraph">
+<ManSection>
+  <Prop Name="IsPartialOrderDigraph" Arg="digraph"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    This is a synonym for <Ref Prop="IsReflexiveDigraph"/> and
+    <Ref Prop="IsAntisymmetricDigraph"/> and <Ref Prop="IsTransitiveDigraph"/>.
+
```
wilfwilson: Is it `x <= y` if and only if `x -> y` is an edge? This matters for later on when we're talking about upper and lower bounds.

gap/prop.gi
wilfwilson: Also, seeing as every vertex has a loop, it is certainly true that if `intr = out_nbrs[i]` for some `i`, then that `i` must be one of the elements of `intr`. So with this method, instead of looking through the whole of `out_nbrs`, you can narrow down the search.

gap/prop.gi
```diff
@@ -8,7 +8,47 @@
...
+        fi;
+      od;
+    od;
+    return true;
+  else
+    return false;
+  fi;
+end);
+
+InstallMethod(IsJoinSemilatticeDigraph, "for a digraph",
```
wilfwilson: Or of combining the two?

doc/prop.xml
wilfwilson: On reflection, you've implicitly defined the partial order to be `x >= y` if and only if `x -> y` is an edge. So, for instance, the digraph `Digraph([[1, 2, 3], [2], [3]])`, which looks like (ignoring the loops):

```
   --> 2
 /
1 
 \
   --> 3
```
is, according to your code below, a join semilattice, but not a meet semilattice. For what it's worth, I would expect it to be the other way round, although I accept that it's arbitrary. However there's probably a well-established convention in use.

gap/prop.gi
```diff
@@ -8,7 +8,47 @@
...
+    in_nbrs := InNeighbours(digraph);
+    for i in [1 .. Size(in_nbrs)] do
+      for j in [i + 1 .. Size(in_nbrs)] do
+        # Is there a better way of doing this? JDM
+        intr := Intersection(in_nbrs[i], in_nbrs[j]);
+        if not intr in in_nbrs then
+          return false;
+        fi;
+      od;
+    od;
```
wilfwilson: Unless I'm mistaken, the function `IsJoinSemilatticeDigraph(digraph)` does exactly the same as `IsMeetSemilatticeDigraph(DigraphReverse(digraph))`. Is this mathematically the right thing to do? I think so; in which case you should replace this whole function which one that just calls `IsMeetSemilatticeDigraph` on the reversed digraph.

gap/prop.gi
```diff
@@ -8,7 +8,47 @@
...
+[IsDigraph],
+function(digraph)
+  local in_nbrs, intr, i, j;
+  if IsPartialOrderDigraph(digraph) then
+    in_nbrs := InNeighbours(digraph);
+    for i in [1 .. Size(in_nbrs)] do
+      for j in [i + 1 .. Size(in_nbrs)] do
+        # Is there a better way of doing this? JDM
+        intr := Intersection(in_nbrs[i], in_nbrs[j]);
+        if not intr in in_nbrs then
```
wilfwilson: I think this will usually work, since by the way the C function works, `InNeighbours` is usually sorted by construction.

doc/prop.xml
```diff
@@ -8,6 +8,106 @@
...
+    <Example><![CDATA[
+gap> gr := Digraph([[1, 3], [2, 3], [3]]);
+<digraph with 3 vertices, 5 edges>
+gap> IsPartialOrderDigraph(gr);
+true
+gap> gr := CycleDigraph(5);
```
wilfwilson: Run the manual examples; there will currently be errors. For example, this line is terminated by a single `;`, however nothing is printed on the following line. It should say `<digraph with 5 vertices, 5 edges>`.

doc/prop.xml
```diff
@@ -8,6 +8,106 @@
...
+false
+gap> gr := Digraph([[1], [2], [1 .. 3]]);
+<digraph with 3 vertices, 5 edges>
+gap> IsJoinSemilatticeDigraph(gr);
+true
+gap> IsMeetSemilatticeDigraph(gr);
+false
+gap> IsLatticeDigraph(gr);
+false
+gap> gr := Digraph([[1 .. 4], [2, 4], [3 ,4], [4]]);
```
wilfwilson: Again, this should have an output, saying `<digraph with 4 vertices, 9 edges>`.
Also, fix the spacing in `[3 ,4]` to `[3, 4]` to fit in with the rest of the package.

tst/standard/prop.tst
```diff
@@ -1028,6 +1028,222 @@ gap> g := Digraph([[2, 3], [3], [1]]);
...
+> , 6, 35, 103, 124, 152, 153, 159, 160, 161, 198, 243, 245, 258 ], [ 2, 3, 4, 5\
+> , 6, 17, 28, 35, 36, 56, 69, 93, 106, 120, 128, 131, 152, 153, 160, 162, 171, \
+> 174, 178, 181, 182, 243, 245, 259 ], [ 2, 3, 4, 6, 17, 35, 105, 123, 127, 152,\
+>  153, 160, 201, 247, 249, 254, 260 ], [ 2, 3, 6, 126, 152, 248, 251, 261 ], [ \
+> 2, 3, 4, 5, 6, 94, 108, 131, 177, 180, 182, 249, 254, 262 ], [ 2, 3, 4, 6, 107\
+> , 130, 200, 202, 248, 249, 251, 254, 263 ], [ 2, 3, 6, 129, 213, 248, 251, 252\
+> , 264 ], [ 2, 6, 252, 265 ], [ 6, 86, 92, 102, 119, 146, 255, 256, 257, 258, 2\
+> 59, 260, 261, 262, 263, 264, 265, 266 ] ] );
+<digraph with 266 vertices, 3121 edges>
+gap> gr := DigraphReflexiveTransitiveClosure(gr);
```
wilfwilson: Also, on reflection, I'm not sold on the purpose of this test. It is mostly testing `IsPartialOrderDigraph` - but since by construction the digraph is reflexive and transitive, the only thing it is checking is `IsAntisymmetricDigraph`. And since this digraph is neither a meet nor a join semilattice, that code isn't really being stretched either – presumably the methods there deduce almost immediately that the thing isn't a semilattice.

If you're going to go to the effort of putting such a large and fiddling test into the test file, I'd suggest doing one which is actually going to exhibit interest behaviour. So, for instance, one which is maybe a meet semilattice, but not a join semilattice. This way you're testing the code much more thoroughly.

wilfwilson:
This is an example of how your function fails, due to it assuming that the list of out-neighbours is sorted:
```
gap> a := Digraph([[1, 2, 3, 4], [2, 3, 4], [3, 4], [4]]);
<digraph with 4 vertices, 10 edges>
gap> b := Digraph([[4, 3, 2, 1], [4, 3, 2], [4, 3], [4]]);
<digraph with 4 vertices, 10 edges>
gap> IsMeetSemilatticeDigraph(a);
true
gap> IsMeetSemilatticeDigraph(b);
false
gap> a = b;
true
```
`a` and `b` are equal digraphs, so their graph-theoretic properties should be the same. However, your function thinks that `a` is a meet semilattice but that `b` isn't. This is because the out-neighbours of `b` are not sorted into ascending order; they are not a "set".

wilfwilson:
Also, what do we do about multiple edges? Can a multidigraph define a partial order? Should
```
gap> gr := Digraph([[1, 2, 2], [2]]);
```
define a partial order? I'd argue that multiple edges just confuses things, and that they shouldn't be allowed to define partial orders. Because this is strange:
```
gap> gr := Digraph([[1,2,2,2,3,3,3],[2,3,3,3],[3,3]]);
<multidigraph with 3 vertices, 13 edges>
gap> IsJoinSemilatticeDigraph(gr);
false
gap> IsMeetSemilatticeDigraph(gr);
false
gap> IsPartialOrderDigraph(gr);
true
gap> gr := DigraphRemoveAllMultipleEdges(gr);
<digraph with 3 vertices, 6 edges>
gap> IsLatticeDigraph(gr);
true
```

wilfwilson:
For now, let's go by your definition that a partial order digraph defines a partial order by `x >= y` if and only if `x -> y` is an edge. I believe that this is a better implementation of `IsMeetSemilatticeDigraph`:
```
InstallMethod(IsMeetSemilatticeDigraph, "for a digraph",
[IsDigraph],
function(digraph)
  local topo, gr, nbs, n, x, k, i, j;

  if not IsPartialOrderDigraph(digraph) then
    return false;
  fi;

  topo := DigraphTopologicalSort(digraph);
  gr := OnDigraphs(digraph, PermList(topo));
  nbs := OutNeighboursMutableCopy(gr);
  Apply(nbs, Set);
  n := DigraphNrVertices(gr);

  for i in [1 .. n] do
    for j in [i + 1 .. n] do
      if j in nbs[i] or i in nbs[j] then
        continue;
      fi;
      x := Intersection(nbs[i], nbs[j]);
      if IsEmpty(x) then
        return false;
      fi;
      k := Maximum(x); # Check whether <k> is the meet of <i> and <j>
      if Length(nbs[k]) < Length(x) then
        return false;
      fi;
    od;
  od;
  return true;
end);

InstallMethod(IsJoinSemilatticeDigraph, "for a digraph",
[IsDigraph],
digraph -> IsMeetSemilatticeDigraph(DigraphReverse(digraph)));
```

Let me know @ChristopherRussell if you want help understanding it or working out why it should be quicker.

gap/prop.gi
james-d-mitchell: The problem with this is that it will use more memory for the same calculation, is this really a good idea?

wilfwilson: Yes I suppose so.

The way to solve this is to have two user-facing methods, `IsJoinSemilatticeDigraph` and `IsMeetSemilatticeDigraph`, and one hidden method `HiddenFunction` (say).

The hidden function implements what is currently implemented in `IsMeetSemilatticeDigraph`, except that it accepts a list of adjacencies rather than a digraph.

`IsMeetSemilatticeDigraph`  checks that the digraph is a partial order, then calls `HiddenFunction(OutNeighbours(digraph))`.

`IsJoinemilatticeDigraph`  checks that the digraph is a partial order, then calls `HiddenFunction(InNeighbours(digraph))`.

This uses no extra memory.

One of the first things that the method for `IsJoinSemilatticeDigraph` does is `in_nbrs := InNeighbours(digraph);`.
All that `DigraphReverse(digraph)` does is `Digraph(InNeighbours(digraph))`.

wilfwilson: (I had some extraneous extra text in that comment from a previous version of what I was going to write but I've deleted it)

ChristopherRussell:
Thank you @wilfwilson for the feature comments as well as the helpful code style suggestions. I would like to ask you about how your IsMeetSemilatticeDigraph function works. I am not sure why you do

`gr := OnDigraphs(digraph, PermList(topo));`

but it would probably be best to talk with you in person. I am also thinking of adding a function, say DigraphMeet, which takes as input a digraph and two of its vertices and returns the meet of these two vertices.

wilfwilson:
I'll happily do my best to explain it to you tomorrow @ChrisJefferson at the Wednesday meeting. Briefly, this line relabels the vertices of the digraph according to `DigraphTopologicalSort`. This guarantees that if `x -> y` is an edge in the digraph, then the number x is greater than the number y. It allows us to make some time-saving assumptions later in the function. In particular, if the meet of two vertices exists, it would have to be the greatest number that is an out-neighbour of both x and y.

james-d-mitchell:
Think you meant @ChristopherRussell not @ChrisJefferson

wilfwilson:
Whoops!

ChristopherRussell:
> This guarantees that if x -> y is an edge in the digraph, then the number x is greater than the number y.

I don't see this happening. Maybe you want to use OnDigraphs with the inverse of PermList(topo)? Try my example and see that x -> y doesn't imply x > y in the splash of the digraph gr but it does in the splash of gr2.

```
gap> a:=Digraph([[3],[5],[5],[3],[]]);;
gap> Splash(DotDigraph(a));
gap> topo:=DigraphTopologicalSort(a);;
gap> gr := OnDigraphs(a, PermList(topo));;
gap> Splash(DotDigraph(gr));
gap> gr2 := OnDigraphs(a, PermList(topo)^-1);
gap> Splash(DotDigraph(gr2));
```

wilfwilson:
Yes, using the inverse seems to do the trick.

james-d-mitchell:
@wilfwilson @ChristopherRussell can this be closed, as it is superseded by PR #59

wilfwilson:
Yes. @ChristopherRussell talking about closing this yesterday, it's just not happened yet.

Issue #52, opened 2017-03-20
IsAsymmetricDigraph causes seg fault


james-d-mitchell:
The following example:

```
gap> gr := Digraph([[1], [2], [1 .. 3]]);;
gap> IsAntisymmetricDigraph(gr);
```

causes a segmentation fault in version 0.7.0 (and all previous versions where this function exists).

I am about to commit a pull request which fixes this.

james-d-mitchell:
This is the issue causing the tests to fail in PR #51.

wilfwilson:
Resolved by PR #53.

PR #53, opened 2017-03-20
Fix IsAntisymmetricDigraph

james-d-mitchell:
This PR fixes a (probably long standing) bug in the kernel function `FuncIS_ANTISYMMETRIC_DIGRAPH` which assumed incorrectly that a variable was a plain list. 

This resolves Issue #52 and is the cause of the segmentation fault on travis for PR #51.

### DIFF
--- a/src/digraphs.c
+++ b/src/digraphs.c
@@ -701,7 +701,7 @@ static Obj FuncIS_ANTISYMMETRIC_DIGRAPH(Obj self, Obj adj) {
           level++;
           nbs = ELM_PLIST(adj, j);
           stack += 4;
-          stack[0] = INT_INTOBJ(ADDR_OBJ(nbs)[k]);
+          stack[0] = INT_INTOBJ(ELM_LIST(nbs, k));
           stack[1] = 1;
           stack[2] = j;  // I am wasting memory here, duplicating info
           stack[3] = last1;
--- a/tst/testinstall.tst
+++ b/tst/testinstall.tst
@@ -271,6 +271,11 @@ gap> AsDigraph(f);
 gap> AsDigraph(f, 4);
 fail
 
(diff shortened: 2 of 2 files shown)


james-d-mitchell:
When merged, this should also be put into stable-0.7

tst/testinstall.tst
```diff
@@ -271,6 +271,11 @@ gap> AsDigraph(f);
 gap> AsDigraph(f, 4);
 fail
 
+# Fix bug in FuncIS_ANTISYMMETRIC_DIGRAPH
```
wilfwilson: I'd recommend referencing Issue #52 in the comment here.

src/digraphs.c
```diff
@@ -701,7 +701,7 @@ static Obj FuncIS_ANTISYMMETRIC_DIGRAPH(Obj self, Obj adj) {
           level++;
           nbs = ELM_PLIST(adj, j);
           stack += 4;
-          stack[0] = INT_INTOBJ(ADDR_OBJ(nbs)[k]);
+          stack[0] = INT_INTOBJ(ELM_LIST(nbs, k));
```
wilfwilson: I wouldn't have realised that `ADDR_OBJ` wasn't valid here. But I'm glad I know now.

james-d-mitchell: It's not that `ADDR_OBJ(nbs)` isn't valid, it's that when `nbs` is a range, you cannot get the `k`th element of the range by `ADDR_OBJ(nbs)[k]`.

james-d-mitchell:
I've made the change you suggested by adding the issue number to the test.

src/digraphs.c
wilfwilson: Yes of course, that makes sense.

PR #54, opened 2017-03-20
Add debug make target

james-d-mitchell:
Add a debug make target, set DEBUG/NDEBUG as a flag in `Makefile.am` rather than manually, fix compiler warnings, and add gap-debug.h for easier debugging of the kernel module.


src/digraphs.c
```diff
@@ -896,24 +905,24 @@ static Obj FuncDIGRAPH_SYMMETRIC_SPANNING_FOREST(Obj self, Obj adj) {
       stack[1] = 1;
 
       while (1) {
-        j = stack[0];
-        k = stack[1];
+        j   = stack[0];
```
wilfwilson: What's the reasoning behind this kind of formatting change?

src/digraphs.c
```diff
@@ -896,24 +905,24 @@ static Obj FuncDIGRAPH_SYMMETRIC_SPANNING_FOREST(Obj self, Obj adj) {
...
+        k   = stack[1];
         nbs = ELM_PLIST(adj, j);
 
         // idea: is <nbs[k]> a new vertex? add edges <j> <-> <nbs[k]> if so
 
         // if we're finished with <j>, or <nbs[k]> doesn't exist, then backtrack
         if (ptr[j] || k > (UInt) LEN_LIST(nbs)) {
-          ptr[j] = 1; // vertex <j> and its descendents are now fully explored
+          ptr[j] = 1;  // vertex <j> and its descendents are now fully explored
```
wilfwilson: And this here – it would be nice to know what style guide you're going by so that we can all be using it.

wilfwilson:
Don't think I'm qualified to review this, but from looking over the code, I don't see anything surprising (beyond what I commented about the formatting changes you made).

james-d-mitchell:
Maybe @mtorpey can take a look? I guess you could also try `make debug` to see if it works for you @wilfwilson

src/digraphs.c
james-d-mitchell: I'm using clang-format, which is configures in the file `src/.clang-format`. This means that beyond the choices made in the configuration file (which were meant to be roughly the same choices as we were (mostly) using already), all I am doing is `clang-format -i src/*`, which automatically reformats all the kernel files.

src/digraphs.c
james-d-mitchell: See the comment above.

src/digraphs.c
wilfwilson: Great, thanks

wilfwilson:
When running `make debug`, it compiles, but with the following warning:

```
  CC       src/digraphs_la-homos.lo
src/homos.c:554:1: warning: 'inline' is not at beginning of declaration [-Wold-style-declaration]
 static bool inline is_adjacent_digraph(Digraph* digraph, Vertex i, Vertex j) {
 ^
src/homos.c:1210:1: warning: 'inline' is not at beginning of declaration [-Wold-style-declaration]
 static bool inline is_adjacent_graph(Graph* graph, Vertex i, Vertex j) {
 ^
cc1: warning: unrecognized command line option '-Wno-c++11-extensions'
```
Digraphs runs fine for me when it's compiled this way.

mtorpey:
Having looked through the diff of your first commit all looks pretty sane (though I don't claim to understand all the ins and outs) and doesn't seem to break anything.  I like having the accessible test for whether Digraphs is optimised or not.

Like Wilf, I'm getting that little warning when I compile.  Full output is as follows:
```
[14:45:27] ~/gap/pkg/digraphs $ make debug 
test -z "digraphs.la" || rm -f digraphs.la
rm -f ./so_locations
rm -rf .libs _libs
rm -rf src/.libs src/_libs
rm -rf src/bliss-0.73/.libs src/bliss-0.73/_libs
rm -f *.o
rm -f src/*.o
rm -f src/*.lo
rm -f src/bliss-0.73/*.o
rm -f src/bliss-0.73/*.lo
rm -f *.lo
  CC       src/digraphs_la-digraphs.lo
  CC       src/digraphs_la-homos.lo
src/homos.c:554:1: warning: ‘inline’ is not at beginning of declaration [-Wold-style-declaration]
 static bool inline is_adjacent_digraph(Digraph* digraph, Vertex i, Vertex j) {
 ^~~~~~
src/homos.c:1210:1: warning: ‘inline’ is not at beginning of declaration [-Wold-style-declaration]
 static bool inline is_adjacent_graph(Graph* graph, Vertex i, Vertex j) {
 ^~~~~~
cc1: warning: unrecognized command line option ‘-Wno-c++11-extensions’
  CC       src/digraphs_la-perms.lo
  CC       src/digraphs_la-schreier-sims.lo
  CXX      src/bliss-0.73/digraphs_la-defs.lo
  CXX      src/bliss-0.73/digraphs_la-graph.lo
  CXX      src/bliss-0.73/digraphs_la-partition.lo
  CXX      src/bliss-0.73/digraphs_la-orbit.lo
  CXX      src/bliss-0.73/digraphs_la-uintseqhash.lo
  CXX      src/bliss-0.73/digraphs_la-heap.lo
  CXX      src/bliss-0.73/digraphs_la-timer.lo
  CXX      src/bliss-0.73/digraphs_la-utils.lo
  CXX      src/bliss-0.73/digraphs_la-bliss_C.lo
  CXXLD    digraphs.la
/bin/mkdir -p ./bin/x86_64-pc-linux-gnu-gcc-default64
cp .libs/digraphs.so /home/mtorpey/gap/pkg/digraphs/bin/x86_64-pc-linux-gnu-gcc-default64/digraphs.so
SUCCESS!
```
These warnings only show up when `make debug` is run.

I've noticed that after running `make debug`, simply running `make` does nothing.  Is there a way to compile with optimisations *explicitly turned on*, and if not, should there be?  I'm thinking something like `make optimize` or `make ndebug`.

james-d-mitchell:
I've updated this to remove the warnings you mentioned and some more warnings too

PR #57, opened 2017-03-22
Added DigraphClosure

sriubas:
Adding an operation DigraphClosure which returns k-closure for a simple graph.

### DIFF
--- a/doc/oper.xml
+++ b/doc/oper.xml
@@ -1449,7 +1449,7 @@ fail
     <Example><![CDATA[
 gap> gr := CompleteDigraph(6);;
 gap> DigraphRemoveEdges(gr, [[1, 2], [2, 1]]);;
-gap> closure := DigraphClosure(gr, 6);                                                    
+gap> closure := DigraphClosure(gr, 6);
 <digraph with 6 vertices, 30 edges>
 gap> IsCompleteDigraph(closure);
 true
--- a/doc/oper.xml
+++ b/doc/oper.xml
@@ -1429,3 +1429,31 @@ fail
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="DigraphClosure">
+<ManSection>
+  <Oper Name="DigraphClosure" Label="for a digraph and positive integer"
+        Arg="digraph, k"/>
+  <Returns>A digraph</Returns>
+  <Description>
+    Given a symmetric loopless digraph with no multiple edges <A>digraph</A>,
+    the <E><A>k</A>-closure of <A>digraph</A></E> is defined to be the unique
+    smallest symmetric loopless digraph <C>C</C> with no multiple edges on the
+    vertices of <A>digraph</A> that contains all the edges of <A>digraph</A>
+    and satsifies the property that the sum of the degrees of every two
+    non-adjacenct vertices in <C>C</C> is less than <A>k</A>. See <Ref
+    Oper="IsSymmetricDigraph"/>, <Ref Oper="DigraphHasLoops"/>, <Ref
+    Oper="IsMultiDigraph"/>, and <Ref Oper="OutDegreeOfVertex"/>. <P/>
+
+    The operation <C>DigraphClosure</C> returns the <A>k</A>-closure of <A>digraph</A>.
+    <Example><![CDATA[
+gap> gr := CompleteDigraph(6);;
+gap> DigraphRemoveEdges(gr, [[1, 2], [2, 1]]);;
+gap> closure := DigraphClosure(gr, 6);                                                    
+<digraph with 6 vertices, 30 edges>
+gap> IsCompleteDigraph(closure);
+true
+]]></Example>
+  </Description>
+</ManSection>
+<#/GAPDoc>
--- a/gap/oper.gi
+++ b/gap/oper.gi
@@ -1730,3 +1730,44 @@ function(digraph, i, j)
 
   return fail;
 end);
+
+#
+
+InstallMethod(DigraphClosure,
+"for a digraph and an integer",
+[IsDigraph, IsPosInt],
+function(digraph, k)
+  local out, adj_mat, degs, n, new_edge, row, i, j;
+
+  if not IsSymmetricDigraph(digraph) or DigraphHasLoops(digraph)
+      or IsMultiDigraph(digraph) then
+    ErrorNoReturn("Digraphs: DigraphClosure: usage,\n",
+                  "the digraph must by symmetric, without loops, and no ",
+                  "multiple edges,");
+  fi;
+
+  out := [];
+  adj_mat := BooleanAdjacencyMatrixMutableCopy(digraph);
+  for row in adj_mat do
+    Add(out, ShallowCopy(row));
+  od;
+  degs := ShallowCopy(InDegrees(digraph));
+  n   := DigraphNrVertices(digraph);
+  new_edge := true;
+  while new_edge do
+    new_edge := false;
+    for i in [1 .. n] do
+      for j in [1 .. n] do
+        if j <> i and not out[i][j] and degs[i] + degs[j] >= k then
+          new_edge := true;
+          out[i][j] := true;
+          out[j][i] := true;
+          degs[i] := degs[i] + 1;
+          degs[j] := degs[j] + 1;
+        fi;
+      od;
+    od;
+  od;
+
+  return DigraphByAdjacencyMatrix(out);
+end);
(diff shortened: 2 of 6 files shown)


gap/oper.gi
```diff
@@ -1684,3 +1684,45 @@ function(digraph, list)
...
+  fi;
+
+  out := [];
+  adj_mat := BooleanAdjacencyMatrix(digraph);
+  for row in adj_mat do
+    Add(out, ShallowCopy(row));
+  od;
+  degs:= ShallowCopy(InDegrees(digraph));
+  n   := DigraphNrVertices(digraph);
+  while true do
```
wilfwilson: In my opinion, the previous line should say `new_edge := true` and this while loop should say `while new_edge do`. This removes the need for the `if` statement at the end of the this `while` loop.

gap/oper.gi
```diff
@@ -1684,3 +1684,45 @@ function(digraph, list)
...
+
+# 
+
+InstallMethod(DigraphClosure,
+"for a digraph and an integer",
+[IsDigraph, IsPosInt],
+function(digraph, k)
+  local out, adj_mat, degs, n, new_edge, row, i, j;
+
+  if not IsSymmetricDigraph(digraph) or DigraphHasLoops(digraph) then
```
wilfwilson: This doesn't check for multiple edges.

gap/oper.gi
```diff
@@ -1684,3 +1684,45 @@ function(digraph, list)
...
+    ErrorNoReturn("Digraphs: DigraphClosure: usage,\n",
+                  "the graph must by symmetric and without loops,");
```
wilfwilson: graph -> digraph

wilfwilson:
This should be submitted to `master` rather than `stable-0.7`. We now use `master` for new features, and `stable-0.7` for bug fixes only.

doc/oper.xml
```diff
@@ -1387,3 +1387,27 @@ gap> DigraphShortestDistance(gr, [1, 3], [3, 5]);
...
 <#/GAPDoc>
+
+<#GAPDoc Label="DigraphClosure">
+<ManSection>
+  <Oper Name="DigraphClosure" Label="for a digraph and positive integer"
+        Arg="digraph, k"/>
+  <Returns>A digraph</Returns>
+  <Description>
+    Given a symmetric loopless digraph with no multiple edges <A>digraph</A>,
+    the <A>k</A>-closure of <A>digraph</A> is defined to be the unique smallest
```
wilfwilson: Since you're defining k-closure, I think you should enclose it in E tags: `<E><A>k</A>-closure</E>`

doc/oper.xml
```diff
@@ -1387,3 +1387,27 @@ gap> DigraphShortestDistance(gr, [1, 3], [3, 5]);
...
+    symmetric loopless digraph <C>C</C> with no multiple edges on the vertices
+    of <A>digraph</A> which contains all the edges of <A>digraph</A> and
```
wilfwilson: which -> that

doc/oper.xml
```diff
@@ -1387,3 +1387,27 @@ gap> DigraphShortestDistance(gr, [1, 3], [3, 5]);
...
+    satsifies the property that the sum degrees of every two non-adjacenct
+    vertices in <C>C</C> is less than <A>k</A>.<P/> 
+
+    The function returns the <A>k</A>-closure of <A>digraph</A>.
```
wilfwilson: Should be either `This operation` (not "the"), or `The operation <C>DigraphClosure</C>`.

doc/oper.xml
```diff
@@ -1387,3 +1387,27 @@ gap> DigraphShortestDistance(gr, [1, 3], [3, 5]);
...
+<ManSection>
+  <Oper Name="DigraphClosure" Label="for a digraph and positive integer"
+        Arg="digraph, k"/>
+  <Returns>A digraph</Returns>
+  <Description>
+    Given a symmetric loopless digraph with no multiple edges <A>digraph</A>,
+    the <A>k</A>-closure of <A>digraph</A> is defined to be the unique smallest
+    symmetric loopless digraph <C>C</C> with no multiple edges on the vertices
+    of <A>digraph</A> which contains all the edges of <A>digraph</A> and
+    satsifies the property that the sum degrees of every two non-adjacenct
```
wilfwilson: "sum degrees" sounds ungrammatical. Perhaps "the sum of thes degree of each two...".

doc/oper.xml
```diff
@@ -1387,3 +1387,27 @@ gap> DigraphShortestDistance(gr, [1, 3], [3, 5]);
...
+    vertices in <C>C</C> is less than <A>k</A>.<P/> 
+
```
wilfwilson: I'd recommend adding a `<Ref ...>` reference to `IsSymmetricDigraph`, `DigraphHasLoops`, `IsMultiDigraph`, and `OutDegreeOfVertex`.

doc/oper.xml
```diff
@@ -1387,3 +1387,27 @@ gap> DigraphShortestDistance(gr, [1, 3], [3, 5]);
...
+    The function returns the <A>k</A>-closure of <A>digraph</A>.
+    <Example><![CDATA[
+gap> gr := CompleteDigraph(6);;
+gap> DigraphRemoveEdges(gr, [[1,2],[2,1]]);;
+gap> DigraphClosure(gr,6);
```
wilfwilson: Space between `gr` and `6` please.

doc/oper.xml
```diff
@@ -1387,3 +1387,27 @@ gap> DigraphShortestDistance(gr, [1, 3], [3, 5]);
...
+    the <A>k</A>-closure of <A>digraph</A> is defined to be the unique smallest
+    symmetric loopless digraph <C>C</C> with no multiple edges on the vertices
+    of <A>digraph</A> which contains all the edges of <A>digraph</A> and
+    satsifies the property that the sum degrees of every two non-adjacenct
+    vertices in <C>C</C> is less than <A>k</A>.<P/> 
+
+    The function returns the <A>k</A>-closure of <A>digraph</A>.
+    <Example><![CDATA[
+gap> gr := CompleteDigraph(6);;
+gap> DigraphRemoveEdges(gr, [[1,2],[2,1]]);;
```
wilfwilson: I'm being picky, but I prefer the manual examples to follow our linting guidelines (although we can't currently run the linter directly on these files yet). So this line would be: `gap> DigraphRemoveEdges(gr, [[1, 2], [2, 1]]);;`

doc/oper.xml
```diff
@@ -1387,3 +1387,27 @@ gap> DigraphShortestDistance(gr, [1, 3], [3, 5]);
...
+gap> DigraphClosure(gr,6);                                                    
+<digraph with 6 vertices, 30 edges>
+]]></Example>
```
wilfwilson: Maybe call the last digraph `closure`, and then in another line, ask `IsCompleteDigraph(closure);`, to demonstrate that the closure is the complete digraph in this case.

gap/oper.gi
```diff
@@ -1684,3 +1684,43 @@ function(digraph, list)
...
+InstallMethod(DigraphClosure,
+"for a digraph and an integer",
+[IsDigraph, IsPosInt],
+function(digraph, k)
+  local out, adj_mat, degs, n, new_edge, row, i, j;
+
+  if not IsSymmetricDigraph(digraph) or DigraphHasLoops(digraph)
+      or IsMultiDigraph(digraph) then
+    ErrorNoReturn("Digraphs: DigraphClosure: usage,\n",
+      "the digraph must by symmetric, without loops, and no multiple edges,");
```
wilfwilson: The linter isn't yet set up to catch these kinds of problems. This line should be aligned with the one above:

```
    ErrorNoReturn("Digraphs: DigraphClosure: usage,\n",
                           "the digraph...
```
However, this makes your line too long, so you'll need to break it, so it's something like"
```
    ErrorNoReturn("Digraphs: DigraphClosure: usage,\n",
                  "the digraph must by symmetric, without loops, and no ",
                  "multiple edges,");
```

gap/oper.gi
```diff
@@ -1684,3 +1684,43 @@ function(digraph, list)
...
+  fi;
+
+  out := [];
+  adj_mat := BooleanAdjacencyMatrix(digraph);
```
wilfwilson: Not something you need to do anything about, but since we have `AdjacencyMatixMutableCopy` now, I think we should also have `BooleanAdjacencyMatixMutableCopy`.

james-d-mitchell: Agreed, `BooleanAdjacencyMatrixMutableCopy` is a good one.

wilfwilson:
@sriubas you should merge `master` into your branch - currently there are conflicts, which means that this pull request can't be merged.

wilfwilson:
Thanks for sorting this out @sriubas. There's just one problem left: you've used the word `BooleanAdjacencyMatixMutableCopy` - missing an `r` from `Matrix`. That's why Travis is failing.

PR #58, opened 2017-03-22
Adding DigraphClosure

sriubas:
Adding an operation DigraphClosure which returns k-closure for a simple graph.


sriubas:
Closing, open by mistake.

PR #59, opened 2017-03-23
Adding 4 properties and 2 operations relating to partial order digraphs

ChristopherRussell:
This pull request is to replace #51 and has been updated to take on board all comments in that thread. I have now added two functions, which find the meet and the join of two vertices of a digraph. They are called PartialOrderDigraphMeetOfVertices and PartialOrderDigraphJoinOfVertices.


doc/oper.xml
```diff
@@ -1387,3 +1387,34 @@ gap> DigraphShortestDistance(gr, [1, 3], [3, 5]);
...
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="PartialOrderDigraphMeetOfVertices">
+<ManSection>
+  <Oper Name="PartialOrderDigraphMeetOfVertices" Label="for a digraph and two vertices"
+        Arg="digraph, u, v"/>
+  <Oper Name="PartialOrderDigraphJoinOfVertices" Label="for a digraph and two vertices"
+        Arg="digraph, u, v"/>
+  <Returns>An integer</Returns>
```
wilfwilson: I think we decided the following functionality for these operations:

If the digraph is not a partial order digraph, then do an `ErrorNoReturn`. 
Otherwise, if the meet/join does not exist, then return `fail`.

The first is an error because it doesn't make sense to ask for meets and joins in a digraph that isn't a partial order. And the second is fail because it's a valid question to ask, it's just that the things don't exist.

Therefore the implementation, and the documentation, should reflect this functionality.

doc/oper.xml
```diff
@@ -1387,3 +1387,34 @@ gap> DigraphShortestDistance(gr, [1, 3], [3, 5]);
...
+    <Description>
+    If the first argument is a partial order digraph <Ref Prop="IsPartialOrderDigraph"/>
+    then the function returns the meet, or the join, of the two input vertices. 
+    The meet (or join) is guarenteed to exist and be unique when the first
```
wilfwilson: guaranteed is the proper spelling

doc/oper.xml
```diff
@@ -1387,3 +1387,34 @@ gap> DigraphShortestDistance(gr, [1, 3], [3, 5]);
...
+<#GAPDoc Label="PartialOrderDigraphMeetOfVertices">
+<ManSection>
+  <Oper Name="PartialOrderDigraphMeetOfVertices" Label="for a digraph and two vertices"
+        Arg="digraph, u, v"/>
+  <Oper Name="PartialOrderDigraphJoinOfVertices" Label="for a digraph and two vertices"
+        Arg="digraph, u, v"/>
+  <Returns>An integer</Returns>
+    <Description>
+    If the first argument is a partial order digraph <Ref Prop="IsPartialOrderDigraph"/>
+    then the function returns the meet, or the join, of the two input vertices.
```
wilfwilson: "then these operations return..."

doc/oper.xml
```diff
@@ -1387,3 +1387,34 @@ gap> DigraphShortestDistance(gr, [1, 3], [3, 5]);
...
+        Arg="digraph, u, v"/>
+  <Returns>An integer</Returns>
+    <Description>
+    If the first argument is a partial order digraph <Ref Prop="IsPartialOrderDigraph"/>
+    then the function returns the meet, or the join, of the two input vertices. 
+    The meet (or join) is guarenteed to exist and be unique when the first
+    argument satisfies <Ref Prop="IsMeetSemilatticeDigraph"/> 
+    (or <Ref Prop="IsJoinSemilatticeDigraph"/>) - see the documentation for these
+    properties for the definition of the meet (or the join).
+
```
wilfwilson: The documentation should note that, in the case that the meet/join does not exist, the operation returns `<K>fail</K>`.

doc/oper.xml
```diff
@@ -1387,3 +1387,34 @@ gap> DigraphShortestDistance(gr, [1, 3], [3, 5]);
...
+<digraph with 4 vertices, 9 edges>
+gap> PartialOrderDigraphMeetOfVertices(gr, 2, 3);
+4
+gap> PartialOrderDigraphJoinOfVertices(gr, 2, 3);
+1
+gap> PartialOrderDigraphMeetOfVertices(gr, 1, 2);
+2
+gap> PartialOrderDigraphJoinOfVertices(gr, 3, 4);
+3
+]]></Example>
```
wilfwilson: Have one or more examples where the meet/join doesn't exist.

doc/prop.xml
```diff
@@ -8,6 +8,83 @@
...
 
+<#GAPDoc Label="IsPartialOrderDigraph">
+<ManSection>
+  <Prop Name="IsPartialOrderDigraph" Arg="digraph"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    This is a synonym for <Ref Prop="IsReflexiveDigraph"/> and
+    <Ref Prop="IsAntisymmetricDigraph"/> and <Ref Prop="IsTransitiveDigraph"/>.
+    We will consider a digraph as a partial order with relation defined by
+    <A>x</A> <M>\leq</M> <A>y</A> if and only if <A>(x,y)</A> is an edge.
```
wilfwilson: We use `<A>` tags only for the specific arguments defined for the function that is being documented.

Since `x` and `y` are not arguments to `IsPartialOrderDigraph`, so they shouldn't be in these tags. Similarly for `(x,y)`.

wilfwilson: In this package we use the notation `[x, y]` to define an edge. So overall, I think this would be better:

For a partial order digraph `<A>digraph</A>`, the corresponding partial order is the relation `<M>\leq</M>`, defined by `<M>x \leq y</M>` if and only if `<C>[x, y]</C>` is an edge of `<A>digraph</A>`.

doc/prop.xml
```diff
@@ -8,6 +8,83 @@
...
+    <A>digraph</A> is a meet semilattice; <C>IsJoinSemilatticeDigraph</C>
+    returns <K>true</K> if the digraph <A>digraph</A> is a join semilattice;
+    and  <C>IsLatticeDigraph</C> returns <K>true</K> if the digraph
+    <A>digraph</A> is both a meet and a join semilattice.
+    <P/>
+
+    We consider a partial order digraph <Ref Prop="IsPartialOrderDigraph"/>
+    to be a digraph with the relation <A>x</A> <M>\leq</M> <A>y</A> if and only
+    if <A>(x,y)</A> is an edge. A digraph is a <E>meet semilattice</E> if it is
+    a partial order and every pair of vertices has a unique greatest lower bound
```
wilfwilson: The word "unique" should be removed since the word "greatest" implies uniqueness.

doc/prop.xml
```diff
@@ -8,6 +8,83 @@
...
+  <Description>
+    
+    <C>IsMeetSemilatticeDigraph</C> returns <K>true</K> if the digraph
+    <A>digraph</A> is a meet semilattice; <C>IsJoinSemilatticeDigraph</C>
+    returns <K>true</K> if the digraph <A>digraph</A> is a join semilattice;
+    and  <C>IsLatticeDigraph</C> returns <K>true</K> if the digraph
+    <A>digraph</A> is both a meet and a join semilattice.
+    <P/>
+
+    We consider a partial order digraph <Ref Prop="IsPartialOrderDigraph"/>
```
wilfwilson: This is basically what was written in the documentation for `IsPartialOrderDigraph`, and which I've just suggested changes to. So make the same changes here, too.

doc/prop.xml
```diff
@@ -8,6 +8,83 @@
...
+    to be a digraph with the relation <A>x</A> <M>\leq</M> <A>y</A> if and only
+    if <A>(x,y)</A> is an edge. A digraph is a <E>meet semilattice</E> if it is
+    a partial order and every pair of vertices has a unique greatest lower bound
+    (meet) with respect to the aforementioned relation. A <E>join semilattice</E>
+    is a partial order where every pair of vertices has a unique least upper
```
wilfwilson: And again, "unique".

gap/oper.gi
```diff
@@ -1684,3 +1684,53 @@ function(digraph, list)
...
+
+  nbs := OutNeighboursMutableCopy(digraph);
+  intr := Intersection(nbs[i], nbs[j]);
+  for x in intr do
+    if intr = Set(nbs[x]) then
+      return x;
+    fi;
+  od;
+
+  ErrorNoReturn("Digraphs: PartialOrderDigraphJoinOfVertices: \n",
```
wilfwilson: this should be `return fail;`

gap/oper.gi
```diff
@@ -1684,3 +1684,53 @@ function(digraph, list)
...
+                "Join not found, either it does not exist",
+                " or it is not unqiue,");
```
wilfwilson: It doesn't make sense to talk of a pair of vertices of having more than one meet/join, since by definition, a meet/join is a unique element.

gap/oper.gi
```diff
@@ -1684,3 +1684,53 @@ function(digraph, list)
...
+end);
+
+InstallMethod(PartialOrderDigraphMeetOfVertices,
```
wilfwilson: Same comments as for the join version.

gap/oper.gi
```diff
@@ -1684,3 +1684,53 @@ function(digraph, list)
...
+# Partial Order Digraphs
+
+InstallMethod(PartialOrderDigraphJoinOfVertices,
+"for a digraph and two positive integers",
+[IsDigraph, IsPosInt, IsPosInt],
+function(digraph, i, j)
+  local x, nbs, intr;
+
+  if not IsPartialOrderDigraph(digraph) then
+    ErrorNoReturn("Digraphs: PartialOrderDigraphJoinOfVertices: usage, \n",
```
wilfwilson: get rid of the space in `usage, \n"`

gap/oper.gi
```diff
@@ -1684,3 +1684,53 @@ function(digraph, list)
...
+                  "the first argument must be a partial order digraph,");
```
wilfwilson: it's often helpful to give the problematic variable name so that the user can debug if necessary. So if this instead said

```
...first argument <digraph> must be...`
```
Then the user knows that they can access the digraph with the variable name digraph.

doc/prop.xml
```diff
@@ -8,6 +8,83 @@
...
+    <Example><![CDATA[
+gap> gr := Digraph([[1, 3], [2, 3], [3]]);
+<digraph with 3 vertices, 5 edges>
+gap> IsPartialOrderDigraph(gr);
+true
+gap> gr := CycleDigraph(5);
+<digraph with 5 vertices, 5 edges>
+gap> IsPartialOrderDigraph(gr);
+false
+]]></Example>
```
wilfwilson: Would be good to have one or two examples that are multidigraphs too.

james-d-mitchell:
@wilfwilson the commits should be squashed before merging

wilfwilson:
@james-d-mitchell certainly

gap/oper.gi
```diff
@@ -1684,3 +1684,51 @@ function(digraph, list)
...
+
+InstallMethod(PartialOrderDigraphMeetOfVertices,
+"for a digraph and two positive integers",
+[IsDigraph, IsPosInt, IsPosInt],
+function(digraph, i, j)
+  local x, nbs, intr;
+
+  if not IsPartialOrderDigraph(digraph) then
+    ErrorNoReturn("Digraphs: PartialOrderDigraphMeetOfVertices: usage,\n",
+                  "the first argument <digraph> must be a partial order",
```
wilfwilson: there should be a space after "order" here, since the word "digraph" on the line below doesn't have a space before it. As things stand, this will print "...partial orderdigraph,"

gap/oper.gi
```diff
@@ -1684,3 +1684,51 @@ function(digraph, list)
...
+end);
+
+InstallMethod(PartialOrderDigraphMeetOfVertices,
+"for a digraph and two positive integers",
+[IsDigraph, IsPosInt, IsPosInt],
+function(digraph, i, j)
+  local x, nbs, intr;
+
+  if not IsPartialOrderDigraph(digraph) then
+    ErrorNoReturn("Digraphs: PartialOrderDigraphMeetOfVertices: usage,\n",
```
wilfwilson: Please add a test that tests this line of code, `i.e.` asking for the `PartialOrderDigraphMeetOfVertices`, since it doesn't have any code coverage yet.

gap/oper.gi
```diff
@@ -1684,3 +1684,51 @@ function(digraph, list)
...
+
+InstallMethod(PartialOrderDigraphJoinOfVertices,
+"for a digraph and two positive integers",
+[IsDigraph, IsPosInt, IsPosInt],
+function(digraph, i, j)
+  local x, nbs, intr;
+
+  if not IsPartialOrderDigraph(digraph) then
+    ErrorNoReturn("Digraphs: PartialOrderDigraphJoinOfVertices: usage,\n",
+                  "the first argument <digraph> must be a partial order",
```
wilfwilson: As below, the word "order" should have a space after it here.

tst/standard/oper.tst
```diff
@@ -1791,6 +1791,60 @@ true
...
+
+# Testing error messages
+gap> gr1 := Digraph([[1],[2],[1,2,3],[1,2,4]]);
+<digraph with 4 vertices, 8 edges>
+gap> gr2 := DigraphReverse(gr1);
+<digraph with 4 vertices, 8 edges>
+
+# Meet does not exist
+gap> PartialOrderDigraphMeetOfVertices(gr1, 3, 4);
+Error, Digraphs: PartialOrderDigraphMeetOfVertices:
```
wilfwilson: These test outputs have not been updated - that's why the Travis tests are failing.

doc/oper.xml
```diff
@@ -1387,3 +1387,45 @@ gap> DigraphShortestDistance(gr, [1, 3], [3, 5]);
...
+<ManSection>
+  <Oper Name="PartialOrderDigraphMeetOfVertices" Label="for a digraph and two vertices"
+        Arg="digraph, u, v"/>
+  <Oper Name="PartialOrderDigraphJoinOfVertices" Label="for a digraph and two vertices"
+        Arg="digraph, u, v"/>
+  <Returns>A positive integer or <K>fail</K></Returns>
+    <Description>
+    If the first argument is a partial order digraph <Ref Prop="IsPartialOrderDigraph"/>
+    then these operations return the meet, or the join, of the two input vertices. If
+    the meet (or join) is does not exist or is not unique then <K>fail</K> is returned.
```
wilfwilson: Get rid of "or is not unique" since this is included in the definition of a meet/join.

doc/oper.xml
```diff
@@ -1387,3 +1387,45 @@ gap> DigraphShortestDistance(gr, [1, 3], [3, 5]);
...
+    The meet (or join) is guaranteed to exist and be unique when the first
```
wilfwilson: Get rid of "unique"

doc/prop.xml
```diff
@@ -8,6 +8,94 @@
...
+<digraph with 3 vertices, 5 edges>
+gap> IsPartialOrderDigraph(gr);
+true
+gap> gr := CycleDigraph(5);
+<digraph with 5 vertices, 5 edges>
+gap> IsPartialOrderDigraph(gr);
+false
+gap> gr := Digraph([[1, 1], [1, 1, 2], [3], [3, 3, 4, 4]]);
+<multidigraph with 4 vertices, 10 edges>
+]]></Example>
```
wilfwilson: Add a line that actually tests if this multidigraph is a partial order.

wilfwilson:
Hi @ChristopherRussell - thanks for making the changes. I've noticed a few more small things for you to change, but most importantly, you need to update the output of `oper.tst` since it is currently failing.

wilfwilson:
Good work @ChristopherRussell, and thanks for your patience with my numerous reviews! I think these are valuable functions to have in the package, and I'm happy to merge this.

PR #60, opened 2017-03-28
Add BooleanAdjacencyMatrixMutableCopy

wilfwilson:
We've got `OutNeighboursMutableCopy`, `InNeighboursMutableCopy`, and `AdjacencyMatrixMutableCopy`, so I thought I should add `BooleanAdjacencyMatrixMutableCopy`. Julius should be able to make use of it in PR #57.


james-d-mitchell:
Looks good.

Issue #62, opened 2017-03-28
DigraphColouring (greedy algorithm) ignores loops


wilfwilson:
`DigraphColouring` is documented as follows:

> A proper colouring of a digraph is a labelling of its vertices in such a way that adjacent vertices have different labels. A proper n-colouring is a proper colouring that uses exactly n colours. Equivalently, a proper (n-)colouring of a digraph can be defined to be a `DigraphEpimorphism` from a digraph onto the complete digraph.

I think this is fine. When given the second argument `n`, `DigraphColouring` tries to find an actual `n` colouring; it fails if the digraph has loops. However, the 1-argument version, which uses a greedy algorithm to find any semi-decent colouring, ignores loops, and returns what I believe to be incorrect colourings:
```
gap> gr := Digraph([[1, 2], [1, 2, 3], [2, 3]]);
<digraph with 3 vertices, 7 edges>
gap> DigraphColouring(gr);
Transformation( [ 1, 2, 1 ] )
gap> DigraphColouring(gr, 2);
fail
```
This should be nice and easy to fix.

james-d-mitchell:
Is this issue resolved? @wilfwilson

wilfwilson:
Yes, @mtorpey merged the PR last week.

PR #63, opened 2017-03-29
Fix DigraphColouring to return fail for digraphs with loops (Issue #62)

wilfwilson:
Currently, the one-argument version of `DigraphColouring` returns a colouring for a digraph, even if it has loops. This was inconsistent, and became issue #62.

This commit resolves issue #62.


mtorpey:
Looks great!

PR #64, opened 2017-04-08
cliques: fix error message

james-d-mitchell:
Previously the error referred to the fourth argument <hook>, when it
should be the second argument.


james-d-mitchell:
This should go into master and stable-0.7 if ok.

james-d-mitchell:
I'll do this later today, you are correct, of course.

james-d-mitchell:
@wilfwilson I've rebased and fixed the tests.

PR #65, opened 2017-04-10
InitLibrary: use InitGVarFuncsFromTable, avoid NewFunctionC

fingolfin:


james-d-mitchell:
@fingolfin thanks for the PR. Can you say a couple of words about the purpose of this change? 

Also it looks like travis failed because of the code formatting in the change you made, if you run `clang-format` on the file with the changes, this should allow the tests to pass.

fingolfin:
The PR changes Digraphs to use the recommend way of registering kernel function on the GAP level. This is shorter, and also makes it possible for us to revamp how that is done in the future by changing a single function, as opposed to every single package out there.

I run `make format` now to fix the formatting.

fingolfin:
The tests now fail because some `curl` invocation fails. 

Perhaps your scripts could use `set -ex` instead of just `set -e`, that would make debugging failures easier. I just add a HACK commit to this PR to do just, to see what's going on.

fingolfin:
OK, seems the build failure was a temporary hickup, so I removed the HACK commit and restarted the build.

james-d-mitchell:
Thanks Max!

wilfwilson:
Thanks @fingolfin

PR #66, opened 2017-04-12
Add functions for matchings

AlanMT:
Includes functions recognizing matchings, perfect matchings, and maximal matchings, where a subset of the symmetric closure of the edge set is a matching iff no two distinct edges in the edge set intersect at a vertex.

### DIFF
--- a/doc/oper.xml
+++ b/doc/oper.xml
@@ -1457,3 +1457,53 @@ true
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="IsMatching">
+<ManSection>
+  <Oper Name="IsMatching" Arg="digraph, edges"/>
+  <Oper Name="IsMaximalMatching" Arg="digraph, edges"/>
+  <Oper Name="IsPerfectMatching" Arg="digraph, edges"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+
+    A matching <C>M</C> on a digraph <C>G</C> is a
+    subset of the edges of the symmetric closure of G, such that no two 
+    distinct edges are adjacent.  Note that this definition allows matchings with loops,
+    and matchings with edges that are not themselves in the edge set of G,
+    but whose reverse is in the edge set of G. See <Ref Attr="DigraphSymmetricClosure" /> and
+    <Ref Prop="DigraphHasLoops" />. <P/>
+
+    A maximal matching N of a digraph G is a matching of G such that adding
+    any edge of G to N means that N is no longer a matching.  A perfect
+    matching P of a digraph G is a matching of G such that every vertex of
+    G is included in some edge in P. Every perfect matching is maximal. <P/>
+
+    If <A>digraph</A> is a digraph and <A>edges</A> is a list of pairs of vertices of digraph,
+    then <C>IsMatching(<A>digraph</A>, <A>edges</A>)</C> returns <K>true</K> if <A>edges</A>
+    defines a matching of <A>digraph</A>, and <K>false</K> 
+    otherwise.  Similarly, <C>IsMaximalMatching</C> returns <K>true</K> if <A>edges</A> defines a maximal
+    matching of <A>digraph</A>, and <C>IsPerfectMatching</C> returns 
+    <K>true</K> if <A>edges</A> defines a perfect matching of <A>digraph</A>.
+
+    <Example><![CDATA[
+gap> gr := Digraph([[2], [1], [2, 3, 4], [3, 5], [1]]);
+<digraph with 5 vertices, 8 edges>
+gap> IsMatching(gr, [[2, 1], [3, 2]]);
+false
+gap> edges := [[3, 2]];;
+gap> IsMatching(gr, edges);
+true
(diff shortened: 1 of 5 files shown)


doc/oper.xml
```diff
@@ -1457,3 +1457,53 @@ true
...
+<ManSection>
+  <Oper Name="IsMatching" Arg="digraph, edges"/>
+  <Oper Name="IsMaximalMatching" Arg="digraph, edges"/>
+  <Oper Name="IsPerfectMatching" Arg="digraph, edges"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+
+    A matching <C>M</C> on a digraph <C>G</C> is a
+    subset of the edges of the symmetric closure of <C>G</C>, such that no two 
+    distinct edges are adjacent.  Note that this definition allows matchings with loops,
```
wilfwilson: We talk about edges being 'incident' rather than 'adjacent'.

gap/oper.gd
```diff
@@ -97,3 +97,7 @@ DeclareOperation("PartialOrderDigraphJoinOfVertices",
 DeclareOperation("PartialOrderDigraphMeetOfVertices",
 [IsDigraph, IsPosInt, IsPosInt]);
 DeclareOperation("DigraphClosure", [IsDigraph, IsPosInt]);
+
+DeclareOperation("IsMatching", [IsDigraph, IsList]);
```
wilfwilson: These three things should be declared for `IsHomogeneousList` rather than `IsList`.

gap/oper.gi
```diff
@@ -1771,3 +1771,106 @@ function(digraph, k)
...
+  for e in edges do
+    #If edge set is not a subset of graph, not a matching
+    if not e in gredges and not Reversed(e) in gredges then
+      return false;
+    fi;
+    #Take e from prob
+    Remove(prob, Position(prob, e));
+    # Make sure intersection of edge with all other edges not yet checked is
+    # empty
+    for f in prob do;
```
wilfwilson: You don't need a semicolon after a `do`

gap/oper.gi
```diff
@@ -1771,3 +1771,106 @@ function(digraph, k)
...
+function(gr, edges)
+
+  local e, eset;
+
+  #If the edge set is not a matching, it's not a perfect matching
+  if not IsMatching(gr, edges) then
+    return false;
+  fi;
+
+  # Edge set will contain all vertices covered by an edge in the potential
```
wilfwilson: Would it be valid for this function, instead of doing what it doesn, at this point instead just checking:

`return Length(edges) = DigraphNrVertices(gr) / 2;`

?

gap/oper.gi
```diff
@@ -1771,3 +1771,106 @@ function(digraph, k)
...
+  local e, graph_less_matching, revedges;
+
+  #Check it's a matching
+  if not IsMatching(gr, edges) then
+    return false;
+  fi;
+
+  #Form the symmetric closure of the edge set
+  revedges := [];
+  for e in revedges do
```
wilfwilson: This loop will never be triggered since `revedges` is empty at this point...

gap/oper.gi
```diff
@@ -1771,3 +1771,106 @@ function(digraph, k)
...
+    Append(revedges, [Reversed(e)]);
+  od;
+
+  #This the graph without the edges corresponding to the matching
+  graph_less_matching := DigraphRemoveEdges(gr, edges);
+  graph_less_matching := DigraphRemoveEdges(graph_less_matching, revedges);
```
wilfwilson: This doesn't do anything since `revedges` is empty.

gap/oper.gi
```diff
@@ -1771,3 +1771,106 @@ function(digraph, k)
...
+
+  #Check to see that each edge, when added to the matching, results in a
+  #non-matching
+  for e in DigraphEdges(graph_less_matching) do
+    if IsMatching(gr, Concatenation(edges, [e])) then
+      return false;
+    fi;
+  od;
+  return true;
```
wilfwilson: If `v` is the set of vertices that are involved in the matching of `gr`, is it true that the matching is maximal if and only if the induced subdigraph of `gr` on the vertices *not* in `v` is empty?

gap/oper.gi
```diff
@@ -1771,3 +1771,106 @@ function(digraph, k)
...
+  local e, f, prob, gredges;
+
+  #gredges will contain the edges of gr
+  gredges := DigraphEdges(gr);
+
+  # Prob will hold the edges in our edge set we haven't yet checked
+  # (ie the potential problem edges)
+  prob := ShallowCopy(edges);
+
+  for e in edges do
```
wilfwilson: So essentially, this method checks whether each 'edge' in `edges` is an edge of the digraph (or the reverse of an edge), and then checks that no vertex appears in more than one edge?

wilfwilson:
I will make these changes.

gap/oper.gi
wilfwilson: No, because of loops.

wilfwilson:
This PR has been replaced by PR #71

PR #67, opened 2017-05-08
Don't set StructInitInfo members filename and isGapRootRelative

fingolfin:
Note that GAP always overrides these values anyway. And right now I am working on a patch to remove them in GAP, which would arrive in GAP 4.9. To be more precise, I am not removing but moving them to another struct, in order to not expose this internal implementation detail needlessly to kernel extensions.

The one drawback of this I see is that it might trigger some warnings when compiling Digraph.

### DIFF
--- a/src/digraphs.c
+++ b/src/digraphs.c
@@ -2689,9 +2689,8 @@ static StructInitInfo module = {
     /* checkInit   = */ 0,
     /* preSave     = */ 0,
     /* postSave    = */ 0,
-    /* postRestore = */ 0,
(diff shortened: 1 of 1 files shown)


james-d-mitchell:
@fingolfin thanks! Will the compiler warnings disappear in the future, when the changes you are released in GAP 4.9?

fingolfin:
Yes, because those struct members will be gone.

That change will hopefully be "soon" in master (I'll finish the PR today, and submit, then we'll have to see :)

james-d-mitchell:
Great, thanks @fingolfin

PR #70, opened 2017-06-23
attr: add ArticulationPoints and IsBiconnectedDigraph

james-d-mitchell:
This PR adds two features: the ability to compute `ArticulationPoints` and as a consequence check if a digraph is biconnected. This uses a pretty straightforward depth first search originally by Tarjan.

### DIFF
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -8,6 +8,42 @@
 #############################################################################
 ##
 
+<#GAPDoc Label="ArticulationPoints">
+<ManSection>
+  <Attr Name="ArticulationPoints" Arg="digraph"/>
+  <Returns>A list of vertices.</Returns>
+  <Description>
+
+    A connected digraph is <E>biconnected</E> if it is still connected (in the
+    sense of <Ref Prop="IsConnectedDigraph"/>) when any vertex is removed. 
+    If the digraph <A>digraph</A> is not biconnected but is connected, then any
+    vertex <C>v</C> of <A>digraph</A> whose removal makes the resulting digraph
+    disconnected is called an <E>articulation point</E>.<P/>
+
+    <C>ArticulationPoints</C> returns a list of the articulation points of
+    <A>digraph</A>, if any, and, in particular, returns the empty list if
+    <A>digraph</A> is not connected. <P/>
+
+    Multiple edges and loops are ignored by this method. <P/>
+    
+    The method used in this operation has complexity <M>O(m+n)</M> where
+    <M>m</M> is the number of edges (counting multiple edges as one, and not
+    counting loops) and <M>n</M> is the number of vertices in the digraph.
+
+    See also <Ref Attr="IsBiconnectedDigraph"/>.
+<Example><![CDATA[
+gap> ArticulationPoints(CycleDigraph(5));
+[  ]
+gap> ArticulationPoints(Digraph([[2, 7], [3, 5], [4], [2], [6], [1], []]));
+[ 2, 1 ]
+gap> ArticulationPoints(ChainDigraph(5));
+[ 4, 3, 2 ]
+gap> ArticulationPoints(NullDigraph(5));
+[  ]]]></Example>
+  </Description>
+</ManSection>
+<#/GAPDoc>
+
 <#GAPDoc Label="ChromaticNumber">
 <ManSection>
   <Attr Name="ChromaticNumber" Arg="digraph"/>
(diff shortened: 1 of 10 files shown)


james-d-mitchell:
The tests failed because I used a 2 argument lambda in the test file. Fixed it now.

doc/attr.xml
```diff
@@ -8,6 +8,42 @@
...
+    <A>digraph</A>, if any, and, in particular, returns the empty list if
+    <A>digraph</A> is not connected. <P/>
+
+    Multiple edges and loops are ignored by this method. <P/>
+    
+    The method used in this operation has complexity <M>O(m+n)</M> where
+    <M>m</M> is the number of edges (counting multiple edges as one, and not
+    counting loops) and <M>n</M> is the number of vertices in the digraph.
+
+    See also <Ref Attr="ArticulationPoints"/>.
```
mtorpey: This is a reference to itself.  Perhaps you meant `See also <Ref Attr="IsBiconnectedDigraph"/>.`

doc/attr.xml
```diff
@@ -8,6 +8,42 @@
...
 ##
 
+<#GAPDoc Label="ArticulationPoints">
+<ManSection>
+  <Attr Name="ArticulationPoints" Arg="digraph"/>
+  <Returns>A list of vertices.</Returns>
+  <Description>
+
+    A connected digraph is <E>biconnected</E> if it is still connected (in the
+    sense of <Ref Prop="IsConnectedDigraph"/>) if any vertex is removed.
```
mtorpey: Replace `if any vertex` by `when any vertex` to avoid a slightly confusing double-if.

doc/prop.xml
```diff
@@ -8,6 +8,36 @@
 #############################################################################
 ##
 
+<#GAPDoc Label="IsBiconnectedDigraph">
+<ManSection>
+  <Prop Name="IsBiconnectedDigraph" Arg="digraph"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    A connected digraph is <E>biconnected</E> if it is still connected (in the
+    sense of <Ref Prop="IsConnectedDigraph"/>) if any vertex is removed.
```
mtorpey: Perhaps `when any vertex`, as above.

gap/attr.gi
```diff
@@ -8,6 +8,85 @@
...
+  counter     := 0;
+  visited     := BlistList([1 .. DigraphNrVertices(copy)], []);
+  num         := [];
+  low         := [];
+  parent      := [1];
+  points      := [];
+
+  v_stack     := [1];
+  w_stack     := [0];
+  depth       := 1;
```
mtorpey: Many of these lines, like `points      := [];`, would not pass our own linting checks if they were in the Semigroups package, due to multiple spaces.  I don't mind, but would you rather Digraphs conformed to gaplint standard?

gap/attr.gi
```diff
@@ -8,6 +8,85 @@
...
+# http://www.eecs.wsu.edu/~holder/courses/CptS223/spr08/slides/graphapps.pdf
+
+InstallMethod(ArticulationPoints, "for a digraph", [IsDigraph],
+function(digraph)
+  local copy, nbs, counter, visited, num, low, parent, points, v_stack,
+  w_stack, depth, 1st_root_nb, v, w, i;
+
+  if HasIsConnectedDigraph(digraph) and not IsConnectedDigraph(digraph) then
+    return [];
+  elif DigraphNrVertices(digraph) < 1 then
```
mtorpey: Minor point: this `<` could be replaced by `<=` to avoid a tiny little bit more work.

gap/attr.gi
james-d-mitchell: Whitespace before `:=` does not violate anything in gaplint, so also not in the Semigroups package. Indeed clang-format will line up assignments in consecutive lines...

james-d-mitchell:
I've made all the requested changes, except the thing about := which I don't believe violates any conventions we've used anywhere.

doc/attr.xml
james-d-mitchell: Good idea

doc/attr.xml
james-d-mitchell: Ooops

mtorpey:
Good stuff!  Good to know whitespace around `:=` is flexible - I had thought it was like `..` or other operators.

PR #72, opened 2017-07-12
Fix some anomalies in the doc

james-d-mitchell:
In particular, 
1. fixed some things pointed out by `DIGRAPHS_CheckManualConsistency`
2. removed the two sections for partial orders which each only contained a single man section (these are redeployed elsewhere)
3. move the doc for `DigraphBicomponents` to the same section as the connected components.


wilfwilson:
I should start running `DIGRAPHS_CheckManualConsistency` as part of the release process.

PR #73, opened 2017-07-12
Add more attributes for Cayley digraphs

james-d-mitchell:
Namely, `GroupOfCayleyDigraph`, `SemigroupOfCayleyDigraph`,
and `GeneratorsOfCayleyDigraph`, so that the digraph output by
`CayleyDigraph` can recover the semigroup and generating set
used to create the digraph.

### DIFF
--- a/doc/digraph.xml
+++ b/doc/digraph.xml
@@ -903,6 +903,12 @@ gap> CycleDigraph(123);
 
     If the optional second argument <A>gens</A> is not present, then the
     generators of <A>G</A> are used by default.
+
+    The digraph created by this operation belongs to the category <Ref
+      Filt="IsCayleyDigraph"/>, the group <A>G</A> can be recovered from the
+    digraph using <Ref Attr="GroupOfCayleyDigraph"/>, and the generators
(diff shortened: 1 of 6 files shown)


wilfwilson:
I like the idea of this, I'll review it soon.

james-d-mitchell:
So, the intention with `GeneratorsOfCayleyDigraph` was that it returns the generators used to create the Cayley graph whatever they are (semigroup generators or group generators). If they are not explicitly given when the Cayley graph is produced, then this defaults to `GeneratorsOfGroup` or `GeneratorsOfSemigroup`. Maybe I should reword the doc to reflect this better, got any suggestions @wi

wilfwilson:
That's what I thought. No suggestions, it's just one of those things you've got to make sure you're precise about. Since elsewhere in this PR you're adding stuff to do with semigroup Cayley digraphs, then I think it makes sense to make it explicit what happens when it's a digraph related to a semigroup rather than a group. Currently the doc about `GeneratorsOf...` just concerns groups.

PR #74, opened 2017-07-12
Add DotVertexLabelledDigraph

james-d-mitchell:
This commit changes `DotDigraph` from a mutable attribute to an attribute,
and introduces `DotVertexLabelledDigraph`. I don't know why `DotDigraph` was a mutable
attribute since it was never used as such. I guess the original intention was that this could be recomputed after something about the digraph changed, but this was not implemented and so it makes more sense if it is an attribute. Since vertex labels can change `DotVertexLabelledDigraph` must be an operation.

### DIFF
--- a/doc/display.xml
+++ b/doc/display.xml
@@ -90,19 +90,26 @@ gap> Splash(DotDigraph(RandomDigraph(4)));
 <#GAPDoc Label="DotDigraph">
   <ManSection>
     <Attr Name="DotDigraph" Arg="digraph"/>
+    <Oper Name="DotVertexLabelledDigraph" Arg="digraph"/>
     <Returns>A string.</Returns>
     <Description>
-      This function produces a graphical representation of the digraph
+      <C>DotDigraph</C> produces a graphical representation of the digraph
       <A>digraph</A>. Vertices are displayed as circles, numbered consistently
       with <A>digraph</A>. Edges are displayed as arrowed lines between
       vertices, with the arrowhead of each line pointing towards the range
       of the edge.<P/>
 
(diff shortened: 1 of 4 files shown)


doc/display.xml
```diff
@@ -120,7 +127,7 @@ gap> FileString("dot/k4.dot", DotDigraph(gr));
 
 <#GAPDoc Label="DotSymmetricDigraph">
   <ManSection>
-    <Attr Name="DotSymmetricDigraph" Arg="digraph"/>
+    <Oper Name="DotSymmetricDigraph" Arg="digraph"/>
```
wilfwilson: If `DotDigraph` is an `Attr`, shouldn't `DotSymmetricDigraph` also be an attribute (again, not mutable)?

gap/display.gd
```diff
@@ -1,12 +1,13 @@
...
 ##  Licensing information can be found in the README file of this package.
 ##
 #############################################################################
 ##
 
-DeclareAttribute("DotDigraph", IsDigraph, "mutable");
-DeclareAttribute("DotSymmetricDigraph", IsDigraph, "mutable");
+DeclareAttribute("DotDigraph", IsDigraph);
+DeclareOperation("DotVertexLabelledDigraph", [IsDigraph]);
+DeclareOperation("DotSymmetricDigraph", [IsDigraph]);
```
wilfwilson: Same comment, that I think this should be an attribute.

james-d-mitchell:
@wilfwilson I do not know why this is failing, any ideas @mtorpey ??

mtorpey:
Looks like another instance of the occasional issue with Travis retrieving packages from a PPA.  Restarting the offending test suite fixed it.

PR #75, opened 2017-07-17
Kernel fixes and cleanup

james-d-mitchell:
This PR fixes a critical bug in the kernel module relating to homomorphism finding, and cleans up  some other minor historical oddities in the kernel module.


james-d-mitchell:
This is based off master but should be based of stable-0.9 if you want to include it in a release of 0.9.1. That will require some rebasing.

james-d-mitchell:
I'm going to relabel this and make the PR into master to avoid any confusion.

wilfwilson:
We should probably just make `0.10` the next release anyway, since it'll contain these bugfixes and presumable some of your other PRs.

PR #76, opened 2017-07-17
Update DigraphsTestStandard/Extreme/ManualExamples

james-d-mitchell:
To be more compact and more informative, and more similar to what Semigroups does.


james-d-mitchell:
Tests failed because I was using `args..` from GAP 4.9, and because I messed one other thing up too.

wilfwilson:
It's quite hard to review this closely, since the diff is quite messy. But I like the idea and I'm pretty sure it's good, so I'll merge.

PR #77, opened 2017-07-17
Add CliqueNumber and use it to improve ChromaticNumber

james-d-mitchell:
This PR adds the attribute `CliqueNumber` and uses it to improve `ChromaticNumber`. Several further minor modifications are made to improve the performance of `ChromaticNumber`, these include 

* sorting the connected components by size from smallest to largest, since colouring smaller digraphs is quicker than larger ones, and information about the colourings of the smaller components can be used to determine the overall chromatic number more quickly

* using `DigraphColouring` to obtain an upper bound for the chromatic number

* only trying to colour the digraph with a number of colours between the clique number plus one, and the number of colours used by `DigraphColouring` minus 1, instead of all values from `3` to the number of vertices minus 1

* using the chromatic number of smaller connected components as a lower bound for the other connected components.

The added test took 30 seconds before this PR and only about 300ms after. 

This PR depends on PR #75.

### DIFF
--- a/gap/attr.gi
+++ b/gap/attr.gi
@@ -87,7 +87,8 @@ end);
 
 InstallMethod(ChromaticNumber, "for a digraph", [IsDigraph],
 function(digraph)
-  local nr, comps, chrom, n, subdigraph, comp, i;
+  local nr, comps, upper, chrom, tmp_comps, tmp_upper, n, comp, bound, clique,
+  cmp, c, i;
 
   nr := DigraphNrVertices(digraph);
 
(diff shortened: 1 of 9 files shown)


gap/attr.gi
```diff
@@ -128,18 +135,29 @@ function(digraph)
...
         fi;
       else
-        for i in [3 .. n - 1] do
-          if DigraphColoring(subdigraph, i) <> fail then
-            if i > chrom then
-              chrom := i;
+        upper := RankOfTransformation(DigraphColouring(subdigraph), nr);
+        if chrom = fail or upper > chrom then
+          upper := Minimum(upper, nr - 1);
+          lower := Maximum(chrom, CliqueNumber(subdigraph));
```
wilfwilson: If `chrom` is still `fail` at this point, then since `fail` is greater than any integer, this will set `lower = fail`, which you don't want.

wilfwilson: You should do `CliqueNumber(DigraphSymmetricClosure(subdigraph))` since this gives a higher upper bound in general.

gap/attr.gi
```diff
@@ -128,18 +135,29 @@ function(digraph)
...
+          if upper = lower then
+            # This line is hard to test since DigraphColouring is
+            # non-deterministic now
+            chrom := upper; # upper > chrom
+          else # lower < upper
+            chrom := upper - 1;
+            repeat
+              coloring := DigraphColouring(subdigraph, chrom);
+              chrom    := chrom - 1;
+            until coloring = fail or chrom = lower;
```
wilfwilson: I get the `until coloring = fail` bit. If `coloring = fail`, then the connected component can not be coloured with `chrom + 1` colours, but since we know that we can colour it with `chrom + 2` colours, `chrom + 2` must be the chromatic number of this connected component. And hence `chrom + 2` is how many colours we need to colour the whole graph, so far. Great.

I'm unsure about the other condition. If `chrom = lower`, then we know that we can colour the connected component with `lower + 1` colours, and that the connected component is allowed to use, or requires, `lower` colours. However, I think it's possible that we never actually test whether this connected component *can* in fact be coloured with `lower` colours - it might not be possible to do it with exactly `lower` colours. The logic of the current code seems to suggest that if the loop terminates because `chrom = lower`, then the connected component *can* be coloured using only `lower` colours. However I'm not sure that this is guaranteed.

I'll try and find an example.

gap/attr.gi
```diff
@@ -128,18 +135,29 @@ function(digraph)
...
           chrom := n;
         fi;
       else
-        for i in [3 .. n - 1] do
-          if DigraphColoring(subdigraph, i) <> fail then
-            if i > chrom then
-              chrom := i;
+        upper := RankOfTransformation(DigraphColouring(subdigraph), nr);
+        if chrom = fail or upper > chrom then
+          upper := Minimum(upper, nr - 1);
```
wilfwilson: The problem with `n` vs `nr` is causing this kind of bug:

`gr := DigraphFromGraph6String("KmKk~K??G@_@");`

This is giving a chromatic number of 11, though the correct answer is 4. That's because it's trying to colour one of its components (that has 4 vertices) using 10 colours, which returns `fail` because not all 10 colours can be used for 4 vertices! That's because `upper` was originally set to 11, from the greedy colouring.

gap/attr.gi
wilfwilson: Once I fix the other bugs (changing `nr` to `n`, and the initialisation bug with `lower` being equal to `fail`, being greater than any integer), this example highlights the bug I'm on about:

```
gr := DigraphFromGraph6String("KmKk~K??G@_@");
ChromaticNumber(gr);
DigraphColouring(gr, 3);
DigraphColouring(gr, 4);
```
The chromatic number is 4, but we're getting the answer 3. That's because the clique number is 3, which is our lower bound, and we've found using the homomorphisms code that we are able to colour it with 4 colours. Your code is then assuming that that means that it is 3-colourable.

It should be simple to fix.

gap/attr.gi
james-d-mitchell: Oops, I'll look at this again tomorrow

james-d-mitchell:
I rewrote this completely with @sriubas help, hopefully resolving all problems.

gap/attr.gi
```diff
@@ -103,47 +104,89 @@ function(digraph)
...
+        max := Maximum(OutDegreeSet(comp));
+        if not (chrom > n or max <= chrom or IsNullDigraph(comp) or
+                IsBipartiteDigraph(comp)) then
+          # If <comp> is null or bipartite or the chromatic number is known to
+          # be less than the current best overall lower bound for the chromatic
+          # number of <digraph>, then its chromatic number is 1 or 2, and,
+          # since the chromatic number of <digraph> is >= 3, this component can
+          # be ignored.
+          clique := CliqueNumber(comp);
+          if clique = max and clique > chrom then
```
james-d-mitchell: Agreed.

gap/attr.gi
```diff
@@ -103,47 +104,89 @@ function(digraph)
...
-      if IsCompleteDigraph(DigraphSymmetricClosure(subdigraph)) then
+  # Prepare a list of connected components of digraph whose chromatic number we
+  # do not yet know.
+  if IsConnectedDigraph(digraph) then
+    comps := [digraph];
+    upper := [RankOfTransformation(DigraphColoring(digraph), nr)];
+  else
+    tmp_comps := [];
+    for comp in DigraphConnectedComponents(digraph).comps do
+      n    := Length(comp);
```
james-d-mitchell: Yup, good idea

gap/attr.gi
```diff
@@ -103,47 +104,89 @@ function(digraph)
               # <digraph> has at least 2 vertices at this stage
   fi;
 
-  digraph := DigraphRemoveAllMultipleEdges(digraph);
+  # The chromatic number of <digraph> is at least 3 and at most nr - 1
```
james-d-mitchell: Moved it

gap/attr.gi
```diff
@@ -103,47 +104,89 @@ function(digraph)
...
-    fi;
-
-    if not IsNullDigraph(subdigraph)
-        and not IsBipartiteDigraph(subdigraph) then
-      if IsCompleteDigraph(DigraphSymmetricClosure(subdigraph)) then
+  # Prepare a list of connected components of digraph whose chromatic number we
+  # do not yet know.
+  if IsConnectedDigraph(digraph) then
+    comps := [digraph];
+    upper := [RankOfTransformation(DigraphColoring(digraph), nr)];
```
james-d-mitchell: Yup

james-d-mitchell:
I think I've resolved all the issues.

PR #78, opened 2017-07-20
Add .mailmap file

fingolfin:
This makes the output of `git shortlog` more useful. Try e.g.
  git shortlog -sne


james-d-mitchell:
Thanks @fingolfin, good idea!

PR #80, opened 2017-07-26
kernel: use compiler intrinsic popcnt

james-d-mitchell:
This PR replaces the function used by the macro`COUNT_TRUES_BLOCK` with a compiler intrinsic when compiling with GCC or clang.


james-d-mitchell:
Since the tests passed and this provides a bit of speed up, I think it should be merged now.

PR #89, opened 2017-10-11
Change the way that digraphs are objectified

wilfwilson:
In this PR, I change the names of the record components of digraphs to match their attribute names. eg `nrvertices -> DigraphNrVertices`. This change makes the code nicer, and combined with a change to how we objectify digraphs, it leads to reduced memory usage. Previously, if you had a mutable list of out-neighbours, and you called `Digraph` on that list, then the record components `!.OutNeighbours` and `!.nbs` were *not* identical objects, and hence we were using twice as much memory as necessary. There is now just the one copy.

I've also changed `DigraphType` to include the filter `HasDigraphNrVertices`, since every digraph does have its number of vertices. If you look at the documentation in GAP for `ObjectifyWithAttributes`, you can see why including this filter would be desirable in the long run.

### DIFF
--- a/gap/attr.gi
+++ b/gap/attr.gi
@@ -345,13 +345,7 @@ end);
 
 #
 
-InstallMethod(DigraphNrEdges, "for a digraph", [IsDigraph],
-function(gr)
-  if IsBound(gr!.DigraphNrEdges) then
-    return gr!.DigraphNrEdges;
-  fi;
-  return DIGRAPH_NREDGES(gr);
-end);
+InstallMethod(DigraphNrEdges, "for a digraph", [IsDigraph], DIGRAPH_NREDGES);
 
 #
 
--- a/gap/attr.gi
+++ b/gap/attr.gi
@@ -345,8 +345,13 @@ end);
 
 #
 
-InstallMethod(DigraphNrEdges, "for a digraph",
-[IsDigraph], DIGRAPH_NREDGES);
+InstallMethod(DigraphNrEdges, "for a digraph", [IsDigraph],
+function(gr)
+  if IsBound(gr!.DigraphNrEdges) then
+    return gr!.DigraphNrEdges;
+  fi;
+  return DIGRAPH_NREDGES(gr);
+end);
 
 #
 
@@ -384,16 +389,16 @@ InstallMethod(DigraphRange, "for a digraph",
 [IsDigraph],
 function(digraph)
   DIGRAPH_SOURCE_RANGE(digraph);
-  SetDigraphSource(digraph, digraph!.source);
-  return digraph!.range;
+  SetDigraphSource(digraph, digraph!.DigraphSource);
+  return digraph!.DigraphRange;
 end);
 
 InstallMethod(DigraphSource, "for a digraph",
 [IsDigraph],
 function(digraph)
   DIGRAPH_SOURCE_RANGE(digraph);
-  SetDigraphRange(digraph, digraph!.range);
-  return digraph!.source;
+  SetDigraphRange(digraph, digraph!.DigraphRange);
+  return digraph!.DigraphSource;
 end);
 
 #
--- a/gap/digraph.gd
+++ b/gap/digraph.gd
@@ -23,10 +23,6 @@ DeclareProperty("IsMultiDigraph", IsDigraph);
 
 BindGlobal("DigraphFamily", NewFamily("DigraphFamily", IsDigraph));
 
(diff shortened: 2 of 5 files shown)


wilfwilson:
@james-d-mitchell Here is the only "problem" that I've identified. This is what currently happens in the `master` branch:
```
gap> gr := Digraph(List([1 .. 5], x -> [1 .. 5]));;
gap> gr!.adj;
[ [ 1 .. 5 ], [ 1 .. 5 ], [ 1 .. 5 ], [ 1 .. 5 ], [ 1 .. 5 ] ]
gap> gr!.OutNeighbours;
[ [ 1 .. 5 ], [ 1 .. 5 ], [ 1 .. 5 ], [ 1 .. 5 ], [ 1 .. 5 ] ]
gap> IsMultiDigraph(gr);;
gap> gr!.adj;
[ [ 1, 2, 3, 4, 5 ], [ 1, 2, 3, 4, 5 ], [ 1, 2, 3, 4, 5 ], [ 1, 2, 3, 4, 5 ],
  [ 1, 2, 3, 4, 5 ] ]
gap> gr!.OutNeighbours;
[ [ 1 .. 5 ], [ 1 .. 5 ], [ 1 .. 5 ], [ 1 .. 5 ], [ 1 .. 5 ] ]
```
And this is what happens in this PR:
```
gap> gr := Digraph(List([1 .. 5], x -> [1 .. 5]));;
gap> gr!.OutNeighbours;
[ [ 1 .. 5 ], [ 1 .. 5 ], [ 1 .. 5 ], [ 1 .. 5 ], [ 1 .. 5 ] ]
gap> IsMultiDigraph(gr);;
gap> gr!.OutNeighbours;
[ [ 1, 2, 3, 4, 5 ], [ 1, 2, 3, 4, 5 ], [ 1, 2, 3, 4, 5 ], [ 1, 2, 3, 4, 5 ],
  [ 1, 2, 3, 4, 5 ] ]
```
What happens is that `IsMultiDigraph` calls the C function `IS_MULTI_DIGRAPH`. This runs through the out-neighbours of `gr`, and calls `PLAIN_LIST` on each entry, in order to avoid segfaults in subsequent code, which assume that the lists are plain lists. In `master`, whilst this changes `gr!.adj`, it doesn't change the user-facing `gr!.OutNeighbours`. In this PR, there's only the one copy, and it gets changed.

I don't think this is a bug as such, but it would be nice if this didn't happen.

The only solution I see is to change the C code where we call `PLAIN_LIST`, to either produce a copy of the range that is now a plist, or to get rid of `PLAIN_LIST` altogether, and adapt our code to accept ranges.

ChrisJefferson:
Just a general note, we are (eventually) going to ban PLAIN_LIST on immutable lists (not sure if these are immutable...), So it's worth trying to move away from using it if possible.

ChrisJefferson:
Just to give a bit more detail (I realise this isn't the place to do it really :) ) HPC-GAP assumes that immutable objects really are immutable, which requires that they won't change value, or representation, so we can access them in multiple threads at the same time. You can imagine how running `PLAIN_LIST` in one thread would upset another one.

james-d-mitchell:
This looks good, I'm in favour of allowing out neighbours to be plists or ranges, but perhaps that's for another PR?

wilfwilson:
I am favour of that too. But supporting ranges natively will require some work, and should be done in another PR.

gap/attr.gi
```diff
@@ -345,8 +345,13 @@ end);
 
 #
 
-InstallMethod(DigraphNrEdges, "for a digraph",
-[IsDigraph], DIGRAPH_NREDGES);
+InstallMethod(DigraphNrEdges, "for a digraph", [IsDigraph],
+function(gr)
+  if IsBound(gr!.DigraphNrEdges) then
```
james-d-mitchell: Is this check really necessary? Is it the case that setting the record component in `gr` does not set the attribute with the same name, but the other way around is true, i.e. setting/getting the attribute sets the record component?

wilfwilson: Yes that is the case. Setting the record component does not set the attribute, but setting the attribute sets the record component.

I can get rid of this check, and everything will happen in the C code.

wilfwilson:
@james-d-mitchell I've changed `DigraphNrEdges`.

james-d-mitchell:
Ok, the last thing is that shouldn't this go into `stable-0.10`, and not `master`?

wilfwilson:
Done.

PR #94, opened 2017-11-01
add VertexConnectivity

ffloresbrito:
This pull request adds an attribute called `VertexConnectivity` which for a graph G = (V, E) computes the least cardinality |S| of a subset S of V, such that G - S is either disconnected, is trivial, or has no vertices. Documentation and tests are included.

### DIFF
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -1352,3 +1352,38 @@ gap> HamiltonianPath(g);
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="VertexConnectivity">
+<ManSection>
+  <Attr Name="VertexConnectivity" Arg="digraph"/>
+  <Returns>An non-negative integer.</Returns>
+  <Description>
+    For a digraph <A>digraph</A> with set of vertices <C>V</C>, the attribute
+    <C>VertexConnectivity(<A>digraph</A>)</C> returns the least cardinality
+    <C>|S|</C> of a subset <C>S</C> of <C>V</C> such that their removal either
+    disconnects <A>digraph</A>, makes it trivial, or makes it have no vertices.
+    The alrorithm makes <C>O(n - d - 1 + d * (d - 1) / 2)</C> calls to a
+    max-flow algorithm wich itself has complexity <C>O(n ^ 2 * e)</C>, where
+    <C>n</C> is the number of vertices of <A>digraph</A>, and <C>e, d</C> are
+    the number of edges and the degree (respectively) of the symmetric closure
+    of the digraph that results by removing all multiple edges of
+    <A>digraph</A>.
+
+    <Example><![CDATA[
+gap> J := JohnsonDigraph(9, 2);
+<digraph with 36 vertices, 504 edges>
+gap> VertexConnectivity(J);
+14
+gap> D := Digraph([[2, 4, 5], [1, 4], [4, 7], [1, 2, 3, 5, 6, 7], [1, 4],
+>                  [4, 7], [3, 4, 6]]);
+<digraph with 7 vertices, 20 edges>
+gap> VertexConnectivity(D);
+1
+gap> T := Digraph([]);
+<digraph with 0 vertices, 0 edges>
+gap> VertexConnectivity(T);
+0
+]]></Example>
+  </Description>
+</ManSection>
+<#/GAPDoc>
--- a/gap/attr.gi
+++ b/gap/attr.gi
@@ -1582,3 +1582,119 @@ function(gr)
   od;
   return fail;
 end);
+
+InstallMethod(VertexConnectivity,
+"for a digraph",
+[IsDigraph],
+function(digraph)
+  local edges, edmondskarp, k, kappa, kappas, mindegv, Nv, outdegs, relabel,
+  sdigraph, set1v, x, y;
+
+  if DigraphNrVertices(digraph) = 0 then
+    return 0;
+  fi;
+
+  sdigraph := DigraphSymmetricClosure(DigraphRemoveAllMultipleEdges(digraph));
+  edges := DigraphEdges(sdigraph);
+  kappas := [DigraphNrVertices(digraph) - 1];
+
+  relabel := function(edge, source, sink, wvert)
+    local nedge, i;
+    nedge := [];
+
+    for i in [1, 2] do
+      if edge[i] = source then
+        nedge[i] := 1;
+      elif edge[i] = sink then
+        nedge[i] := 2;
+      else
+        nedge[i] := Position(wvert, edge[i]) * 2 + i;
+      fi;
+    od;
+    return nedge;
+  end;
+
+  kappa := function(source, sink, digraph, edges)
+    local wvert, nedges, i;
+    wvert := ShallowCopy(DigraphVertices(digraph));
+    Remove(wvert, Position(wvert, source));
+    Remove(wvert, Position(wvert, sink));
+    nedges := List(edges, x -> relabel(x, source, sink, wvert));
+
+    for i in [1 .. Size(wvert)] do
+      Add(nedges, [i * 2 + 2, i * 2 + 1]);
+    od;
+
+    return edmondskarp(DigraphByEdges(nedges), 1, 2);
+  end;
+
+  edmondskarp := function(network, source, sink)
+    local flow, predecessor, queue, i, current, flag, e, capacity, outn, n;
+    flag := true;
+    flow := 0;
+    outn := OutNeighbours(network);
+    capacity := List(outn, x -> List(x, y -> true));
+
+    while flag do
+      queue := [source];
+      predecessor := List(outn, x -> 0);
+
+      while not IsEmpty(queue) do
+        current := queue[1];
+        Remove(queue, 1);
+
+        n := 0;
+        for e in outn[current] do
+          n := n + 1;
+          if predecessor[e] = 0 and e <> source and capacity[current][n] then
+            predecessor[e] := [current, n];
+            Add(queue, e);
+          fi;
+        od;
+      od;
+
+      if predecessor[sink] <> 0 then
+        e := predecessor[sink];
+        while e <> 0 do
+          capacity[e[1]][e[2]] := false;
+          e := predecessor[e[1]];
+        od;
+        flow := flow + 1;
+      else
+        flag := false;
+      fi;
+    od;
+    return flow;
+  end;
+
+  outdegs := OutDegrees(sdigraph);
+  mindegv := Position(outdegs, Minimum(outdegs));
+  set1v := ShallowCopy(DigraphVertices(sdigraph));
+  Remove(set1v, mindegv);
+  Nv := OutNeighboursOfVertex(sdigraph, mindegv);
+
+  for x in set1v do
+    if not [mindegv, x] in edges then
+      k := kappa(mindegv, x, sdigraph, edges);
+      if k = 0 then
+        return 0;
+      else
+        AddSet(kappas, k);
+      fi;
+    fi;
+  od;
+
+  for x in [1 .. Size(Nv) - 1] do
+    for y in [x + 1 .. Size(Nv)] do
+      if not [Nv[x], Nv[y]] in edges then
+        k := kappa(Nv[x], Nv[y], sdigraph, edges);
+        if k = 0 then
+          return 0;
+        else
+          AddSet(kappas, k);
+        fi;
+      fi;
+    od;
+  od;
+  return Minimum(kappas);
+end);
--- a/doc/z-chap4.xml
+++ b/doc/z-chap4.xml
@@ -59,6 +59,7 @@
     <#Include Label="DigraphDegeneracy">
     <#Include Label="DigraphDegeneracyOrdering">
     <#Include Label="HamiltonianPath">
+    <#Include Label="VertexConnectivity">
   </Section>
   
   <Section><Heading>Cayley graphs of groups</Heading>
--- a/gap/attr.gd
+++ b/gap/attr.gd
@@ -70,6 +70,7 @@ DeclareOperation("DIGRAPHS_MaximalSymmetricSubdigraph", [IsDigraph, IsBool]);
 DeclareAttribute("UndirectedSpanningTree", IsDigraph);
 DeclareAttribute("UndirectedSpanningForest", IsDigraph);
 DeclareAttribute("HamiltonianPath", IsDigraph);
+DeclareAttribute("VertexConnectivity", IsDigraph);
 
 # AsGraph must be mutable for grape to function properly
 DeclareAttribute("AsGraph", IsDigraph, "mutable");
--- a/tst/standard/attr.tst
+++ b/tst/standard/attr.tst
@@ -1656,6 +1656,38 @@ gap> gr := DigraphAddEdges(DigraphAddVertex(CycleDigraph(600)),
 gap> HamiltonianPath(gr);
 fail
 
+#T# VertexConnectivity
+gap> D := CompleteDigraph(10);
+<digraph with 10 vertices, 90 edges>
+gap> VertexConnectivity(D);
+9
+gap> D := JohnsonDigraph(9, 2);
+<digraph with 36 vertices, 504 edges>
+gap> VertexConnectivity(D);
+14
+gap> D := Digraph([]);
+<digraph with 0 vertices, 0 edges>
+gap> VertexConnectivity(D);
+0
+gap> D := Digraph([[]]);
+<digraph with 1 vertex, 0 edges>
+gap> VertexConnectivity(D);
+0
+gap> D := Digraph([[2, 4, 5], [1, 4], [4, 7], [1, 2, 3, 5, 6, 7], [1, 4],
+>                  [4, 7], [3, 4, 6]]);
+<digraph with 7 vertices, 20 edges>
+gap> VertexConnectivity(D);
+1
+gap> D := Digraph([[2, 4, 5], [1, 3, 4], [4, 7], [1, 2, 3, 5, 6, 7], [1, 4],
+>                  [4, 7], [3, 4, 6]]);
+<digraph with 7 vertices, 21 edges>
+gap> VertexConnectivity(D);
+2
+gap> D := Digraph([[2, 3], [3, 5], [1, 2, 4], [2, 3], [3]]);
+<digraph with 5 vertices, 10 edges>
+gap> VertexConnectivity(D);
+2
+
 #T# DIGRAPHS_UnbindVariables
 gap> Unbind(adj);
 gap> Unbind(adj1);


doc/attr.xml
```diff
@@ -1352,3 +1352,38 @@ gap> HamiltonianPath(g);
...
+<#GAPDoc Label="VertexConnectivity">
+<ManSection>
+  <Attr Name="VertexConnectivity" Arg="digraph"/>
+  <Returns>An non-negative integer.</Returns>
+  <Description>
+    For a digraph <A>digraph</A> with set of vertices <C>V</C>, the attribute
+    <C>VertexConnectivity(<A>digraph</A>)</C> returns the least cardinality
+    <C>|S|</C> of a subset <C>S</C> of <C>V</C> such that their removal either
+    disconnects <A>digraph</A>, makes it trivial, or makes it have no vertices.
+    The alrorithm makes <C>O(n - d - 1 + d * (d - 1) / 2)</C> calls to a
```
wilfwilson: Typo: algorithm.

doc/attr.xml
```diff
@@ -1352,3 +1352,38 @@ gap> HamiltonianPath(g);
...
+    max-flow algorithm wich itself has complexity <C>O(n ^ 2 * e)</C>, where
+    <C>n</C> is the number of vertices of <A>digraph</A>, and <C>e, d</C> are
+    the number of edges and the degree (respectively) of the symmetric closure
```
wilfwilson: What is the degree of a graph?

ffloresbrito: The degree of a digraph is the minimum of all the in-degrees and out-degrees of all vertices in the graph. Should I have specified this definition?

wilfwilson: You could therefore replace "degree" by "the minimum out-degree" to avoid using this new notion.

gap/attr.gi
```diff
@@ -1582,3 +1582,119 @@ function(gr)
   od;
   return fail;
 end);
+
```
wilfwilson: Can you give some information about where this algorithm comes from?

doc/attr.xml
```diff
@@ -1352,3 +1352,38 @@ gap> HamiltonianPath(g);
...
+
+<#GAPDoc Label="VertexConnectivity">
+<ManSection>
+  <Attr Name="VertexConnectivity" Arg="digraph"/>
+  <Returns>An non-negative integer.</Returns>
+  <Description>
+    For a digraph <A>digraph</A> with set of vertices <C>V</C>, the attribute
+    <C>VertexConnectivity(<A>digraph</A>)</C> returns the least cardinality
+    <C>|S|</C> of a subset <C>S</C> of <C>V</C> such that their removal either
+    disconnects <A>digraph</A>, makes it trivial, or makes it have no vertices.
```
wilfwilson: `such that their removal either disconnects <A>digraph</A>, makes it trivial, or makes it have no vertices.`

I think this would be better phrased as

`such that the induced subdigraph of <A>digraph</A> on <C>V\S</C> is disconnected, or has at most one vertex.`

doc/attr.xml
```diff
@@ -1352,3 +1352,38 @@ gap> HamiltonianPath(g);
...
+    The algorithm makes <C>O(n - d - 1 + d * (d - 1) / 2)</C> calls to a
```
wilfwilson: Make this a new paragraph

doc/attr.xml
```diff
@@ -1352,3 +1352,38 @@ gap> HamiltonianPath(g);
...
+    max-flow algorithm wich itself has complexity <C>O(n ^ 2 * e)</C>, where
```
wilfwilson: Does this mean n^(2*e) or (n^2)*e? Maybe put in brackets to emphasise.

gap/attr.gi
```diff
@@ -1582,3 +1582,125 @@ function(gr)
...
+        k := kappa(Nv[x], Nv[y], sdigraph, edges);
+        if k = 0 then
+          return 0;
+        else
+          AddSet(kappas, k);
+        fi;
+      fi;
+    od;
+  od;
+  return Minimum(kappas);
```
wilfwilson: Since `kappas` is sorted, the minimum is its first element. So you could save time by doing `kappas[1];` here.

gap/attr.gi
```diff
@@ -1582,3 +1582,125 @@ function(gr)
...
+        nedge[i] := Position(wvert, edge[i]) * 2 + i;
+      fi;
+    od;
+    return nedge;
+  end;
+
+  kappa := function(source, sink, digraph, edges)
+    local wvert, nedges, i;
+    wvert := ShallowCopy(DigraphVertices(digraph));
+    Remove(wvert, Position(wvert, source));
```
wilfwilson: Replace `Position` by `PositionSorted`, since `wvert` is a sorted list, and it's much faster.

gap/attr.gi
```diff
@@ -1582,3 +1582,125 @@ function(gr)
...
+    Remove(wvert, Position(wvert, sink));
```
wilfwilson: Same here.

gap/attr.gi
```diff
@@ -1582,3 +1582,125 @@ function(gr)
...
+        od;
+        flow := flow + 1;
+      else
+        flag := false;
+      fi;
+    od;
+    return flow;
+  end;
+
+#The following lines are an implementation of Algorithm Eleven from
```
wilfwilson: Can you put a space after the # on these three lines?

wilfwilson (dismissed):

doc/attr.xml
```diff
@@ -1352,3 +1352,40 @@ gap> HamiltonianPath(g);
...
+    <C>|S|</C> of a subset <C>S</C> of <C>V</C> such that the induced subdigraph
+    of <A>digraph</A> on <C>V \ S</C> is disconnected, or has at most one
+    vertex. <P/>
+
+    The algorithm makes <C>O(n - d - 1 + d * (d - 1) / 2)</C> calls to a
+    max-flow algorithm wich itself has complexity <C>O((n ^ 2) * e)</C>, where
+    <C>n</C> is the number of vertices of <A>digraph</A>, and <C>e, d</C> are
+    the number of edges and the minimum out-degree (respectively) of the
+    symmetric closure of the digraph that results by removing all multiple edges
+    of <A>digraph</A>.
```
james-d-mitchell: Wouldn't it be more straightforward to say "the number edges and the minimum degree of the underlying undirected graph of <A>digraph</A>."?

gap/attr.gi
james-d-mitchell: Yes, indeed, I second @wilfwilson's comment here. You should give a link to wherever you read about this algorithm here, so that anyone looking at this can also read about it. You should probably also add a citation to the documentation, stating where this is taken from.

gap/attr.gi
```diff
@@ -1582,3 +1582,125 @@ function(gr)
...
+        nedge[i] := Position(wvert, edge[i]) * 2 + i;
+      fi;
+    od;
+    return nedge;
+  end;
+
+  kappa := function(source, sink, digraph, edges)
+    local wvert, nedges, i;
+    wvert := ShallowCopy(DigraphVertices(digraph));
+    Remove(wvert, PositionSorted(wvert, source));
```
james-d-mitchell: If `wvert` is sorted (which the `PositionSorted` here would indicate), then you can replace this with `RemoveSet(wvert, source)`. Same for the next line

gap/attr.gi
```diff
@@ -1582,3 +1582,125 @@ function(gr)
...
+
+# The following function is an implementation of the Edmonds-Karp algorithm with
+# some minor adjustments that take into account the fact that the capacity of
+# all edges is 1.
+  edmondskarp := function(network, source, sink)
+    local flow, predecessor, queue, current, flag, e, capacity, outn, n;
+    flag := true;
+    flow := 0;
+    outn := OutNeighbours(network);
+    capacity := List(outn, x -> List(x, y -> true));
```
james-d-mitchell: You could replace this with `capacity := List(outn, x -> BlistList(x, x));` which will use less memory than the previous line.

gap/attr.gi
```diff
@@ -1582,3 +1582,125 @@ function(gr)
...
+[IsDigraph],
+function(digraph)
+  local edges, edmondskarp, k, kappa, kappas, mindegv, Nv, outdegs, relabel,
+  sdigraph, set1v, x, y;
+
+  if DigraphNrVertices(digraph) = 0 then
+    return 0;
+  fi;
+
+  sdigraph := DigraphSymmetricClosure(DigraphRemoveAllMultipleEdges(digraph));
```
james-d-mitchell: I don't think this line should be necessary, since `digraph` anyway contains all the information about its symmetric closure.

gap/attr.gi
```diff
@@ -1582,3 +1582,125 @@ function(gr)
...
+  edges := DigraphEdges(sdigraph);
```
james-d-mitchell: It is probably better to avoid using `DigraphEdges` wherever possible, and instead use `DigraphOutNeighbours` (replacing any test `edge in DigraphEdges(sdigraph)` by `edge[1] in DigraphOutNeighbours(sdigraph)[edge[2]] or edge[2] in DigraphOutNeighbours(sdigraph)[edge[1]]`), or if you know you want to do lots and lots of checks if something is an edge in `sdigraph`, then you should so use `BooleanAdjacencyMatrix`.

james-d-mitchell: The time complexity of:

* `edge in DigraphEdges(sdigraph)` is big-O of the number of edges in `sdigraph`, 

* `edge[1] in DigraphOutNeighbours(sdigraph)[edge[2]] or edge[2] in DigraphOutNeighbours(sdigraph)[edge[1]]` is big-O of `Size(DigraphOutNeighbours(sdigraph)[edge[1]]) + Size(DigraphOutNeighbours(sdigraph)[edge[2]])` (or maybe log of this depending on `sdigraph`)

* `BooleanAdjacency(sdigraph)[edge[1]][edge[2]] or BooleanAdjacency(sdigraph)[edge[2]][edge[1]]` is constant.

gap/attr.gi
```diff
@@ -1582,3 +1582,125 @@ function(gr)
...
+    local nedge, i;
+    nedge := [];
+
+    for i in [1, 2] do
+      if edge[i] = source then
+        nedge[i] := 1;
+      elif edge[i] = sink then
+        nedge[i] := 2;
+      else
+        nedge[i] := Position(wvert, edge[i]) * 2 + i;
```
james-d-mitchell: If `wvert` is the set of vertices of `Digraph`, then `Position(wvert, edge[i])` is always `edge[i]` (right @wilfwilson) so this check is pointless.

gap/attr.gi
james-d-mitchell: Also copying the vertices is slow, and removing things from a list is also slow. I suggest you refactor this to instead have a boolean list which indicates which vertices are active.

gap/attr.gi
```diff
@@ -1582,3 +1582,125 @@ function(gr)
...
+    outn := OutNeighbours(network);
+    capacity := List(outn, x -> List(x, y -> true));
+
+    while flag do
+      queue := [source];
+      predecessor := List(outn, x -> 0);
+
+      while not IsEmpty(queue) do
+        current := queue[1];
+        Remove(queue, 1);
```
james-d-mitchell: This is probably quite costly in GAP, better not to use a queue of this type. Please refactor this to remove `Remove`

gap/attr.gi
```diff
@@ -1582,3 +1582,125 @@ function(gr)
...
+    return flow;
+  end;
+
+# The following lines are an implementation of Algorithm Eleven from
+# Abdol-Hossein Esfahanian's ``Connectivity Algorithms'' which can be found at
+# https://www.cse.msu.edu/~cse835/Papers/Graph_connectivity_revised.pdf
+  outdegs := OutDegrees(sdigraph);
+  mindegv := Position(outdegs, Minimum(outdegs));
+  set1v := ShallowCopy(DigraphVertices(sdigraph));
+  Remove(set1v, mindegv);
```
james-d-mitchell: Same comment.

james-d-mitchell (dismissed):
This looks quite reasonable, apart from some design decisions that are likely rather costly (in terms of time and space). I'd like to see the code in this PR refactored to remove these things.

wilfwilson:
@james-d-mitchell Did @ffloresbrito address your changes? Is @ffloresbrito still involved in any GAP-related activity?

wilfwilson:
@james-d-mitchell I have squashed and rebased this PR on the latest master branch, and fixed the CI errors, and pushed. Assuming that @ffloresbrito is no longer involved with this, we can now decide how to proceed with this PR (or whether just to just close it).

On the whole, I'd say it's better to have a slow implementation than no implementation - it could always be a student project to improve or rewrite it. But that assumes that this implementation is correct, and it's been long enough that I don't remember if it is. So I'll review it again sometime.

wilfwilson (commented):
I've gone through the specific changes that James and I requested, and everything (bar the comment above) was resolved. So I've marked the comments as such.

I think we can assume that @ffloresbrito is no longer involved with this PR. I'm going to give it a proper review, and if it requires only small changes to make it acceptable, then I'll do this and get the PR merged.

gap/attr.gi
```diff
@@ -2613,3 +2613,146 @@ function(D)
...
+  od;
+
+  mindegv := 0;
+  mindeg  := DigraphNrVertices(digraph) + 1;
+  for i in DigraphVertices(digraph) do
+    if degs[i] < mindeg then
+      mindeg  := degs[i];
+      mindegv := i;
+    fi;
+  od;
```
MeikeWeiss, line 2729: I think mindegv:=PositionMinimum(degs) and mindeg:=degs[mindegv] should do the same and is an easier computation and easier to read.

gap/attr.gi
```diff
@@ -2613,3 +2613,146 @@ function(D)
...
+  mat  := BooleanAdjacencyMatrix(digraph);
+  degs := ListWithIdenticalEntries(DigraphNrVertices(digraph), 0);
+  for i in DigraphVertices(digraph) do
+    for j in [i + 1 .. DigraphNrVertices(digraph)] do
+      if mat[i][j] or mat[j][i] then
+        degs[i] := degs[i] + 1;
+        degs[j] := degs[j] + 1;
+      fi;
+    od;
+  od;
```
MeikeWeiss, line 2720: It should be also possible to compute degs with OutDegrees and InDegrees such that it is maybe shorter and easier to read.

MeikeWeiss (commented):
For us this method would be very helpful. I looked at the algorithm and the implementation and to me it looks like a useful implementation of vertex connectivity.
In my opinion two parts are written a bit awkwardly as you can see below. Is there anything else that needs to be done apart from adding a citation to the documentation?

wilfwilson:
The method currently seems to have a bug:
```gap
testing: /tmp/gaproot/pkg/Digraphs/tst/standard/attr.tst
########> Diff in /tmp/gaproot/pkg/Digraphs/tst/standard/attr.tst:
3152
# Input is:
VertexConnectivity(D);
# Expected output:
4
# But found:
3
########
    4903 ms (366 ms GC) and 1.16GB allocated for attr.tst
```
The correct answer is indeed 4 (as proved by brute force, according to some additional lines that I've since added to the test).

reiniscirpons:
Ok I think I fixed the bug. The new `DigraphMaximumFlow` function made it quite easy, the actual function is not too long, but I included quite a lot of verbose comments to justify what is going on. Happy to remove later on.

There are still a few questions that should be resolved (currently down as TODOs in the source), namely:
1. Should we rename the function from `VertexConnectivity` to `DigraphVertexConnectivity`? This would be more in line with other digraphs functions.
2. Should we change the convention for vertex-connectivity of the complete graphs? At the moment we follow the wikipedia definition (https://en.wikipedia.org/wiki/Connectivity_(graph_theory)), see also (https://mathworld.wolfram.com/VertexConnectivity.html), which imply that `VertexConnectivity(CompleteDigraph(n)) == n-1`. This has the potentially confusing consequence that for the singleton digraph, `VertexConnectivity(Digraph([[]])) == 0` but at the same time `IsConnectedDigraph(Digraph([[]])) == true`. Similarly, because of the convention, `VertexConnectivity(CycleDigraph(2)) == 1` and yet `IsBiconnectedDigraph(CycleDigraph(2)) == true`. The reason we need a convention in the first place is that the simple definition for vertex connectivity - "the least set of vertices whose removal disconnects the digraph" - works for all digraphs except for the complete ones. This is because no set of vertices you remove from a complete digraph disconnects it, unless you count the empty digraph as disconnected (which has its own set of issues). To quote the MathWorld article the convention `VertexConnectivity(CompleteDigraph(n)) == n-1` "allows most general results about connectivity to remain valid on complete graphs (West 2001, p. 149)". Other reasonable choices would be `VertexConnectivity(CompleteDigraph(n)) == n` (this would implicitly make us assume that the empty digraph is not connected) or `VertexConnectivity(CompleteDigraph(n)) == infinity` (indicating that no set of vertices disconnects the graph, but of course now you need to deal with infinities in your computations). At the moment I take the `n-1` convention and just explain the issues with connectivity and biconnectivity for the singleton and 2-cycle in the docs.
3. To speed up the computation we can remove loops and multiple edges, furthermore we need to symmetrize the graph. At the moment I check if the graph has loops, multiple edges or is not symmetric and if either condition holds then I do all three modifications if any condition holds. In theory I could also test each separately and then perform the change or somehow incorporate all of these modifications when constructing the `doubled_digraph` digraph. See line 3386 in attr.gi. Might also be premature optimizations, this probably doesn't take too much time when compared to the max-flow computations. Should we optimize these graph modifications?
4. I need to check adjacency for quite a lot of vertices (we only run a max-flow on non-adjacent vertices), at worst I will need to look at every pair of vertices, I'm currently using `DigraphAdjacencyFunction(digraph)` for this, is there a better way of checking for adjacency?

@wilfwilson could you give the code a review, when you have the chance? Thanks!

gap/attr.gi
```diff
@@ -3358,6 +3358,157 @@ function(D)
...
+  # In particular, we reduce the number of local vertex connectivity
+  # computations to n-d-1 + d*(d-1)/2 where n is the total number of vertices
+  # and d is the minimum degree of any vertex.
+  v := PositionMinimum(OutDegrees(digraph));
+  neighbours_v := OutNeighboursOfVertex(digraph, v);
+  # TODO: is adjacency function the best way to check adjacency?
+  adjacency_function := DigraphAdjacencyFunction(digraph);
+  kappa_min := fail;
+  for u in DigraphVertices(digraph) do
+    if u <> v and not adjacency_function(v, u) then
```
wilfwilson, line 3483: Does the `u <> v` check here mean that you can avoid removing all loops above? I haven't checked.

Also, I think the most efficient thing you can do in general is simply:
```suggestion
    if u <> v and not u in neighbours_v then
```

gap/attr.gi
```diff
@@ -3358,6 +3358,157 @@ function(D)
...
+        kappa_min := kappa;
+      fi;
+    fi;
+  od;
+
+  for i in [1 .. Length(neighbours_v)] do
+    for j in [i + 1 .. Length(neighbours_v)] do
+      u := neighbours_v[i];
+      v := neighbours_v[j];
+      if not adjacency_function(v, u) then
```
wilfwilson, line 3495: Probably the best you can do here is:
```suggestion
      if not IsDigraphEdge(D, v, u) then
```
Although I'm not totally sure.

But this change would let you get rid of the line 
```gap
adjacency_function := DigraphAdjacencyFunction(digraph);
```
above.

wilfwilson (approved):
This is looking good, thank you for working on it.

Perhaps to avoid any weirdness we could only define vertex connectivity for digraphs with at least two vertices? (Or whatever the relevant number would be to avoid the weird cases).

I'm happy with the file having lengthy comments, they're harmless and potentially helpful!

I've got a few suggestions

wilfwilson:
@reiniscirpons In response to your specific questions above, here are my answers:
1. Yes
2. I don't mind. 
3. I think it'd be easy enough to do these separately:
```gap
multi := IsMultiDigraph(D);
loops := DigraphHasLoops(D);
nsymm := not IsSymmetricDigraph(D);

if multi or loops or nsymm then
  D := DigraphMutableCopy(D);
  if multi then
    DigraphRemoveAllMultipleEdges(D);
  fi;
  if loops then
    DigraphRemoveLoops(D);
  fi;
  if nsymm then
    DigraphSymmetricClosure(D);
  fi;
  MakeImmutable(D);
fi;
```
4. I commented about this directly on the code.

gap/attr.gi
reiniscirpons: Yep youre right, looking at the source, if the digraph does not have a particular adjacency function, then `DigraphAdjcencyFunction` actually just calls the `IsDigraphEdge` function under the hood, and otherwise, `IsDigraphEdge` does some clever checks to use a faster method if possible. So `IsDigraphEdge` should be best here and in the other place where `adjacency_function` was used.

reiniscirpons:
@wilfwilson thanks for the comments! I've made the changes requested, good catch on improving the adjacency checks and the changes to the loop and multiple edge removal logic. I renamed the method and variables as per  the convention in the database. I ended up not changing the behavior on complete graphs because its not all that confusing and is also the convention used in literature anyway, but I did update the docs to be more clear about the caveats for small graphs.

Should be good to merge if the tests pass unless someone else has anything to add!

PR #96, opened 2017-11-02
Autoconf

james-d-mitchell:
This PR introduces some minor changes in the build system:

1. An --enable-debug configuration option is added, and some deprecated functionality is removed
2. A make target `superclean` is added, that will remove all untracked files from the digraphs folder
3. `__builtin_popcnt` is properly detected during `./configure`
4. the `gap-debug.h` header is removed since it is no longer required.


james-d-mitchell:
Merge it!

Issue #97, opened 2017-11-03
Try using NautyTracesInterface as an alternative to bliss


sebasguts:
It would be nice to have different AutomorphismGroup algorithms play against each other. The [NautyTracesInterface](https://github.com/sebasguts/NautyTracesInterface) package aims to provide a C-level-link to nauty and traces.

It would be nice to have digraphs work with this, to see how the different algorithms work with each other.

james-d-mitchell:
I'd like to do this. 

What do you envision? Some sort of global variable or whatever that specifies whether to use bliss or nauty? Or a per digraph option, or what?

Or multithreaded code that actually runs both bliss and nauty and uses which ever one returns the answer faster?  

I don't think doing any of these will be all that hard, but I'd like to know what you think before starting @sebasguts @wilfwilson ?

sebasguts:
I think all of this :)

First, one could have a global switch indicating which should be used if nothing is specified: bliss, nauty, or traces.

Then, additionally, it might be useful to be able to set it per digraph.

Last, it would be cool if one could use all three in parallel, seeing what finishes first, and discard the other computations.

wilfwilson:
It would be nice to not rely on bliss, since it does not seem to be actively developed, whereas nauty/traces is. So even for that reason it would be nice to have this in place.

Personally, I've never been bothered by the performance of bliss, so I'm not really sure what I would find the most useful. All of your suggestions seem natural @james-d-mitchell.

Having a per-digraph option would be nice. Perhaps that would require some sort of options record like you use for semigroups (which we use to specify "acting/non-acting", for instance), and the default options record would specify the default behaviour.

I think the most important discussion would be: what is the default?

PR #99, opened 2017-11-17
Use NautyTracesInterface

james-d-mitchell:
This PR allows the use of Nauty-Traces via the NautyTracesInterface package in Digraphs. 

The documentation is missing, or out of date, but otherwise I think this is complete.


james-d-mitchell:
Looks like NautyTracesInterface doesn't work in 32 bit mode

wilfwilson:
How shall we resolve the problem with 32-bit mode? Certainly we don't want Travis to try and run tests involving NautyTracesInterface when in 32-bit mode. Is that all we need to do?

sebasguts:
I am currently looking at the 32-bit mode problems, and hopefully solve them in NautyTracesInterface.

james-d-mitchell:
This requires a bit more work before it can be merged, but seems to work ok now, thanks @sebasguts

james-d-mitchell:
This PR resolves Issue #97

gap/isomorph.gi
```diff
@@ -0,0 +1,556 @@
...
+##
+
+# This file contains methods using bliss or nauty, for computing isomorphisms,
+# automorphisms, and canonical labellings of digraphs.
+
+InstallGlobalFunction(DigraphsUseBliss,
+function()
+  # Just do nothing if NautyTracesInterface is not available.
+  if DIGRAPHS_NautyAvailable then
+    Info(InfoWarning,
```
wilfwilson: Why are you using `InfoWarning` here (and elsewhere) rather than `InfoDigraphs`?

doc/isomorph.xml
```diff
@@ -1,14 +1,95 @@
...
+gap> BlissAutomorphismGroup(JohnsonDigraph(5, 2));
+Group([ (3,4)(6,7)(8,9), (2,3)(5,6)(9,10), (2,5)(3,6)(4,7), (1,2)(6,8)
+(7,9) ])]]></Example>
+    </Description>
+  </ManSection>
+<#/GAPDoc>
+
+<#GAPDoc Label="NautyAutomorphismGroup">
+<ManSection>
+  <Attr Name="NautyAutomorphismGroup" Arg="digraph[, colours]"/>
```
wilfwilson: Is it true that nauty can only handle digraphs without multiple edges? If so, this should be documented in all the relevant places.

james-d-mitchell:
This was intentional, as I'd like to make a PR soon to withdraw multidigraphs, and as such I didn't think it was worth writing documentation for a feature, only to remove it almost immediately. What do you think @wilfwilson ?

wilfwilson:
That's fine.

james-d-mitchell:
So, merge or wait?

PR #101, opened 2017-11-22
Add IsChainDigraph

AshleyClayton:
Adds a method "IsChainDigraph" to prop, with related documentation and tests.

### DIFF
--- a/doc/prop.xml
+++ b/doc/prop.xml
@@ -8,6 +8,7 @@
 #############################################################################
 ##
 
+<#GAPDoc Label="IsChainDigraph">
 <ManSection>
   <Prop Name="IsChainDigraph" Arg="digraph"/>
   <Returns><K>true</K> or <K>false</K>.</Returns>
--- a/doc/prop.xml
+++ b/doc/prop.xml
@@ -19,8 +19,8 @@
       Oper="ChainDigraph"/>.<P/>
 
     A digraph is a <E>chain</E> if and only if it is a directed tree, in which
-    every vertex has out degree at most one; see <Ref Oper="IsDirectedTree"/>
-    and <Ref Oper="OutDegrees"/>.
+    every vertex has out degree at most one; see <Ref Prop="IsDirectedTree"/>
+    and <Ref Attr="OutDegrees"/>.
 
 <Example><![CDATA[
 gap> gr := Digraph([[1, 3], [2, 3], [3]]);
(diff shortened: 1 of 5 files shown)


gap/prop.gi
```diff
@@ -8,6 +8,17 @@
...
 ##
 
+InstallMethod(IsChainDigraph, "for a digraph", [IsDigraph],
+function(digraph)
+  if DigraphNrEdges(digraph) = 0 and DigraphNrVertices(digraph) = 1 then
+    return true;
+  else
+    return IsConnectedDigraph(digraph) and InDegreeSet(digraph) = [0, 1]
+           and OutDegreeSet(digraph) = [0, 1]
+           and IsCycleDigraph(digraph) = false;
```
james-d-mitchell: This could be `not IsCycleDigraph(digraph)` instead of `IsCycleDigraph(digraph) = false`.

doc/prop.xml
```diff
@@ -8,6 +8,36 @@
...
+  <Prop Name="IsChainDigraph" Arg="digraph"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    <C>IsChainDigraph</C> returns <K>true</K> if the digraph <A>digraph</A> is
+    isomorphic to the chain digraph with the same number of vertices as
+    <A>digraph</A>, and <K>false</K> if it is not; see <Ref
+      Oper="ChainDigraph"/>.<P/>
+
+    A digraph is a <E>chain</E> if and only if it is connected,
+    not a cycle and both the in and out degree of any vertex is at most one.
```
james-d-mitchell: So, in the method for `IsChainDigraph` you require `InDegreeSet(digraph) = [0, 1]` and `OutDegreeSet(digraph) = [0, 1]` which is not exactly the same as what you've written in the documentation. Can you correct this please?

AshleyClayton: Thanks for the suggestion! In the special case of `ChainDigraph(1)`, the in and out degree sets are `[0]`, but I think that still makes my documentation statement incorrect, so I will rewrite that to clarify that case :)

gap/prop.gi
```diff
@@ -8,6 +8,17 @@
 #############################################################################
 ##
 
+InstallMethod(IsChainDigraph, "for a digraph", [IsDigraph],
+function(digraph)
+  if DigraphNrEdges(digraph) = 0 and DigraphNrVertices(digraph) = 1 then
```
wilfwilson: You could replace this whole function by

```
return IsDirectedTree(digraph) and IsSubset([0, 1], OutDegreeSet(digraph));
```

PR #103, opened 2017-12-07
Fix bug in ArticulationPoints

### LINKED ISSUE #102: Bug in IsBiconnectedDigraphs
gap> D := Digraph([[2, 4, 5], [1, 4], [4, 7], [1, 2, 3, 5, 6, 7],
    >                  [1, 4], [4, 7], [3, 4, 6]]);
    <digraph with 7 vertices, 20 edges>
    gap> IsBiconnectedDigraph(D);
    true
    gap> H := DigraphRemoveVertex(D, 4);
    <digraph with 6 vertices, 8 edges>
    gap> IsConnectedDigraph(H);
    false

wilfwilson:
I don't deeply understand the `ArticulationPoints` algorithm, but I changed the one line that looked weird to me, to something that I thought made more sense, and it seems to fix the issue.

Resolves #102.

### DIFF
--- a/gap/attr.gi
+++ b/gap/attr.gi
@@ -70,7 +70,7 @@ function(digraph)
           w_stack[depth] := 0;
           break;
         elif parent[v] <> w and num[w] < low[v] then
-          low[v] := low[w];
+          low[v] := num[w];
         fi;
       fi;
     od;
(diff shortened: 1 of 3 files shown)


mtorpey:
I've gone through `ArticulationPoints` and I think I understand it.  `num[i]` is the position of vertex `i` in the order that vertices were encountered in a depth-first search.  `low[i]` is the lowest vertex that `i` has an edge to (lowest in the sense of `num[i]`) other than its parent.  When we find an edge from `v` to `w`, we therefore must set `low[v]` to be at most `num[w]`, to see how low `v` can get by following edges without its parent.  So @wilfwilson's change makes sense.

Furthermore, this fixes the bug shown, and the PR does everything it needs to.  I'm therefore merging this pull request.

PR #104, opened 2017-12-07
Use C99 init for StructInitInfo

fingolfin:


james-d-mitchell:
I think we should merge this, what say you @wilfwilson ?

Issue #105, opened 2018-01-22
Add a method for IsDigraphAutomorphism (for a digraph and a permutation)


ChristopherRussell:
This could be useful for performance reasons, particularly for `McAlisterTripleSemigroups`.

james-d-mitchell:
This is related to Issue #40.

wilfwilson:
This was done by PR #106 and 0.12 will be released imminently.

PR #106, opened 2018-01-24
add method IsDigraphAutomorphism

ChristopherRussell:
This operation takes a digraph and a permutation and returns true if the
permutation is an automorphism of the digraph. See also #105.

### DIFF
--- a/doc/isomorph.xml
+++ b/doc/isomorph.xml
@@ -811,3 +811,31 @@ gap> OutNeighbours(canon);
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="IsDigraphAutomorphism">
+<ManSection>
+  <Oper Name="IsDigraphAutomorphism" Label="for a digraph and permutation"
+    Arg="digraph, x"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description> 
+    This operation returns true if the permutation <A>x</A> is an automorphism
+    of the digraph <A>digraph</A>. An permutation <C>g</C> on the vertices of a
+    digraph <C>gr</C> is an automorphism if it satisfies: <C>(u, v)</C> is an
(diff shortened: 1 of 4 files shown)


gap/isomorph.gi
```diff
@@ -554,3 +554,10 @@ function(n, partition, method)
                 "<partition[i]> is the colour of vertex i; in the second\n",
                 "form, <partition[i]> is the list of vertices with colour i,");
 end);
+
+InstallMethod(IsDigraphAutomorphism,
+"for a digraph and a permutation",
+[IsDigraph, IsPerm],
+function(gr, x)
+  return OnDigraph(gr, x) = gr;
```
wilfwilson: Please first add a check for whether `gr` is a multidigraph, if so, then have an error message saying it doesn't work for digraphs with multiple edges.

doc/isomorph.xml
```diff
@@ -811,3 +811,30 @@ gap> OutNeighbours(canon);
...
+<ManSection>
+  <Oper Name="IsDigraphAutomorphism" Label="for a digraph and permutation"
+    Arg="digraph, x"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description> 
+    This operation returns true if the permutation <A>x</A> is an automorphism
+    of the digraph <A>digraph</A>. An permutation <C>g</C> on the vertices of a
+    digraph <C>gr</C> is an automorphism if it satisfies: <C>(u, v)</C> is an
+    edge of <C>gr</C> if and only if <C>(g(u), g(v))</C> is an edge of
+    <C>gr</C>. See also <Ref Attr="AutomorphismGroup" Label="for a digraph" />.
```
wilfwilson: Please change it to:

...of `<C>gr</C>` for all vertices <C>u</C> and <C>v</C> in <C>digraph</A>. (To quantify the vertices)

PR #108, opened 2018-01-27
Use IS_CONSTRUCTOR instead of CONSTRUCTORS

alex-konovalov:
CONSTRUCTORS has been removed from the GAP master branch: see https://github.com/gap-system/gap/pull/2114/commits/b597f5cbf18b89e9264d2fff4362c0c652d22270

### DIFF
--- a/gap/utils.gi
+++ b/gap/utils.gi
@@ -327,7 +327,7 @@ DIGRAPHS_ManSectionType := function(op)
     class := "Oper";
     if IS_IDENTICAL_OBJ(op, IS_OBJECT) then
       class := "Filt";
-    elif op in CONSTRUCTORS then
+    elif IS_CONSTRUCTOR(op) then
       class := "Constructor";
       # seem to never get one
     elif IsFilter(op) then


wilfwilson:
Thanks for this @alex-konovalov. You've replaced `op in CONSTRUCTORS` by `IS_CONSTRUCTOR(op)`, which makes sense. However, a problem is that GAP 4.8 does not contain `IS_CONSTRUCTOR`. See for instance:

```
Syntax warning: unbound global variable in /home/travis/gap/pkg/digraphs/gap/u\
tils.gi line 330
    elif IS_CONSTRUCTOR(op) then
```
from https://travis-ci.org/gap-packages/Digraphs/jobs/334208726#L2269

Can you see a way around this? I'm reluctant to have syntax warnings when using Digraphs with GAP 4.8.

alex-konovalov:
True... `IS_CONSTRUCTOR` is present in stable-4.9 branch. You may wish to wait with this until you will be updating Digraphs for GAP 4.9 then... Alternatively, you may have some compatibility code for GAP 4.8, calling `BIND_GLOBAL( "IS_CONSTRUCTOR", op -> op in CONSTRUCTORS );` but only if the version is less than 4.9.0

james-d-mitchell:
@alex-konovalov can you please update this PR with the second of your suggestions, adding 
`BIND_GLOBAL( "IS_CONSTRUCTOR", op -> op in CONSTRUCTORS );`

wilfwilson:
See #109 - @fingolfin suggests we simply remove the relevant code, since it is not used by our package. I suggest that we do that.

wilfwilson:
I'll close this since I have now removed `CONSTRUCTORS` entirely.

PR #118, opened 2018-04-26
Add IsDigraphHomomorphism and related operations

james-d-mitchell:
This PR adds new operations `IsDigraphHomomorphism`, `IsDigraphIsomorphism`, `IsDigraphMonomorphism`, `IsDigraphEpimorphism`, and `IsDigraphEndomorphism`. Also refactors `IsDigraphAutomorphism` and links the doc for `IsDigraphAutomorphism` into the manual for the first time. The arguments of any of these operations are digraph (or digraphs) and a perm or transformation. 

This breaks backwards compatibility with `IsDigraphAutomorphism`, which previously returned `true` for permutations that did not fix non-vertices. With this PR, the `IsDigraphAutomorphism` will return `true` if and only if the permutation belongs to the automorphism group, or the permutation defined by the transformation does.

### DIFF
--- a/doc/isomorph.xml
+++ b/doc/isomorph.xml
@@ -1,7 +1,7 @@
 #############################################################################
 ##
 #W  isomorph.xml                                        James D. Mitchell
-#Y  Copyright (C) 2014-17                                  Wilf A. Wilson 
+#Y  Copyright (C) 2014-18                                  Wilf A. Wilson 
 ##
 ##  Licensing information can be found in the README file of this package.
 ##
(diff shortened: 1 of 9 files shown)


doc/isomorph.xml
```diff
@@ -814,27 +814,59 @@ gap> OutNeighbours(canon);
...
+    <C>IsDigraphIsomorphism</C> returns <K>true</K> if the permutation or
+    transformation <A>x</A> is an isomorphism from the digraph <A>src</A> to
+    the digraph <A>ran</A>.
+    <P/>
+    
+    <C>IsDigraphAutomorphism</C> returns <K>true</K> if the permutation or
+    transformation <A>x</A> is an automorphism of the digraph <A>digraph</A>.
+    <P/>
+    
+    A permutation or transformation <A>x</A> is an <E>isomorphism</E> from a
```
wilfwilson: You should add that a transformation must be injective/bijective/surjective. With the definition you've got here, `Transformation([1, 1, 1])` would be an isomorphism from an empty digraph with 3 vertices to an empty digraph with 1 vertex.

gap/grahom.gi
```diff
@@ -430,3 +430,122 @@ function(gr1, gr2)
...
+
+InstallMethod(IsDigraphHomomorphism,
+"for digraph, digraph, and transformation",
+[IsDigraph, IsDigraph, IsTransformation],
+function(src, ran, x)
+  local i, j;
+  if IsMultiDigraph(src) or IsMultiDigraph(ran) then
+    ErrorNoReturn("Digraphs: IsDigraphHomomorphism: usage,\n",
+                  "the first 2 arguments must not have multiple edges,");
+  elif DegreeOfTransformation(x) > DigraphNrVertices(src) then
```
ChristopherRussell: Should the transformation which send 1 to 2 and fixes all other points be a digraph homomorphism between the edgeless digraph on 1 point and the edgeless digraph on 2 points? If so then there is an issue since `DegreeOfTransformation(Transformation([2,2])) = 2` but `DigraphNrVertices(Digraph([[]])) = 1` so this example will be rejected

ChristopherRussell: Could this be fixed by replacing `DegreeOfTransformation` with `LargestMovedPoint`?

james-d-mitchell: Good point @ChristopherRussell, I will fix this now.

PR #125, opened 2018-06-18
travis: update to test against GAP stable-4.9 instead of stable-4.8

flsmith:


james-d-mitchell:
@wilfwilson are you happy with this change? Do we need to test against GAP 4.8 any longer?

wilfwilson:
Yeah it’s fine to change from stable-4.8 to 4.9 as long as we also keep testing against the required version (4.8.2 I think), which we are still doing.

james-d-mitchell:
Thanks @wilfwilson

PR #128, opened 2018-07-04
Add DigraphDijkstraS and DigraphDijkstraST (WORK IN PROGRESS)

markuspf:
These compute shortest (weighted) distances in a digraph.

 * `DigraphDijkstraS` computes least cost distances and paths
   to all vertices reachable from a source vertex
 * `DigraphDijkstraST` computes the least cost distance between a source
   vertex and a target vertex. The only difference between this function
   and `DigraphDijkstraS` is that `DigraphDijkstraST` breaks early when the
   target vertex is found

### DIFF
--- a/PackageInfo.g
+++ b/PackageInfo.g
@@ -212,7 +212,8 @@ PackageDoc := rec(
 Dependencies := rec(
   GAP := ">=4.8.2",
   NeededOtherPackages := [["io", ">=4.5.1"],
-                          ["orb", ">=4.7.5"]],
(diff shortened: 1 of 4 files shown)


markuspf:
This pull-request should be viewed as a base for discussion how to (or whether) to add these functions to `digraphs`.

Note that this patch adds a dependency on `datastructures`, which I submitted for redistribution this morning.

james-d-mitchell:
Supercede

PR #131, opened 2018-09-13
Use full names instead of initials in PackageInfo.g

alex-konovalov:
This is the same kind of changes that in https://github.com/gap-packages/Semigroups/pull/505


james-d-mitchell:
I’m happy with this, @wilfwilson ?

PR #136, opened 2018-09-23
add method which converts semilattice digraphs to semigroups

ChristopherRussell:
This PR implements a method for `AsSemigroup` which take the filter `IsPartialPermSemigroup` and a digraph satisfying either `IsJoinSemilatticeDigraph` (or `IsMeetSemilatticeDigraph`) and returns a semigroup of partial perms isomorphic to this digraph with the join (or meet) operation on the vertices. It also implements a method for `AsMonoid` which functions identically but only accepts the filter `IsPartialPermMonoid` and digraphs satisfying `IsLatticeDigraph`.

See [Semigroups PR #529](https://github.com/gap-packages/Semigroups/pull/529#issuecomment-423505417) for the previous PR with this method.

Since making a PR for this in the semigroups package I have improved the part which determines a small generating set. I now just set the generating set to be all the vertices in the `DigraphReflexiveTransitiveReduction` whose set of `InNeighbours` has size 0 or 1, rather than adding vertices one at a time and seeing if the whole semigroup had been generated yet.


ChristopherRussell:
I am not sure how to reference `IsPartialPermSemigroup` and `IsPartialPermMonoid` in the documentation, could somebody help me out here? I tried `<Ref Func="IsPartialPermMonoid" BookName="ref"/>` and `<Ref Prop="IsPartialPermMonoid" BookName="ref"/>` but neither worked right.

doc/oper.xml
```diff
@@ -1503,3 +1503,47 @@ true
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="AsSemigroup">
+  <ManSection>
+    <Func Name="AsSemigroup" Arg="filt, digraph"/>
```
wilfwilson: These should be documented as an operation, since it is an operation, as should IsMonoid

doc/oper.xml
```diff
@@ -1503,3 +1503,47 @@ true
...
+    <Func Name="AsMonoid" Arg="filt, digraph"/>    
+    <Returns> A semilattice of partial perms. </Returns>
+    <Description>
+      The function <C>AsSemigroup</C> requires that filt be equal to
```
wilfwilson: function -> operation, and filt -> `<A>filt</A>`

doc/oper.xml
```diff
@@ -1503,3 +1503,47 @@ true
...
+      <A>digraph</A> with the binary operation
+      <Ref Oper="PartialOrderDigraphJoinOfVertices"/>. If <A>digraph</A>
+      satisfies  <Ref Prop="IsMeetSemilatticeDigraph"/> but not
+      <Ref Prop="IsJoinSemilatticeDigraph"/> then
+      <C>AsSemigroup</C> returns a semigroup of partial perms which is
+      isomorphic to the semigroup whose elements are the vertices of
+      <A>digraph</A> with the binary operation
+      <Ref Oper="PartialOrderDigraphMeetOfVertices"/>. <P/>
+
+      The function <C>AsMonoid</C> functions identically to <C>AsSemigroup</C>
```
wilfwilson: function -> operation
functions -> behaves
identically -> similarly

doc/oper.xml
```diff
@@ -1503,3 +1503,47 @@ true
...
+      but requires that filt be equal to
```
wilfwilson: but -> except that
filter -> `<A>filt</A>`

doc/oper.xml
```diff
@@ -1503,3 +1503,47 @@ true
...
+      <Ref Func="IsParitalPermMonoid" BookName="ref"/> and
```
wilfwilson: I think you should also allow the filter `IsPartialPermSemigroup` as well. Also, this is misspelled as Parital not Partial. . Also it's a `Filt` not a `Func`.

doc/oper.xml
```diff
@@ -1503,3 +1503,47 @@ true
...
+      that <A>digraph</A> satisfies <Ref Prop="IsLatticeDigraph"/>.
+      <Example><![CDATA[
```
wilfwilson: Perhaps mention that you return something with minimal generating set, and of minimal degree (if that is the case?)

gap/oper.gi
```diff
@@ -1833,3 +1833,60 @@ function(gr, edges)
...
+
+  # Determine a small generating set
+  gens := Filtered([1 .. n], j -> Size(InNeighboursOfVertex(red, j)) < 2);
+  return Semigroup(List(gens, g -> PartialPerm(im[g], im[g])));
+end);
+
+InstallMethod(AsMonoid,
+"for a function and a digraph", [IsFunction, IsDigraph],
+function(filt, digraph)
+  if not filt = IsPartialPermMonoid then
```
wilfwilson: Again, I think this should allow IsPartialPermSemigroup as well

gap/oper.gi
```diff
@@ -1833,3 +1833,60 @@ function(gr, edges)
...
+      ErrorNoReturn("Digraphs: AsMonoid usage,\n",
+                    "the first argument must be IsPartialPermMonoid,");
+  elif not IsLatticeDigraph(digraph) then
+      ErrorNoReturn("Digraphs: AsMonoid usage,\n",
+                    "the second argument must be a lattice digraph,");
+  fi;
+  return AsSemigroup(IsPartialPermSemigroup, digraph);
```
wilfwilson: Is this definitely guaranteed to return something in the category `IsMonoid`?

tst/standard/oper.tst
```diff
@@ -1979,6 +1979,83 @@ gap> edges := [[1, 1], [2, 3]];
...
+gap> IsIsomorphicDigraph(nat, di);
+true
+gap> di := DigraphFromDigraph6String("+P~~}~??o??_?@o??_?@_?W}?O_??o??_?@_FGoA?_?Wo?O_Wt_");;
+gap> S := AsSemigroup(IsPartialPermSemigroup, di);;
+gap> IsSemilattice(S);
+true
+gap> nat := DigraphReflexiveTransitiveClosure(Digraph(NaturalPartialOrder(S)));;
+gap> IsIsomorphicDigraph(nat, di);
+true
+gap> di := DigraphFromDigraph6String("+u~~~~~~~~~^~w??????Nw???????E????????A????????@_????????_???????AO????????g???????EE???????AA???????@`_???????__??????AQO???????gg??????NwF~~}???E?B~_??????@w????????o????????O????????W????????u????????Q????????Z?????A???_??????@w]???????oK???????OC???????WE???????uL_??????QC_???@_?ZEw????_???G???AO??_C????g???I???EEB~_@~o???@w??{?????o??W?????O??G?????W??K?????u??Z?????Q??H?????Z??L_?AA??_??O???@w]?{N????oK?WE????OC?GA????WE?KB????uL_ZEo???QC_HAO@`_ZEwLb[?__??G??CAQO?_C?OA?gg??I??D");;
```
wilfwilson: Can you try and make these lines a bit shorter? Doing
```
Concatention(".......",
             ".......",
```
etc

doc/oper.xml
```diff
@@ -1503,3 +1503,47 @@ true
...
 <#/GAPDoc>
+
+<#GAPDoc Label="AsSemigroup">
+  <ManSection>
+    <Func Name="AsSemigroup" Arg="filt, digraph"/>
+    <Func Name="AsMonoid" Arg="filt, digraph"/>    
+    <Returns> A semilattice of partial perms. </Returns>
+    <Description>
+      The function <C>AsSemigroup</C> requires that filt be equal to
+      <Ref Func="IsParitalPermSemigroup" BookName="ref"/>.
```
wilfwilson: This is misspelled as Parital not Partial. Also it's a `Filt` not a `Func`.

doc/oper.xml
```diff
@@ -1503,3 +1503,47 @@ true
...
+      If <A>digraph</A> is a
+      <Ref Prop="IsJoinSemilatticeDigraph"/> or <Ref Prop="IsLatticeDigraph"/>
+      then <C>AsSemigroup</C> returns a semigroup of partial perms which is
+      isomorphic to the semigroup whose elements are the vertices of
+      <A>digraph</A> with the binary operation
+      <Ref Oper="PartialOrderDigraphJoinOfVertices"/>. If <A>digraph</A>
```
wilfwilson: You need to add `Label="for a digraph and two vertices"` (without line breaks) to this, ie
```
 <Ref Oper="PartialOrderDigraphJoinOfVertices"
        Label="for a digraph and two vertices"/>
```

doc/oper.xml
```diff
@@ -1503,3 +1503,47 @@ true
...
+      satisfies  <Ref Prop="IsMeetSemilatticeDigraph"/> but not
+      <Ref Prop="IsJoinSemilatticeDigraph"/> then
+      <C>AsSemigroup</C> returns a semigroup of partial perms which is
+      isomorphic to the semigroup whose elements are the vertices of
+      <A>digraph</A> with the binary operation
+      <Ref Oper="PartialOrderDigraphMeetOfVertices"/>. <P/>
```
wilfwilson: You need to add Label="for a digraph and two vertices" (without line breaks) to this, ie
```
<Ref Oper="PartialOrderDigraphMeetOfVertices"
        Label="for a digraph and two vertices" />
```

wilfwilson:
@ChristopherRussell I think I've addressed your questions in my comments. Once they're addressed I think it'll be good to go.

doc/oper.xml
ChristopherRussell: The replacement 

> but -> except that

results in

> ...except that requires that...

wilfwilson: Yeah sorry I meant "except that it requires that `<A>filt</A>`..." – but yeah change it how you like, your suggestion sounds good, just make sure that you say somewhere that it returns a monoid!

gap/oper.gi
wilfwilson: The second one here won't work, since `IsMonoid` is a category and you can't change categories that way. To be safe you could go with the first one of those options, but if you look at the GAP code and convince yourself that it will always return something in the category `IsMonoid` given the generators that correspond to a lattice digraph, then I don't mind if you leave it how it is.

ChristopherRussell: So when I apply AsSemigroup to a lattice digraph it prints `<partial perm monoid ... >` and satisfies `IsMonoid`. However the function `GeneratorsOfMonoid` returns a list containing the identity of the monoid. I would need to remove that somehow to claim the generating set is of minimal degree in the documentation.

```
gap> digraph := Digraph([[1],[1,2],[1,2,3]]);
<digraph with 3 vertices, 6 edges>
gap> S := AsSemigroup(IsPartialPermSemigroup, digraph);
<partial perm monoid of rank 2 with 3 generators>
gap> IsMonoid(S);
true
gap> GeneratorsOfSemigroup(S);
[ <empty partial perm>, <identity partial perm on [ 1 ]>, <identity partial perm on [ 1, 2 ]> ]
gap> GeneratorsOfMonoid(S);
[ <empty partial perm>, <identity partial perm on [ 1 ]>, <identity partial perm on [ 1, 2 ]> ]
 ```

Comment (wilfwilson) [reply gap/oper.gi]: Yes, something is in the category `IsMonoid` if and only if `IsMonoid` returns `true`. It's just a technical thing: it means that for any element of the monoid you can call `One(x)` (or write `x^0`) and you get the identity element (or 'one') of that monoid. So you can have things that are technically semigroups with identity elements that do not satisfy `IsMonoid` in GAP.

So yeah, maybe it's a good idea to get the generators returned by `AsSemigroup(...)`, remove the identity, and then call `Monoid` on the resulting thing!

To clear something up – minimal degree is about the number of points on which the partial perms act on; minimal generating set means the least number of generators is used :)

Comment (ChristopherRussell) [reply gap/oper.gi]: Thanks for clearing that up for me. I will do as you suggest with the generators.

> To clear something up – minimal degree is about the number of points on which the partial perms act on; minimal generating set means the least number of generators is used :)

Oops, I did know that about degree vs generating set. I really meant to say size not degree in my last comment!

**Comment by ChristopherRussell** [comment]
I realised my tests all fail because they involve `IsSemilattice` and `NaturalPartialOrder` which are methods from the semigroups package. I guess I will need to remove all these aspects of the tests!

**Comment by james-d-mitchell** [comment]
I’m happy with this when @wilfwilson and @chrisrussell are

**Comment by ChristopherRussell** [comment]
I have addressed all of @wilfwilson's comments now.

File: doc/oper.xml [hunk 13]
```diff
@@ -1503,3 +1503,57 @@ true
...
+      <Ref Oper="PartialOrderDigraphMeetOfVertices"
+        Label="for a digraph and two vertices"/>. <P/>
+
+      The operation <C>AsMonoid</C> behaves similarly to <C>AsSemigroup</C>
+      except that <A>filt</A> may also be equal to
+      <Ref Filt="IsPartialPermMonoid" BookName="ref"/>,
+      <A>digraph</A> must satisfy <Ref Prop="IsLatticeDigraph"/>, and the
+      output satisfies <Ref Filt="IsMonoid" BookName="ref"/>. <P/>
+
+      The output of both of these operations is guarenteed to be of minimal
```
wilfwilson: guarenteed -> guaranteed. Sorry Chris!

doc/oper.xml
```diff
@@ -1503,3 +1503,57 @@ true
...
+      degree (see <Ref Attr="DegreeOfPartialPermSemigroup" BookName="ref"/>).
+      Furthermore the <Ref Attr="GeneratorsOfSemigroup" BookName="ref"/> of
+      the output is guarenteed to be the unique generating set of minimal size.
```
wilfwilson: And again

ChristopherRussell:
I have changed the spell mistakes. Sorry this required so much reviewing!

wilfwilson:
Great :) Hope my repeated requests didn't annoy you or put you off contributing again. Thanks for your efforts

PR #139, opened 2018-10-16
Various tweaks to the kernel extension

fingolfin:
Please let me know if I should rebase this on some other branch instead of master (say, stable-0.13 ?)

### DIFF
--- a/src/digraphs.c
+++ b/src/digraphs.c
@@ -171,10 +171,8 @@ static Obj FuncGABOW_SCC(Obj self, Obj digraph) {
             comp = NEW_PLIST_IMM(T_PLIST_CYC, nr);
             SET_LEN_PLIST(comp, nr);
 
-            memcpy(
-                (void*) ((char*) (ADDR_OBJ(comp)) + sizeof(Obj)),
-                (void*) ((char*) (ADDR_OBJ(stack1)) + (end1 + 1) * sizeof(Obj)),
-                (size_t)(nr * sizeof(Obj)));
+            memcpy(ADDR_OBJ(comp) + 1, CONST_ADDR_OBJ(stack1) + (end1 + 1),
+                nr * sizeof(Obj));
 
             nr = LEN_PLIST(comps) + 1;
             SET_ELM_PLIST(comps, nr, comp);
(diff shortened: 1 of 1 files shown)


src/digraphs.c
```diff
@@ -25,6 +25,14 @@
 static Obj FuncDIGRAPH_OUT_NBS(Obj self, Obj digraph, Obj source, Obj range);
 static Obj FuncDIGRAPH_SOURCE_RANGE(Obj self, Obj digraph);
 
+#if !defined(GAP_KERNEL_MAJOR_VERSION) || GAP_KERNEL_MAJOR_VERSION < 3
+// compatibility with GAP <= 4.9
+static inline Obj NEW_PLIST_IMM(UInt type, Int plen)
+{
```
james-d-mitchell: Brace at the end of the previous line please @fingolfin

fingolfin: Oops, indeed! Will fix and rebase on stable-0.13

wilfwilson:
Part of the problem with Travis was that `TestDirectory` now returns lines like
```
              0 failures in 4 files
```
which our systems parsed as a failure because the string fail appeared (change introduced here: https://github.com/gap-system/gap/commit/1cd18b77b986b6d4f01cadd9f196dd57fd08876d). I've fixed this by not grepping for `fail' at the end.

For the failing coverage job, this occurred when running a test file (in this instance `tst/standard/attr.tst`) with the required version of Orb, namely 4.7.5. Running
```
echo "Test(\"tst/standard/attr.tst\");" | ../../bin/gap.sh -A
```
from the digraphs directory would give
```
# line 437 of 1698 (25%)dyld: lazy symbol binding failed: Symbol not found: _FuncPermList
  Referenced from: /Users/Wilf/gap/pkg/orb/bin/x86_64-apple-darwin17.7.0-default64-kv6/orb.so
  Expected in: flat namespace

dyld: Symbol not found: _FuncPermList
  Referenced from: /Users/Wilf/gap/pkg/orb/bin/x86_64-apple-darwin17.7.0-default64-kv6/orb.so
  Expected in: flat namespace

Abort trap: 6
```
This doesn't happen with Orb 4.7.6 so I increased the required version of Orb to 4.7.6.

wilfwilson:
@fingolfin If you please rebase one last time onto `stable-0.13` then everything should pass, and we'll merge.

wilfwilson:
Thanks @fingolfin

PR #144, opened 2018-11-15
Improve DigraphColouring

james-d-mitchell:
This PR implements the Welsh-Powell algorithm for digraph colouring. This implementation is almost always as good as the previous one (in terms of the number of colours required), and is faster (approx.
4 times faster on `DigraphRemoveLoops(RandomDigraph(10000, 0.1)))`.

### DIFF
--- a/doc/digraphs.bib
+++ b/doc/digraphs.bib
@@ -94,3 +94,17 @@ MRREVIEWER = {Joseph A. Thas},
        DOI = {10.1007/BF02579178},
        URL = {http://dx.doi.org/10.1007/BF02579178},
 }
+
+@article{Welsh1967aa,
+author = {Welsh, D. J. A. and Powell, M. B.},
+title = {An upper bound for the chromatic number of a graph and its application to timetabling problems},
+journal = {The Computer Journal},
+volume = {10},
+number = {1},
+pages = {85-86},
+year = {1967},
+doi = {10.1093/comjnl/10.1.85},
+URL = {http://dx.doi.org/10.1093/comjnl/10.1.85},
+eprint = {/oup/backfile/content_public/journal/comjnl/10/1/10.1093/comjnl/10.1.85/2/100085.pdf}
+}
+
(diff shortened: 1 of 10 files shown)


wilfwilson:
Nice. Do you think it's faster because you implemented it better, or because it's an inherently faster algorithm?

I'll be happy to merge it, but I do have one request: could you update the documentation for `DigraphColouring` to mention the name of the algorithm? Currently it says quite generically that it uses a greedy algorithm, but now it would be nice to be more specific now that we can be.

james-d-mitchell:
Good idea I’ll do this tomorrow

james-d-mitchell:
@wilfwilson on second reflection, I wonder if it might not be a good idea to have `DigraphGreedyColouring`, `DigraphWelshPowellColouring` and so on. Given that we have an implementation of the standard greedy algorithm, it seems a shame to throw it away, given that there are some cases when it gives a better colouring than the new version. What do you think?

james-d-mitchell:
@wilfwilson I'd propose to keep `DigraphColouring`, and use the version of the algorithm which gives the "best" colouring in the most cases. So, currently, that'd be the Welsh-Powell.

wilfwilson:
@james-d-mitchell Sounds good!

james-d-mitchell:
@wilfwilson I've finished this now. I decided to go in a slightly different direction. I changed to make `DigraphGreedyColouring` for a digraph and an ordering of its vertices, the main function. This way we can get the old behaviour by doing:

~~~~
DigraphGreedyColouring(D, [1 .. DigraphNrVertices(D)]);
~~~~

and the new behaviour by doing:

~~~~
DigraphGreedyColouring(D); # or
DigraphGreedyColouring(D,. DigraphWelshPowellOrder); # or
DigraphGreedyColouring(D,. DigraphWelshPowellOrder(D)); # or
~~~~

Let me know if spot any issues!

james-d-mitchell:
Nuts, thought I did everything

james-d-mitchell:
I just thought of something else to add to this, so please don't merge it yet.

james-d-mitchell:
On second thoughts, I have nothing further to add. Merge if happy @wilfwilson

PR #145, opened 2018-11-15
Add IsDigraphColouring

james-d-mitchell:
A method that verifies if a list or transformation is a colouring of a digraph. There are no tests or doc for this yet.

### DIFF
--- a/doc/grahom.xml
+++ b/doc/grahom.xml
@@ -541,3 +541,47 @@ false]]></Example>
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="IsDigraphColouring">
+  <ManSection>
+    <Oper Name="IsDigraphColouring" Arg="digraph, list"/>
+    <Oper Name="IsDigraphColouring" Arg="digraph, t" 
+      Label="for a transformation"/>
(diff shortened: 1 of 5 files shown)


gap/oper.gi
```diff
@@ -1891,3 +1891,31 @@ function(filt, digraph)
   fi;
   return AsSemigroup(IsPartialPermSemigroup, digraph);
 end);
+
+InstallMethod(IsDigraphColouring, "for a digraph and a list",
+[IsDigraph, IsHomogeneousList],
+function(digraph, colours)
+  local n, out, v, w;
+  n := DigraphNrVertices(digraph);
+  if Length(colours) <> n or (n > 0 and not IsPosInt(colours[1])) then
```
wilfwilson: I know you're not finished with, but just in case it's useful, here's a comment:
```
gap> IsHomogeneousList([1,-1]);
true
```
So just because the first point in `colours` is a positive integer, it doesn't mean that the rest of the integers in `colours` are positive.

james-d-mitchell: Good point :)

james-d-mitchell:
@wilfwilson I think this is now good to go.

PR #146, opened 2018-11-15
Allow an immutable list of nodes to be removed

james-d-mitchell:
Before this PR it was sometimes possible to remove an immutable list of nodes from a digraph, and other times it was not. This PR always allows an immutable list of nodes to be remove from a digraph.


james-d-mitchell:
@wilfwilson done! The test I've added is more or less what prompted me to make this PR in the first place.

PR #148, opened 2018-11-21
Add DigraphShortestPath method

MTWhyte:
This pull request implements a `DigraphShortestPath` method, which takes a digraphs and two positive integers, and returns a list in the same format as `DigraphPath`.

The shortest path is found using a breadth-first search.

### DIFF
--- a/doc/oper.xml
+++ b/doc/oper.xml
@@ -1203,6 +1203,38 @@ fail
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="DigraphShortestPath">
+<ManSection>
+  <Oper Name="DigraphShortestPath" Arg="digraph, u, v"/>
+  <Returns>A pair of lists, or <K>fail</K>.</Returns>
+  <Description>
+    Returns the shortest directed path in the digraph digraph from the vertex u to the vertex v, if such a path exists. If u = v, then the shortest non-trivial cycle is returned, again, if it exists. Otherwise, this operation returns
+    <K>fail</K>.  See Section <Ref Subsect="Definitions" Style="Text"/> for the
+    definition of a directed path and a directed cycle.
+    <P/>
+
+    See <Ref Oper="DigraphPath"/> for details on the output.
+
+    The method for <C>DigraphShortestPath</C> has worst case complexity of <M>O(m +
+      n)</M> where <M>m</M> is the number of edges and <M>n</M> the number of
+    vertices in <A>digraph</A>.
+
+<Example><![CDATA[
+gap> gr := Digraph([[1, 2], [3], [2, 4], [1], [2, 4]]);
+<digraph with 5 vertices, 8 edges>
+gap> DigraphShortestPath(gr, 5, 1);
+[ [ 5, 4, 1 ], [ 2, 1 ] ]
+gap> DigraphShortestPath(gr, 3, 3);
+[ [ 3, 2, 3 ], [ 1, 1 ] ]
+gap> DigraphShortestPath(gr, 5, 5);
+fail
+gap> DigraphShortestPath(gr, 1, 1);
+[ [ 1, 1 ], [ 1 ] ]
+]]></Example>
+  </Description>
+</ManSection>
+<#/GAPDoc>
+
 <#GAPDoc Label="IteratorOfPaths">
 <ManSection>
   <Oper Name="IteratorOfPaths" Arg="digraph, u, v"/>
--- a/gap/oper.gi
+++ b/gap/oper.gi
@@ -1167,6 +1167,75 @@ function(digraph, u, v)
   return DIGRAPH_PATH(OutNeighbours(digraph), u, v);
 end);
 
+# DigraphShortestPath
+
+InstallMethod(DigraphShortestPath, "for a digraph and two pos ints",
+[IsDigraph, IsPosInt, IsPosInt],
+function(digraph, u, v)
+  local current, next, parent, distance, falselist, verts, nbs, path, edge,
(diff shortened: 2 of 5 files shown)


gap/oper.gi
```diff
@@ -1239,6 +1239,67 @@ function(digraph, u, v)
...
+            od;
+            Add(path[1], u);  # Adds the starting vertex to the list of vertices.
+            return [Reversed(path[1]), Reversed(path[2])];
+          fi;
+
+        od;
+      od;
+
+      current := ListBlist(verts, next);
+      next    := IntersectionBlist(next, falselist);
```
james-d-mitchell: I think (not 100%) that you'd be better off doing `IntersectBlist(next, falselist);`. The reason being that `IntersectionBlist` produces a new line, whereas `IntersectBlist` changes `next` in-place.

gap/oper.gi
```diff
@@ -1239,6 +1239,67 @@ function(digraph, u, v)
...
+  next      := [];
+  falselist := [];
+  for i in verts do
+    Add(next, false);
+    Add(falselist, false);
+  od;
+
+  n := 0;  # Counts the loops
+  while current <> [] do
+    n    := n + 1;
```
james-d-mitchell: Weird spacing before the `:=` here.

gap/oper.gi
```diff
@@ -1239,6 +1239,67 @@ function(digraph, u, v)
...
+          fi;
+
+        od;
+      od;
+
+      current := ListBlist(verts, next);
+      next    := IntersectionBlist(next, falselist);
+
+    od;
+      return fail;
```
james-d-mitchell: Bad indentation.

doc/oper.xml
```diff
@@ -1203,6 +1203,41 @@ fail
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="DigraphShortestPath">
+<ManSection>
+  <Oper Name="DigraphShortestPath" Arg="digraph, u, v"/>
+  <Returns>A pair of lists, or <K>fail</K>.</Returns>
+  <Description>
+    If there exists a non-trivial directed path (or a non-trivial cycle, in the
+    case that <A>u</A> <C>=</C> <A>v</A>) from vertex <A>u</A> to vertex
```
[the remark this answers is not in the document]
wilfwilson: This should be either `shortest cycle of length greater than 0`, or `shortest non-trivial cycle`, as discussed below.

gap/oper.gi
```diff
@@ -1239,6 +1239,67 @@ function(digraph, u, v)
...
+[IsDigraph, IsPosInt, IsPosInt],
+function(digraph, u, v)
+  local current, next, parent, distance, falselist, verts, nbs, n, a, b, i, path;
+
+  verts    := DigraphVertices(digraph);
+  nbs      := OutNeighbors(digraph);
+  distance := [];
+
+  if u = v and v in nbs[v] then  # Considers the trivial path from v to v
+    return [[v, v], [Position(nbs[v], v)]];
```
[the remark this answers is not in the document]
wilfwilson: Definition of trivial in the package documentation:
```
The length of a directed walk (v1,e1,v2,e2,...,en−1,vn) is equal to n−1, the number of edges it contains. A directed walk of zero length, i.e. a sequence (v) for some vertex v, is called trivial. A trivial directed walk is considered to be both a circuit and a cycle, as is the empty directed walk ().
```
By the original documentation in this PR:
```
(or a non-trivial cycle, in the case that <A>u</A> <C>=</C> <A>v</A>)
```
a loop therefore qualifies as the shortest path, since a loop is a non-trivial cycle.

However, I therefore suggest adjusting the misleading comment in the line above.

gap/oper.gi
```diff
@@ -1239,6 +1239,67 @@ function(digraph, u, v)
...
+InstallMethod(DigraphShortestPath, "for a digraph and two pos ints",
+[IsDigraph, IsPosInt, IsPosInt],
+function(digraph, u, v)
+  local current, next, parent, distance, falselist, verts, nbs, n, a, b, i, path;
+
+  verts    := DigraphVertices(digraph);
+  nbs      := OutNeighbors(digraph);
+  distance := [];
+
+  if u = v and v in nbs[v] then  # Considers the trivial path from v to v
```
wilfwilson: A loop is not a trivial path. Therefore please change this comment, to something like `# is u = v, and there is a loop at v?`

gap/oper.gi
```diff
@@ -1239,6 +1239,67 @@ function(digraph, u, v)
   return DIGRAPH_PATH(OutNeighbours(digraph), u, v);
 end);
 
+# DigraphShortestPath
+
+InstallMethod(DigraphShortestPath, "for a digraph and two pos ints",
+[IsDigraph, IsPosInt, IsPosInt],
+function(digraph, u, v)
+  local current, next, parent, distance, falselist, verts, nbs, n, a, b, i, path;
```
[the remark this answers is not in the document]
MTWhyte: I've added these tests to the start of the function. I believe the

```
if IsDigraphEdge(digraph, [u, v]) then
      return [[u, v], [Position(OutNeighboursOfVertex(digraph, u), v)]];
```

does much the same thing as the existing

```
if u = v and v in nbs[v] then  # Considers the trivial path from v to v
    return [[v, v], [Position(nbs[v], v)]];
```
which was the subject of a previous comment. I shall just remove the latter for simplicity.

james-d-mitchell:
I'm happy for this to be merged, but it needs to be rebased on the current `master` first.

PR #150, opened 2018-11-22
Modernize ac_find_gap.m4

fingolfin:
* drop CONFIGNAME support
* make it possible to compile against future GAP releases with
  compatibility mode turned off

### DIFF
--- a/m4/ac_find_gap.m4
+++ b/m4/ac_find_gap.m4
@@ -1,6 +1,6 @@
 # Find the location of GAP
 # Sets GAPROOT, GAPARCH and GAP_CPPFLAGS appropriately
-# Can be configured using --with-gaproot=... and CONFIGNAME=...
+# Can be configured using --with-gaproot=...
 #######################################################################
 
 AC_DEFUN([AC_FIND_GAP],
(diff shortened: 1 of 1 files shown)


fingolfin:
BTW I based this on stable-0.13, so if you'd rather merge this there, it's trivial to retarget this PR.

wilfwilson:
@james-d-mitchell I merged this without talking to you, hope that's alright, but I noticed this in the README which is now out of date somehow:

> If you installed GAP on several architectures, you must execute the configure/make step for each of the architectures. You can either do this immediately after configuring and compiling GAP itself on this architecture, or alternatively (when using version 4.5 of GAP or newer) set the environment variable CONFIGNAME to the name of the configuration you used when compiling GAP before running ./configure. Note however that your compiler choice and flags (environment variables CC and CFLAGS) need to be chosen to match the setup of the original GAP compilation. For example you have to specify 32-bit or 64-bit mode correctly!

james-d-mitchell:
No problem, I'm happy for you to merge things without talking to me! 

What's the correct advice for the README now?

wilfwilson:
I don't know what needs to be done to make it correct, that's why mostly I've brought it to your attention. But certainly the stuff about `CONFIGNAME` should be removed. I've no idea about any of this compilation stuff really.

PR #153, opened 2018-11-23
kernel: fix compile warnings

james-d-mitchell:


wilfwilson:
I've not been seeing any warnings myself, so I can't really offer any input on this one.

james-d-mitchell:
@wilfwilson I wasn't seeing the warnings either, until I turned them on ;) This came about since I had an old TODO to fix some compiler warning (which it turns out was already fixed), and then these came up, and I fixed them too. 

Since it passes, can we merge?

PR #156, opened 2018-12-11
Add IsPlanarDigraph and related

james-d-mitchell:
This PR adds functionality to Digraphs to check whether a digraph is planar, and some related functionality. This is essentially just a wrapper around the C library [edge-addition-planarity-suite](https://github.com/graph-algorithms/edge-addition-planarity-suite), plus a little bit more. 

It is currently missing:

- [x] some tests
- [x] all doc
- [x] valgrind

Things to keep in mind:

1. This makes Digraphs dependent on an external C-library (which appears to work well as an autohell subpackage). A copy of the library is included in this PR. I'm not completely sure this is desirable.

2. It might be the case that at some point in the future [edge-addition-planarity-suite](https://github.com/graph-algorithms/edge-addition-planarity-suite) stops working on some platforms or is no longer maintained. At that point, we can introduce a configure flag `--disable-edge-addition-planarity-suite` to disable these features. I don't really want to add the flag now, because then we have to test with and without the flag and everything gets more complex.


wilfwilson:
Thanks, I'll review once there are tests and documentation and the CI passes.

james-d-mitchell:
Interestingly, this almost works ;)

james-d-mitchell:
Also I will split the PR into 2 commits, one adding all the files in the subpackage and the other with the changes to Digraphs itself

james-d-mitchell:
Also @wilfwilson any chance you can merge stable-0.14 into master? So I can rebase on that?

wilfwilson:
Merged!

james-d-mitchell:
Super thanks!

james-d-mitchell:
I think this is more or less good to go. I'm just going to valgrind, and if the CI passes, then this is ready for review.

james-d-mitchell:
Oh wait I forgot, I don't know how to valgrind GAP, I committed some files that are generated by mistake will remove them now.

james-d-mitchell:
Ready to go

doc/planar.xml
```diff
@@ -0,0 +1,314 @@
...
+<#/GAPDoc>
+
+<#GAPDoc Label="IsOuterPlanarDigraph">
+<ManSection>
+  <Prop Name="IsOuterPlanarDigraph" Arg="digraph"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    An <E>outer planar</E> digraph is a digraph that can be embedded in the
+    plane in such a way that its edges do not intersect, and all vertices
+    belong to the unbounded face of the embedding.  A digraph is planar if and
```
wilfwilson: A digraph is planar or _outer_ planar?

doc/planar.xml
```diff
@@ -0,0 +1,314 @@
...
+<#/GAPDoc>
+
+<#GAPDoc Label="KuratowskiPlanarSubdigraph">
+<ManSection>
+  <Attr Name="KuratowskiPlanarSubdigraph" Arg="digraph"/>
+  <Returns>A list or <K>fail</K>.</Returns>
+  <Description>
+    <C>KuratowskiPlanarSubdigraph</C> returns the list of out-neighbours
+    of a (not necessarily induced) subdigraph of the digraph <A>digraph</A>
+    that witness the fact that <A>digraph</A> is not planar, or
```
wilfwilson: witness -> witnesses

doc/planar.xml
```diff
@@ -0,0 +1,314 @@
...
+    <K>fail</K> if <A>digraph</A> is planar.  In other words,
+    <C>KuratowskiPlanarSubdigraph</C> returns the out-neighbours of a
+    subdigraph of <A>digraph</A> that is homeomorphic to the complete graph
+    with four vertices, or to the complete bipartite graph with vertex sets of
```
wilfwilson: Don't you mean K_5 and K_3,3 and not K_4 and K_2,3?

wilfwilson: Also note that here you talk about `the complete graph with four vertices` whereas above you talk about the `complete graph on <C>4</C> vertices` - it's not vital but I'd recommend to use the same wording.

doc/planar.xml
```diff
@@ -0,0 +1,314 @@
...
+
+<#GAPDoc Label="PlanarEmbedding">
+<ManSection>
+  <Attr Name="PlanarEmbedding" Arg="digraph"/>
+  <Returns>A list or <K>fail</K>.</Returns>
+  <Description>
+    If <A>digraph</A> is a planar digraph, then 
+    <C>PlanarEmbedding</C> returns the list of out-neighbours
+    of a subdigraph of <A>digraph</A> such that each vertex's
+    neighbors are given in clockwise order. If <A>digraph</A> is not planar,
```
wilfwilson: Use neighbours for consistency with the rest of the documentation/sentence please.

doc/planar.xml
```diff
@@ -0,0 +1,314 @@
...
+  </Description>
+</ManSection>
+<#/GAPDoc>
+
+<#GAPDoc Label="PlanarEmbedding">
+<ManSection>
+  <Attr Name="PlanarEmbedding" Arg="digraph"/>
+  <Returns>A list or <K>fail</K>.</Returns>
+  <Description>
+    If <A>digraph</A> is a planar digraph, then
```
wilfwilson: I'm not sure this documentation properly conveys what's happening here. The point is that if the vertices are drawn as suggested by the returned list, then you can fill in the edges without edges crossing. I feel like the connection between this clockwise property and planarity could be elucidated a bit more.

doc/planar.xml
```diff
@@ -0,0 +1,314 @@
...
+
+<#GAPDoc Label="OuterPlanarEmbedding">
+<ManSection>
+  <Attr Name="OuterPlanarEmbedding" Arg="digraph"/>
+  <Returns>A list or <K>fail</K>.</Returns>
+  <Description>
+    If <A>digraph</A> is an outer planar digraph, then 
+    <C>OuterPlanarEmbedding</C> returns the list of out-neighbours
+    of a subdigraph of <A>digraph</A> such that each vertex's
+    neighbors are given in clockwise order. If <A>digraph</A> is not outer planar,
```
wilfwilson: Analogous comment here. As things stand, the documentation for this is essentially identical to `PlanarEmbedding`, but they're capturing different properties of a graph.

doc/planar.xml
```diff
@@ -0,0 +1,314 @@
...
+
+<#GAPDoc Label="SubdigraphHomeomorphicToK">
+<ManSection>
+  <Attr Name="SubdigraphHomeomorphicToK23" Arg="digraph"/>
+  <Attr Name="SubdigraphHomeomorphicToK33" Arg="digraph"/>
+  <Attr Name="SubdigraphHomeomorphicToK4" Arg="digraph"/>
+  <Returns>A list or <K>fail</K>.</Returns>
+  <Description>
+    These attributes return the list of out-neighbours of a subdigraph of
+    the digraph <A>digraph</A> which is homeomorphic to one of the following:
```
wilfwilson: Do we want to define homeomorphic somewhere in the doc, perhaps in the intro where we define paths etc?

doc/planar.xml
```diff
@@ -0,0 +1,314 @@
...
+    the complete bipartite graph with vertex sets of sizes <C>2</C> and
+    <C>3</C>; the complete bipartite graph with vertex sets of sizes <C>3</C>
+    and <C>3</C>; or the complete graph with <C>4</C> vertices. If
+    <A>digraph</A> has no such subdigraphs, then <K>fail</K> is returned.
```
wilfwilson: Maybe emphasise the point that a digraph is planar/outer planar if and only if it has no such sub digraph?

Why does `SubdigraphHomeomorphicToK5` not exist?

gap/attr.gi
```diff
@@ -384,14 +384,10 @@ InstallMethod(OutNeighbours, "for a digraph",
 [IsDigraph],
 function(digraph)
   local out;
-  if IsBound(digraph!.adj) then
-    return digraph!.adj;
```
wilfwilson: Why is this change included here? And what is the explanation for its deletion?

gap/attr.gi
```diff
@@ -1487,3 +1483,50 @@ function(gr)
...
+
+InstallMethod(MaximalAntiSymmetricSubdigraph, "for a digraph",
+[IsDigraph],
+function(D)
+  local n, m, out, i, j;
+
+  n := DigraphNrVertices(D);
+  if n <= 1
+      or (HasIsAntisymmetricDigraph(D) and IsAntisymmetricDigraph(D)) then
+    return D;
```
wilfwilson: If `n <= 1` or the digraph is already antisymmetric, it's possible that there are still multiple edges, which the documentation forbids!

gap/oper.gi
```diff
@@ -1005,13 +1005,13 @@ function(digraph, edge)
 end);
 
 InstallMethod(IsDigraphEdge, "for a digraph, pos int, pos int",
```
wilfwilson: Change the string please to agree with the args

gap/oper.gi
```diff
@@ -1005,13 +1005,13 @@ function(digraph, edge)
...
-[IsDigraph, IsPosInt, IsPosInt],
+[IsDigraph, IsInt, IsInt],
 function(digraph, u, v)
   local n;
 
   n := DigraphNrVertices(digraph);
 
-  if u > n or v > n then
+  if u > n or v > n or u < 0 or v < 0 then
```
wilfwilson: Why do we want to allow `0`? Vertices are currently restricted to `[1..n]`

wilfwilson:
I am also a bit wary of including a whole library. But if it seems to work, then let's do it. If it causes problems for us or other users in the future, then we can deal with this then.

james-d-mitchell:
Thanks @wilfwilson I’ll try to update the pr today

doc/planar.xml
james-d-mitchell: Perhaps this could be clearer, I have not been able to find any description of the return value of this method (from the edge-addition-planarity-suite), so I basically copied what is written here from the corresponding manual entry from Sage see [here](https://doc.sagemath.org/html/en/reference/graphs/sage/graphs/generic_graph.html#sage.graphs.generic_graph.GenericGraph.get_embedding).

doc/planar.xml
james-d-mitchell: I've added a link to the doc for `IsPlanarDigraph` and `IsOuterPlanarDigraph`, rather than repeating the doc here again. 

> Why does SubdigraphHomeomorphicToK5 not exist?

Because it isn't implemented in the edge-addition-planarity-suite, and it isn't clear to me how to do this (at least in any good way). I did have a version of this PR where I had a function called `SubdigraphHomeomorphicTo(digraph, subdigraph)` but I abandoned it as too complicated in the first instance (it would only have worked for digraphs homeomorphic to K_4, K_{2, 3}, and K_{3, 3}, anyway.

gap/attr.gi
james-d-mitchell: Didn't we remove the component `adj` from all digraphs at some point in the not too distant past? This was originally used to hold the `OutNeighbours` for access in the C code, but if I recall this was unnecessary, and I think you removed it. I believe this code is a remnant of the older code, and I noticed it when writing tests for `MaximalAntiSymmetricSubdigraph` (it was not covered by any tests, and removing it seems to have no effect). Perhaps it should be in another PR, but it is in a separate commit at least.

gap/attr.gi
james-d-mitchell: Ooops, I swapped the clauses around here.

gap/oper.gi
james-d-mitchell: Sorry,

gap/oper.gi
james-d-mitchell: Double sorry my bad. BTW the reason for this change was that I want to call `IsDigraphEdge` in the C code, and there, sometimes GAP integers don't seem belong to `IsPosInt` even though they are positive and would satisfy `IsPosInt` in GAP.

james-d-mitchell:
I agree with you @wilfwilson, on the other hand, we already rely on Bliss, so maybe it's not that bad either. I'll update the PR just now.

james-d-mitchell:
Updated! Off to Paris now, will check in again tomorrow. Thanks for the review @wilfwilson !

gap/attr.gi
wilfwilson: Ah right, no problem. I'm not sure if I did the changes, but yes this the `adj` component is removed, and certainly if these lines are not being covered by tests then it's not needed. I'm not normally against a PR changing unrelated things, but my default assumption when reviewing a PR is that the changes are related, and so I was confused at this point because I couldn't work out what the relation was here.

doc/planar.xml
wilfwilson: Fair enough, I'm surprised that it doesn't contain a K_5 thing, but that it does contain the other three, but then again I have no idea what the suite is doing!

wilfwilson:
I hope you remembered your yellow vest @james-d-mitchell! I'll merge this tomorrow, to give you the chance to make any additional tweaks.

wilfwilson:
@james-d-mitchell I just thought about licensing issues. Is the license of digraphs compatible with the suite, and vice versa? Do we need to include a copy of their license or something?

wilfwilson:
From https://github.com/graph-algorithms/edge-addition-planarity-suite/blob/master/LICENSE.TXT:
```
Redistribution and use in source and binary forms, with or without
modification, are permitted provided that the following conditions are met:

* Redistributions of source code must retain the above copyright notice, this
  list of conditions and the following disclaimer.

* Redistributions in binary form must reproduce the above copyright notice,
  this list of conditions and the following disclaimer in the documentation
  and/or other materials provided with the distribution.

* Neither the name of The Edge Addition Planarity Suite nor the names of its
  contributors may be used to endorse or promote products derived from
  this software without specific prior written permission.
```
Fair enough. At the top of all the new files you've retained
```
/*
Copyright (c) 1997-2015, John M. Boyer
All rights reserved.
See the LICENSE.TXT file for licensing information.
*/
```
But I think it would be a good idea to actually include that `LICENSE.TXT` file as, for instance,
`edge-addition-planarity-suite-Version_3.0.0.5/LICENSE.TXT`.

james-d-mitchell:
hi @wilfwilson finally back at my keyboard. i thought I did include a copy of the `LICENSE` file, but you're correct, it looks like I didn't. I have just pushed a new version including the `LICENSE` file.

PR #161, opened 2019-01-16
Add AsSemigroup method for strong semilattices of groups

flsmith:
This PR adds a method for constructing Clifford semigroups as partial permutation semigroups, given a semilattice, groups, and homomorphisms.

You can choose between an easily-computed representation (the default) and a more difficult but possibly smaller degree representation. The smaller degree representation is based on 

"The minimal faithful degree of a semilattice of groups"
David Easdown
J. Austral. Math. Soc. (Series A) 45 (1988), 341-350

### DIFF
--- a/doc/oper.xml
+++ b/doc/oper.xml
@@ -1566,3 +1566,45 @@ gap> Size(M);
     </Description>
   </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="AsCliffordSemigroup">
+  <ManSection>
+    <Oper Name="AsSemigroup" Arg="filt, Y, gps, homs"/>
+    <Returns> A Clifford semigroup of partial perms. </Returns>
+    <Description>
+      The operation <C>AsSemigroup</C> requires that <A>filt</A> be equal to
+      <Ref Filt="IsPartialPermSemigroup" BookName="ref"/>.
+      If <A>Y</A> is a <Ref Prop="IsJoinSemilatticeDigraph"/> or
+      <Ref Prop="IsMeetSemilatticeDigraph"/>,
+      <A>gps</A> is a list of groups corresponding to each vertex, and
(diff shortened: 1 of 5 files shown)


james-d-mitchell:
@flsmith and I just discussed this in person, and @flsmith is going to check if there is overlap between the code in this PR and `SmallerDegreePartialPermRepresentation`.

wilfwilson:
What could be the possible outcome of this checking? That this PR somehow gets revised accordingly?

james-d-mitchell:
Yes, that's right, it isn't clear if there is overlap with what's already in Semigroups, with respect to `SmallerDegreePartialPermRepresentation` so this should be figured out, and if there's an overlap it should probably be removed.

flsmith:
I've now removed the overlap, so the fast method which does not attempt to minimise the degree is used. You can then use `SmallerDegreePartialPermRepresentation` on the result.

doc/oper.xml
```diff
@@ -1589,3 +1589,45 @@ gap> Size(M);
     </Description>
   </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="AsCliffordSemigroup">
```
ChristopherRussell: Should you change this to <#GAPDoc Label="AsSemigroup"> ?

doc/z-chap4.xml
```diff
@@ -70,6 +70,7 @@
 
   <Section><Heading>Associated semigroups</Heading>
     <#Include Label="AsSemigroup">
+    <#Include Label="AsCliffordSemigroup">
```
ChristopherRussell: Should this also be "AsSemigroup"?

flsmith:
Resolved comments, ready for merge

james-d-mitchell:
Can you please rebase this onto the current `master` branch? @flsmith, thanks!

Issue #158, opened 2019-01-11
Digraph6 file format incompatibility


jukkakohonen:
Writing a Digraph6 (.d6) file from nauty, and then trying to read it with IteratorFromDigraphFile, fails with the following error.

```
gap> iter := IteratorFromDigraphFile("test.d6");
Error, Digraphs: DigraphFromDigraph6String: usage,
the input string <s> is not in valid digraph6 format, at /home/local/kohonen/gap-4.9.2/pkg/digraphs-0.12.1/gap/io.gi:827 called from
decoder( line ) at /home/local/kohonen/gap-4.9.2/pkg/digraphs-0.12.1/gap/io.gi:253 called from
file!.coder( file ) at /home/local/kohonen/gap-4.9.2/pkg/digraphs-0.12.1/gap/io.gi:198 called from
<function "IteratorFromDigraphFile">( <arguments> )
 called from read-eval loop at *stdin*:5
type 'quit;' to quit to outer loop
```

To reproduce, let the file "test.d6" contain (for example) this single line:
`&O?????C??O?@??C??O?@??C??O?@??C??O?@??C??o??`

The problem seems to be this. Nauty (for example, function writed6) writes digraphs so that each line begins with an ampersand (&). Indeed that is how nauty documentation [http://pallini.di.uniroma1.it/nug27.pdf] (page 81) says.  However, IteratorFromDigraphFile requires the lines to begin with a plus sign (+).

If I change the & into a + in the file, then the digraph is read in just fine. How did this incompatibility come to be? Did the digraph6 format change at some time?

Versions:
GAP 4.9.2
Digraphs 0.12.1

james-d-mitchell:
Thanks for the report @jukkakohonen, we will investigate and let you know.

mtorpey:
I think I can explain this.  Unfortunately, the "digraph6" format used in this package is different from the "digraph6" format used in Nauty.  When Julius and I wrote this code (probably late 2014) we weren't aware of an established format for digraphs, so we invented a new one based on graph6 and just called it digraph6.  I don't know whether Nauty's "digraph6" format existed back then, or whether it was documented, but we certainly didn't know about it.

[...]

PR #162, opened 2019-01-23
Support Nauty digraph6 format

mtorpey:
As discovered in Issue #158, our version of digraph6 differs from that in Nauty, but the two formats cannot be confused with each other (since they start with `+` and `&` respectively).  This PR adds support for reading the Nauty format, while preserving the ability to read the old one (with an info warning).  This PR makes us write the Nauty format, and removes the ability to write the old one (we should use Nauty's format from now on).


gap/io.gi
```diff
@@ -821,8 +826,14 @@ function(s)
...
 
-  # Check for the special '+' character
-  if s[1] <> '+' then
+  # Check for the special '&' character (or the deprecated '+')
+  if s[1] = '&' then
+    legacy := false;
+  elif s[1] = '+' then
+    legacy := true;
+    Info(InfoDigraphs, 1, "Digraph6 strings beginning with '+' are deprecated");
+    Info(InfoDigraphs, 1, "and may not be recognised by other programs.");
```
wilfwilson: I think this warning needs to be more forceful/explicit. You could read this and think that the only thing that changed was that the '+' has been changed to a '&'. But the actual encoding is fundamentally different.

Maybe something like

> Digraph6 strings that begin with '+' use an old specification of the Digraph6 format that is incompatible with the current, and now standard, specification. These old strings can still be read by the Digraphs package, but it is unlikely that they can be read correctly by other programs. We recommend that you re-encode your digraphs with the new format.

mtorpey:
Altered as requested.  See what you think of the new info warning – succinct enough? ;)

mtorpey:
Altered as requested. See what you think of the new info warning – succinct enough? :wink:

james-d-mitchell:
Great, super thanks @mtorpey !!

PR #164, opened 2019-02-06
Characteristic Polynomial of Digraph

le27:


gap/attr.gi
```diff
@@ -1530,3 +1530,12 @@ function(D)
   SetIsAntisymmetricDigraph(out, true);
   return out;
 end);
+
+#
```
wilfwilson: Please could you remove this line and the previous one?

gap/attr.gi
```diff
@@ -1530,3 +1530,12 @@ function(D)
...
+
+InstallMethod(CharacteristicPolynomial,
+"for a digraph",
+[IsDigraph],
+function(gr)
+    return CharacteristicPolynomial(AdjacencyMatrix(gr));
```
wilfwilson: Please could you add a test that provides code coverage for this line?

PR #165, opened 2019-02-13
Implemented checking for vertex/edge transitivity

### LINKED ISSUE #157: Improve methods for deciding vertex and arc transitivity
A digraph is vertex transitive if its automorphism group acts transitively on its vertices, and it is arc transitive if its automorphism group acts transitively on its arcs.

This can be implemented trivially, but that involves computing the automorphism group of the graph.

There are further properties of graphs determined by the automorphism group which might be nice to have, and maybe some of the above properties can be (dis)proven using an algorithm that is more efficient than an automorphism group computation. (In which case I'd like to know said algorithm).

GrahamCampbell:
Closes #157.


wilfwilson:
Thanks @GrahamCampbell. I'm not sure if this closes #157, exactly, since there @markuspf is hoping for methods to decide vertex and edge transitivity that does not require the computation of the automorphism group, at least not in all cases. For example, the least we could do to improve the vertex transitivity test is to first test whether the graph is regular: a vertex transitive digraph is necessarily in- and out-regular. But the changes in this PR are good to have, and I'll happily accept them – they don't stop us from doing further work. I'll squash merge this once the CI passes (I made some small tweaks to the PR, which requires the CI to run again).

PR #167, opened 2019-02-14
PackageInfo.g: add License field

fingolfin:
Not sure which branch to target, please let me know if `master` is wrong,.


wilfwilson:
Thanks @fingolfin I cherry-picked it into the `stable-0.14` branch and tweaked the comment so that it didn't fail our linter.

PR #169, opened 2019-02-19
add DotHighlightedDigraph method

flsmith:
This PR adds a method for producing a dot string where a subdigraph is highlighted.

I'm open to comments if people would like this to work differently - I wrote it this way for a specific application, and it might not be the best generic solution.

### DIFF
--- a/doc/display.xml
+++ b/doc/display.xml
@@ -244,3 +244,31 @@ gap> FileString("dot/preset.dot", DotProrderDigraph(gr));
     </Description>
   </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="DotHighlightedDigraph">
+  <ManSection>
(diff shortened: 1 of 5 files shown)


james-d-mitchell:
This PR is very welcome. I wonder if it might not be a more consistent approach to simply allow a colouring of the vertices of a digraph be the second argument to `DotDigraph`? This would be quite natural, and also serve your purpose, example usage:

`DotDigraph(D, [1, 2, 3], ["blue", "red", "pink"]);`

Since there is built in functonality for checking if a coloring is valid, this might be one way to go. What do you think @wilfwilson and @flsmith ?

flsmith:
I've discussed this with @james-d-mitchell and he's happy with this as it is. 

I'll create a new issue for giving arbitrary colours to vertices/edges, which might be useful too.

PR #171, opened 2019-02-21
Implement mutable digraphs

### LINKED ISSUE #90: Mutable digraphs
I would be useful to have `MutableDigraph` which could have `IsPositionalObjectRep` and not store attributes, because it is common to create a digraph and immediately modify it (remove multiple edges, loops, etc), and this involves unnecessary copying at present since `Digraph`s are always immutable. As with Issue #88, this would have to be done very carefully to modify the object in-place rather than copy for every operation modifying a digraph.

### LINKED ISSUE #98: Refactor `DIGRAPHS_Bipartite` for better performance.
We can find out whether a digraph is bipartite, and simultaneously obtain a 2-colouring for a bipartite digraph, by using a depth first search. Doing this would be much faster than the current implementation.

james-d-mitchell:
This is a work in progress to introduce mutable digraphs with a minimum of fuss. Although this is a work in progress, comments are very welcome. 

Also if anyone else wants to chip in, that'd be greatly appreciated. Basically every method that currently works for `IsDigraph` should become a method for mutable digraphs, then the method for immutable digraphs (all digraphs in the package up to now) will just call the method on a mutable copy, and then call `MakeImmutableDigraph`, and return. 

Resolves #90.
Resolves #98.


wilfwilson:
I've not looked at every single line of diff (and I don't think I will), but I think I've looked at a useful subset of the changes so far. And what I've seen makes sense and seems to be correct.

james-d-mitchell:
If the CI passes, then I propose to squash and merge this branch into `master`. The only remaining thing to do is to update the documentation, otherwise this is ready for release, at least in my opinion.

wilfwilson:
Fine by me. Travis seems to be playing up at the moment and going very slowly, so we might have a bit of a wait. But never mind.

PR #173, opened 2019-02-22
COUNT_TRUES_BLOCK 

### LINKED ISSUE #172: Digraphs doesn't work in GAP master branch
This is because GAP's `COUNT_TRUES_BLOCK` has been moved, so we are not including it any more. (See https://github.com/gap-system/gap/issues/3312). However, it has been moved to a header called `bits_intern.h`, which implies that the GAP developers intend for this to be an internal function. So perhaps the correct solution is not to simply include `bits_intern.h`. We'll see!

james-d-mitchell:
Resolves #172

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -115,6 +115,7 @@ AC_DEFUN([CHECK_COMPILER_BUILTIN],
         )])
 
 CHECK_COMPILER_BUILTIN([__builtin_ctzll],[0]);
+CHECK_COMPILER_BUILTIN([__builtin_popcountl],[0]);
 
 dnl ##
 dnl ## Output everything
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
Yes, I'd be interested to hear the response to your issue on gap too. I don't think a release is urgent, but can we maybe merge this PR so that our CI is fixed?

wilfwilson:
Happy to merge, and change to a better/different solution later if that is more appropriate. This will make the Travis tests for Digraphs pass. But unfortunately one of the Travis jobs for Semigroups tests against the master branch of GAP along with the newest releases of each package. Therefore that job will continue to fail, while ever 0.15.0 is the newest release or GAP master branch maintains its change.

wilfwilson:
James please can you now merge stable into master, and also consider temporarily removing that Travis job from Semigroups if you want those tests to pass

PR #175, opened 2019-02-27
Add RandomLattice method.

reiniscirpons:
This PR adds a method RandomLattice for constructing random lattices, given a positive integer n as input.

It then incrementally calculates a union lattice of sets, terminating once the lattice has at least n vertices. It is guaranteed that the resulting lattice will have at least n and no more than 2*n vertices. The average time complexity seems to be O(n^2).

Some results on the distribution of such lattices can be derived from

"The representation of posets and lattices by sets"
George Markowsky  
Algebra Universalis, 11 (1980), 173-192


tst/standard/digraph.tst
```diff
@@ -647,6 +647,14 @@ gap> RandomTournament(-1);
 Error, Digraphs: RandomTournament: usage,
 the argument <n> must be a non-negative integer,
 
+#  RandomLattice
+gap> RandomLattice(25);;
+gap> RandomLattice(1);
+<digraph with 1 vertex, 1 edge>
+gap> RandomLattice(-1);
+Error, no method found! For debugging hints type ?Recovery from NoMethodFound
+Error, no 1st choice method found for `RandomLattice' on 1 arguments
```
james-d-mitchell: Can you please add some tests that indicate approximately how good the distribution is obtained by `RandomLattice`. For example, something like:

~~~~
D := List([1 .. 100], x -> RandomLattice(7));
out := [];
for DD in D do 
  if ForAll(D, x -> not IsIsomorphicDigraph(x, DD)) then 
    Add(out, DD);
  fi;
od;
~~~~
Where you could replace the numbers 100 and 7 by something more sensible.

reiniscirpons:
I have implemented the requested changes. Could you review if everything is ok and if so, please merge? Thanks :)

doc/digraph.xml
```diff
@@ -773,6 +773,18 @@ gap> RandomTournament(10);
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="RandomLattice">
+<ManSection>
+  <Oper Name="RandomLattice" Arg="n"/>
+  <Returns>A digraph.</Returns>
+  <Description>
+    If <A>n</A> is a positive integer, this function return a random lattice
+    with <A>m</A> vertices, where it is guaranteed that <A>m</A> is between
```
james-d-mitchell: `<A>m</A>` -> `<C>m</C>`

james-d-mitchell: times 2

gap/digraph.gi
```diff
@@ -1104,3 +1104,36 @@ InstallMethod(RandomTournament, "for a func and an integer",
...
+  local fam, rand_blist;
+
+  fam := [BlistList([1 .. n], [])];
+
+  while Length(fam) < n do
+    rand_blist := List([1 .. n], x -> Random([true, false]));
+    UniteSet(fam, List(fam, x -> UnionBlist(x, rand_blist)));
+  od;
+
+  return Digraph(IsMutableDigraph, fam, IsSubsetBlist);
```
james-d-mitchell: Nice.

gap/digraph.gi
```diff
@@ -1104,3 +1104,36 @@ InstallMethod(RandomTournament, "for a func and an integer",
...
+InstallMethod(RandomLatticeCons, "for IsMutableDigraph and a pos int",
+[IsMutableDigraph, IsPosInt],
+function(filt, n)
+  local fam, rand_blist;
+
+  fam := [BlistList([1 .. n], [])];
+
+  while Length(fam) < n do
+    rand_blist := List([1 .. n], x -> Random([true, false]));
+    UniteSet(fam, List(fam, x -> UnionBlist(x, rand_blist)));
```
james-d-mitchell: I have no idea if this would make any difference but you could use `Set` instead of `List` here.

gap/digraph.gi
```diff
@@ -1104,3 +1104,36 @@ InstallMethod(RandomTournament, "for a func and an integer",
...
+  od;
+
+  return Digraph(IsMutableDigraph, fam, IsSubsetBlist);
+end);
+
+InstallMethod(RandomLatticeCons, "for IsMutableDigraph and a pos int",
```
james-d-mitchell: `IsMutableDigraph` -> `IsImmutableDigraph`

gap/digraph.gi
reiniscirpons: `UniteSet(set, list)` is defined for a set and a list, so I do not think that having a `Set` instead of a `List` as the second argument changes much. In practice is a slightly slower with `Set` (presumably since it needs to reorder the list and then `UniteSet` does not speed up).

james-d-mitchell: Ok, that all makes sense.

james-d-mitchell:
Thanks @reiniscirpons, merged!

PR #178, opened 2019-03-12
Fix Digraphs for cygwin

### LINKED ISSUE #177: Digraphs on windows compiles to .dll and not to .so
I still can not provide a working version of digraphs (and hence semigroups, which needs it) for GAP 4.10.1. It compiles to `bin/i686-pc-cygwin-default32-kv3/digraphs.dll` while its `AvailabilityTest` checks for `digraphs.so`.

Furthermore, this happens if I edit `AvailabilityTest` to check for `digraphs.dll`:

```
gap> LoadPackage("dig");
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Loading  orb 4.8.2 (Methods to enumerate orbits)
by Juergen Mueller (http://www.math.rwth-aachen.de/~Juergen.Mueller),
   Max Neunhöffer (http://www-groups.mcs.st-and.ac.uk/~neunhoef), and
   Felix Noeske (http://www.math.rwth-aachen.de/~Felix.Noeske).
Homepage: https://gap-packages.github.io/orb
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Loading  GRAPE 4.8.1 (GRaph Algorithms using PErmutation groups)
by Leonard H. Soicher (http://www.maths.qmul.ac.uk/~lsoicher/).
Homepage: https://gap-packages.github.io/grape
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Error, Variable: 'IS_MULTI_DIGRAPH' must have a value
not in any function at /proc/cygdrive/C/gap-4.10.1/pkg/digraphs-0.15.0/gap/digraph.gi:525
Error, was not in any namespace at /proc/cygdrive/C/gap-4.10.1/lib/variable.g:291 called from
LEAVE_NAMESPACE(  ); at /proc/cygdrive/C/gap-4.10.1/lib/package.gi:1253 called from
<function "ReadPackage">( <arguments> )
 called from read-eval loop at /proc/cygdrive/C/gap-4.10.1/pkg/digraphs-0.15.0/read.g:26
you can 'quit;' to quit to outer loop, or
you can 'return;' to continue
brk>
```

wilfwilson:
This should hopefully resolve #177.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -13,7 +13,7 @@ PLANAR_INCLUDE   = -I$(top_srcdir)/@PLANARITY_SUITE_DIR@/
 DIGRAPHS_INCLUDE = -I$(top_srcdir)/extern/
 
 BINARCHDIR = bin/$(GAPARCH)
-GAPINSTALLLIB = $(abs_top_srcdir)/$(BINARCHDIR)/
+GAPINSTALLLIB = $(abs_top_srcdir)/$(BINARCHDIR)
 
 lib_LTLIBRARIES = digraphs.la
 
@@ -80,9 +80,10 @@ all-local: digraphs.la
 	$(mkdir_p) $(top_srcdir)/$(BINARCHDIR) $(top_srcdir)/bin/lib
 	cp -RL @PLANARITY_SUITE_DIR@/.libs/*  $(top_srcdir)/bin/lib/
 if SYS_IS_CYGWIN
-	cp .libs/digraphs.dll $(GAPINSTALLLIB)
+	cp .libs/digraphs.dll $(GAPINSTALLLIB)/digraphs.so
+	cp bin/lib/cygplanarity-0.dll $(GAPROOT)/.libs
 else
-	cp .libs/digraphs.so $(GAPINSTALLLIB)
+	cp .libs/digraphs.so $(GAPINSTALLLIB)/
 endif
 if KERNEL_DEBUG
 	@echo "SUCCESS (debug mode enabled)!"
--- a/init.g
+++ b/init.g
@@ -14,13 +14,12 @@ if not IsBound(UserHomeExpand) then
 fi;
 
 # load kernel function if it is installed:
-if (not IsBound(DIGRAPHS_C)) and ("digraphs" in SHOW_STAT()) then
+if not IsBound(DIGRAPH_OUT_NBS) and "digraphs" in SHOW_STAT() then
   # try static module
   LoadStaticModule("digraphs");
 fi;
-if (not IsBound(DIGRAPHS_C)) and
-    (Filename(DirectoriesPackagePrograms("digraphs"), "digraphs.so") <> fail)
-    then
+if not IsBound(DIGRAPH_OUT_NBS) and
+    Filename(DirectoriesPackagePrograms("digraphs"), "digraphs.so") <> fail then
   LoadDynamicModule(Filename(DirectoriesPackagePrograms("digraphs"),
                              "digraphs.so"));
 fi;
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,51 @@
+# set clone depth
+clone_depth: 1                  # clone entire repository history if not defined
+
+environment:
+  TEST_SUITE: test
+  PACKAGES: "latest"
+  SUITE: "test"
+  GAP: "master"
+  DIGRAPHS_LIB: "digraphs-lib-0.5"
+  SETUP: "appveyor"
+
+  # To test building GAP both using the bundled libraries (zlib and GMP), as
+  # well as using the versions distributed with cygwin, we do the former in
+  # the 32 bit build and the latter in the 64 bit build. But building zlib
+  # does not see to work if '--coverage' is used, so we only use that flag in
+  # the 64 bit build.
+  matrix:
+    - CYG_ARCH: x86
+      ABI: 32
+      CYG_ROOT: C:\cygwin
+      PKGS: "-P python27,python27-pip"
+      PKG_FLAGS: "CFLAGS=-m32 CXXFLAGS=-m32 LDFLAGS=-m32 LOPTS=-m32"
+    - CYG_ARCH: x86_64
+      ABI: 64
+      CYG_ROOT: C:\cygwin64
+      PKGS: "-P libgmp-devel,zlib-devel,python27,python27-pip"
+
+## change to packages-stable-X.Y.tar.gz in the stable branch
+#cache:
+#  - packages-master.tar.gz
+
+install:
+  - '%CYG_ROOT%\setup-%CYG_ARCH%.exe -qnNdO -R %CYG_ROOT% -s http://cygwin.mirror.constant.com -l %CYG_ROOT%/var/cache/setup %PKGS%'
+
+# scripts that run after cloning repository
+build_script:
+# If there's a newer build queued for the same PR, cancel this one
+  - ps: if ($env:APPVEYOR_PULL_REQUEST_NUMBER -and $env:APPVEYOR_BUILD_NUMBER -ne ((Invoke-RestMethod `
+        https://ci.appveyor.com/api/projects/$env:APPVEYOR_ACCOUNT_NAME/$env:APPVEYOR_PROJECT_SLUG/history?recordsNumber=50).builds | `
+        Where-Object pullRequestId -eq $env:APPVEYOR_PULL_REQUEST_NUMBER)[0].buildNumber) { `
+        throw "There are newer queued builds for this pull request, failing early." }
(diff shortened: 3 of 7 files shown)


Makefile.am
```diff
@@ -81,6 +81,8 @@ all-local: digraphs.la
 	cp -RL @PLANARITY_SUITE_DIR@/.libs/*  $(top_srcdir)/bin/lib/
 if SYS_IS_CYGWIN
 	cp .libs/digraphs.dll $(GAPINSTALLLIB)
+# Cygwin will only look in this directory for dlls
+	cp bin/libs/cygplanarity-0.dll $(GAPROOT)/.libs
```
wilfwilson: I need to check whether this should actually be `libs` or if it should be `lib` (I can't work out where the "s" would come from, but I wrote "libs" in my comments on #177).

ChrisJefferson: it's bin/lib, no s

init.g
```diff
@@ -14,13 +14,11 @@ if not IsBound(UserHomeExpand) then
 fi;
 
 # load kernel function if it is installed:
-if (not IsBound(DIGRAPHS_C)) and ("digraphs" in SHOW_STAT()) then
+if "digraphs" in SHOW_STAT() then
```
wilfwilson: @james-d-mitchell Do you just want the `DIGRAPHS_C` removing? Or replacing with something else/`DIGRAPHS_C` recreating? What is the purpose of it?

james-d-mitchell: The purpose is to prevent loading the static module, and then the dynamic one. It could be replaced with a check that any thing defined in the kernel module is bound though, such as `DIGRAPH_OUT_NEIGHBOURS`

Makefile.am
```diff
@@ -81,6 +81,8 @@ all-local: digraphs.la
 	cp -RL @PLANARITY_SUITE_DIR@/.libs/*  $(top_srcdir)/bin/lib/
 if SYS_IS_CYGWIN
 	cp .libs/digraphs.dll $(GAPINSTALLLIB)
+# Cygwin will only look in this directory for dlls
+	cp bin/lib/cygplanarity-0.dll $(GAPROOT)/.libs
```
embray: I wouldn't do this.   In general it doesn't make sense to be manually copying anything into `.libs`.  This happens to work, because when you run `./gap` out of a source build you're actually running the [wrapper executable](https://www.gnu.org/software/libtool/manual/html_node/Wrapper-executables.html) generated by libtool, which in turns runs the actual `gap.exe` that is under `.libs`.  

Thus, this happens to work, since the planarity.dll used by the package is now under the gap.exe executable's DLL search path.  

I'll take a closer look at this to make sure my suggestion works, but I think a better approach would be to "install" any DLLs to `$(abs_top_srcdir/$(BINARCHDIR)`.

james-d-mitchell:
I think @embray suggestion is the right approach.

init.g
wilfwilson: I'll add something in as you suggest. I get that the second check for `IsBound(whatever)` would stop the dynamic module being loaded immediately after the static module is loaded, but then I don't get why the line check (line 17) is needed. Or is `init.g` re-read sometimes?

Makefile.am
```diff
@@ -80,7 +80,8 @@ all-local: digraphs.la
 	$(mkdir_p) $(top_srcdir)/$(BINARCHDIR) $(top_srcdir)/bin/lib
 	cp -RL @PLANARITY_SUITE_DIR@/.libs/*  $(top_srcdir)/bin/lib/
 if SYS_IS_CYGWIN
-	cp .libs/digraphs.dll $(GAPINSTALLLIB)
+	cp .libs/digraphs.dll $(GAPINSTALLLIB)/digraphs.so
+	cp bin/lib/cygplanarity-0.dll $(GAPINSTALLLIB)
```
wilfwilson: @embray Is this what you meant?

wilfwilson:
Okay, I'm confused. If someone else can tell me what needs to be done/can make the changes, that would be great, because I'm not sure on my own.

As @ChrisJefferson says in https://github.com/gap-packages/Digraphs/issues/177#issuecomment-471978176
> The list of places that GAP series for dlls is unfortunately not super useful. It will look in the \lib or \bin directory of your cygwin install, or in the '.libs' directory of the GAP install. It doesn't look anywhere inside the digraphs directory.
and
> I think that's the best place to put such things. It seems a bit weird, but it's the easiest way to make stuff work on windows.
Which is why I wrote
```
cp bin/lib/cygplanarity-0.dll $(GAPROOT)/.libs
```
However, doing something else like
```
cp bin/lib/cygplanarity-0.dll $(GAPINSTALLLIB)
```
(where `GAPINSTALLLIB = $(abs_top_srcdir)/$(BINARCHDIR)`) means that Digraphs doesn't load on the Windows machine that I have access to. For now I'll put it back to how it know it works.

Makefile.am
embray: Yes, I see that now.  And yes, I think that should work.

However, I also don't think it's good to rename the .dll file to .so.  Maybe it's fine for now if it's needed as a workaround, but ideally GAP should be fixed to use the right filename extension for loading dynamic libs.

embray:
I'm looking into it.  Frankly a lot of what GAP does with w.r.t. Windows is a mess anyways so it's not your fault if you're confused.  If there's something that needs to be fixed in GAP I can do that; all I'm asking is to not make matters worse :)

Makefile.am
ChrisJefferson: Looking at windows, the places cygwin tries to find this DLL is:

/bin, /lib, /usr/bin and friends (don't really want to start putting things there)

The only places it looks inside the GAP directory are:

the gap root directory and the subdirectories .libs and gap/extern/install/zlib/zlib/lib

ChrisJefferson:
@wilfwilson :  gap-system/gap#3335 fixes problems with doing lots of forking / making new processes running out of memory, which the digraphs and semigroups testsuites both do.

For the directories problem, at the moment putting them in GAP's .lib directory is the only option, because we need to put them somewhere windows looks, and that is the only place windows looks for them.

(If you are interested, there is a program called procmon (process monitor) which microsoft provide which can be used to see what files a program is trying to access. I can see GAP search through various windows and cygwin system directories looking for cygplanarity-0.dll, before it looks in .libs, and then gives up if it doesn't find it)

embray:
My point is that it doesn't "work" except, as I explained, in the specific case where you're running `./gap` out of the GAP source tree.  Please listen to someone who actually uses/develops for GAP on Windows.

wilfwilson:
Thanks for looking everyone. I agree that we should add Windows tests to our CI, but I don't want to spend time working out how to do that atm.

james-d-mitchell:
@embray I'm very happy to listen to what you have to say, and I look forward to seeing a proper solution. We will hold off doing anything with this particular PR, until we can see @embray's PR.

wilfwilson:
@james-d-mitchell I added Windows tests on AppVeyor. I don't have the permissions to plug AppVeyor into this repository, I think I've requested the correct permissions for that to happen, from whomever has the authority, but we'll see. But for now you can see the result of the tests running on my own fork: https://ci.appveyor.com/project/wilfwilson/digraphs/history

wilfwilson:
@james-d-mitchell The Windows AppVeyor tests seem to pass in 64-bit, but not in 32-bit, due to stuff related to I/O: https://ci.appveyor.com/project/wilfwilson/digraphs/builds/23034071. This is unsurprising to me, because when I ran the Digraphs tests on the Windows computer I had access to yesterday, `DigraphsTestStandard` crashed when it got to `io.tst`. I think that making Digraphs work on 32-bit should, however, be a separate issue and PR. I can disable the 32-bit AppVeyor test for now if you want. If we're going to integrate AppVeyor into this repository in the way that Travis is, it will mean that the CI for every PR/branch will fail until 32-bit Digraphs on Windows is resolved.

wilfwilson:
I disabled the 32-bit cygwin tests on AppVeyor because the `io` package doesn't work properly there, and Digraphs fundamentally requires IO. See https://github.com/gap-packages/io/issues/80.

james-d-mitchell:
Thanks @wilfwilson I think disabling the 32-bit cygwin is a good idea, especially since it isn't a bug in Digraphs that's causing this.

james-d-mitchell:
Did you already merge the appveyor stuff? Or is PR coming?

wilfwilson:
It's in this PR, it's ready to go.

james-d-mitchell:
In the absence of a "proper" fix for the issues that this PR addresses, I'm going to suggest that we merge this, and make a release so that Digraphs can be used in cygwin.

PR #180, opened 2019-03-13
Added DigraphNrStronglyConnectedComponents attribute

MTWhyte:
This pull requests adds an attribute DigraphNrStronglyConnectedComponents, which takes a digraph as its argument and returns the number of strongly connected components of the digraph.

The feature was suggested in [this issue](https://github.com/gap-packages/Digraphs/issues/176).

### DIFF
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -666,7 +666,7 @@ rec( comps := [ [ 3 ], [ 1, 2 ] ], id := [ 2, 2, 1 ] )
   <Description>
     This function returns the number of strongly connected components of
     <A>digraph</A>.  Note that this is no more efficient than calling <Ref
-    Attr="DigraphStronglyConnectedComponents"/>.
+    Attr="DigraphStronglyConnectedComponents"/>.<P/>
     
     For more information, see <Ref Attr="DigraphStronglyConnectedComponents"/>.
     <Example><![CDATA[
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -662,7 +662,7 @@ rec( comps := [ [ 3 ], [ 1, 2 ] ], id := [ 2, 2, 1 ] )
 <#GAPDoc Label="DigraphNrStronglyConnectedComponents">
 <ManSection>
   <Attr Name="DigraphNrStronglyConnectedComponents" Arg="digraph"/>
-  <Returns>A positive integer.</Returns>
+  <Returns>A non-negative integer.</Returns>
   <Description>
     This function returns the number of strongly connected components of
     <A>digraph</A>.  Note that this is no more efficient than calling <Ref
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -659,7 +659,26 @@ rec( comps := [ [ 3 ], [ 1, 2 ] ], id := [ 2, 2, 1 ] )
 </ManSection>
 <#/GAPDoc>
 
-#
+<#GAPDoc Label="DigraphNrStronglyConnectedComponents">
+<ManSection>
+  <Attr Name="DigraphNrStronglyConnectedComponents" Arg="digraph"/>
+  <Returns>A positive integer.</Returns>
+  <Description>
+    This function returns the number of strongly connected components of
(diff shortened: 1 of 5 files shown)


MTWhyte:
I’ll double check this in a bit, but I think I ran it on Digraph([[]]) and it returned 1, so I took it that there must always be at least 1 strongly connected component of a graph.

I think that there must always be at least one strongly connected component is consistent with the strongly connected components being partitions of the set of vertices, too?

MTWhyte:
Oh, unless I’m overlooking the digraph with no edges or vertices, if that is a thing? If so, I will fix in a moment.

wilfwilson:
Yeah I think `EmptyDigraph(0);` has no vertices

MTWhyte:
Thanks, I’ll change it in a moment.

tst/standard/attr.tst
```diff
@@ -316,6 +316,23 @@ gap> DigraphStronglyConnectedComponents(gr2);
 rec( comps := [ [ 1 ], [ 2 ], [ 3 ], [ 4 ], [ 5 ], [ 6 ], [ 7 ], [ 8 ], 
       [ 9 ], [ 10 ] ], id := [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 ] )
 
+#  DigraphNrStronglyConnectedComponents
+gap> D := CycleDigraph(10);;
+gap> for i in [1 .. 1000] do
```
wilfwilson: Can you please decrease this number to something much smaller, like `50` or something. For me this code takes almost half a second, and ideally we'd like each standard test file to take only about a second or two in total.

doc/attr.xml
```diff
@@ -659,7 +659,26 @@ rec( comps := [ [ 3 ], [ 1, 2 ] ], id := [ 2, 2, 1 ] )
...
 
-#
+<#GAPDoc Label="DigraphNrStronglyConnectedComponents">
+<ManSection>
+  <Attr Name="DigraphNrStronglyConnectedComponents" Arg="digraph"/>
+  <Returns>A non-negative integer.</Returns>
+  <Description>
+    This function returns the number of strongly connected components of
+    <A>digraph</A>.  Note that this is no more efficient than calling <Ref
+    Attr="DigraphStronglyConnectedComponents"/>.
```
wilfwilson: If you intend for this to be the end of a paragraph, you need a `<P/>` tag at the end here.

gap/attr.gi
```diff
@@ -476,6 +476,12 @@ function(digraph)
   return GABOW_SCC(OutNeighbours(digraph));
 end);
 
+InstallMethod(DigraphNrStronglyConnectedComponents, "for a digraph",
+[IsDigraph],
+function(digraph)
+  return Length(DigraphStronglyConnectedComponents(digraph).comps);
```
wilfwilson: This works fine, but when you have a one line function like this, instead of using three lines as you have, you can write it shorter (and I think it's what we tend to do in the digraphs package):
```
InstallMethod(DigraphNrStronglyConnectedComponents, "for a digraph",
[IsDigraph],
digraph -> Length(DigraphStronglyConnectedComponents(digraph).comps));
```

tst/standard/attr.tst
```diff
@@ -316,6 +316,23 @@ gap> DigraphStronglyConnectedComponents(gr2);
...
+1001
+gap> D := CayleyDigraph(SymmetricGroup(6));;
+gap> DigraphNrStronglyConnectedComponents(D);
+1
+gap> D := Digraph([[2, 3], [1, 4, 5], [3], [2, 5], [6], [3, 5]]);;
+gap> DigraphNrStronglyConnectedComponents(D);
+3
+gap>  D := Digraph([[]]);;
+gap> DigraphNrStronglyConnectedComponents(D);
+1
```
wilfwilson: Please also add a test for `DigraphNrStronglyConnectedComponents(EmptyDigraph(0));` 🙂

MTWhyte:
Thanks for your comments, I've implemented and pushed them :)

PR #181, opened 2019-03-14
Added PetersenGraph function, along with exmpl file family

MTWhyte:
This pull request adds a function which returns the PetersenGraph, and introduces a set of files "exmpl.*" for adding more graph examples in the future.


wilfwilson:
Can you please use the name `examples` rather than `exmpl`?  I also wonder if there are other 'examples' in the package that would be better located in this new file – perhaps `JohnsonDigraph`? Perhaps you could have a look.

james-d-mitchell:
I've been using the name `exmpl` in the `mutable` PR, and have rearranged much of the code (into different files), it'd probably be best if @MTWhyte doesn't make too many changes in this PR, otherwise it'll be a nightmare to merge `mutable` later.

wilfwilson:
Okay, in that case I would still urge us to change the name at some later point after the merge. I think it's unhelpful to miss out the vowels.

james-d-mitchell:
You know I have an unreasonable dislike of vowels, but I'm happy to rename the file too. I'm guessing that you won't be a fan of `cnstr` either :(

wilfwilson:
`cnstr`... ouch!

james-d-mitchell:
You mean: ch!

MTWhyte:
I remember seeing a list of examples, when I added PetersenGraph to the documentation.

In `z-chap2.xml`, there is:

`  <Section><Heading>Standard examples</Heading>
    <#Include Label="ChainDigraph">
    <#Include Label="CompleteDigraph">
    <#Include Label="CompleteBipartiteDigraph">
    <#Include Label="CompleteMultipartiteDigraph">
    <#Include Label="CycleDigraph">
    <#Include Label="EmptyDigraph">
    <#Include Label="JohnsonDigraph">
    <#Include Label="PetersenGraph">
  </Section>`

I shall migrate all of these to the exmpl files.

james-d-mitchell:
Please don't move any @MTWhyte, see the comments above!

MTWhyte:
Ah right, I will leave as is then, but I’ll keep those listed in mind for the future.

PR #185, opened 2019-03-19
io: make changes for mutable digraphs (unfinished)

ChristopherRussell:
ReadDigraphs - currently the decoder creates the digraph. I wanted to change the decoder to return the OutNeighbours so that ReadDigraphs could create the digraph and thus decide whether or not to make it mutable. We wanted to add a new argument called `constructor` which defaults to `Digraph` but can be set to `MutableDigraph`.

IO_Unpicklers.DIGG - I was thinking about making this return a mutable digraph but I don't know if it's worth it - is the point of this method to store certain attributes and we'd rather do something else if not storing those attributes?

I think I left off unsure about how to approach the decoder functions - at least thats what my notes say from that day.

### DIFF
--- a/gap/io.gi
+++ b/gap/io.gi
@@ -140,7 +140,7 @@ IO_Unpicklers.DIGG := function(file)
     od;
   od;
 
-  digraph := DigraphNC(out);
+  digraph := MutableDigraphNC(out);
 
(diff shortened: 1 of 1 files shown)


james-d-mitchell:
I am working on completing this so will close the PR and add it to my PR for mutable digraphs.

PR #186, opened 2019-03-19
Add edge-coloured digraph automorphism finding method

flsmith:
Resolves https://github.com/gap-packages/Digraphs/issues/183.


gap/isomorph.gi
```diff
@@ -561,6 +612,140 @@ function(n, partition, method)
...
+    for i in adj_colours do
+      if not IsPosInt(i) then
+        ErrorNoReturn("the 2nd argument should be a list of lists of",
+                      "positive integers");
+      fi;
+      AddSet(colours, i);
+    od;
+  od;
+  m := Length(colours);
+  if ForAny([1 .. m], i -> i <> colours[i]) then
```
ChrisJefferson: It won't make much difference, but you could write `if colours = [1..m] ...` here.

markuspf:
It looks like the colouring consistency check takes a really long time: I made an example with 3500 vertices and it doesn't get to computing the autmorphism group at all.

wilfwilson:
Thanks Markus, really useful observation. Hopefully we’ll be able to sort that out.

flsmith:
Fix will come tonight :)

james-d-mitchell:
Any chance you could fix the CI @flsmith and rebase onto the current `master` branch??

flsmith:
I realised recently that this reduction is invalid for multidigraphs (even with unique edge colours) and haven't found a way to fix it efficiently yet.

ChrisJefferson:
When I've done something similar in the past, I first made a function from edge-labelled multidigraphs to edge-labelled digraphs, by making a pass over the graph and making new a new label whenever I find a new multiset of edges. I think a function that did that might have other uses, so it's probably not a silly thing to write anyway.

It's probably not possible to do this (easily) in a single pass -- you possibly could do something crazy, where you only add a new layer when you realise you need one as the number of edge labels grows, but I certainly wouldn't do that as a first pass

james-d-mitchell:
Cough cough @flsmith

james-d-mitchell:
For the record I think this should be rebased, and then hopefully the CI will pass.

james-d-mitchell:
I'm happy with this, and think it should be merged sooner rather than later.

.travis.yml
```diff
@@ -82,8 +82,8 @@ matrix:
       - ABI=64
       - GAP=master
       - PACKAGES=required
-      - THRESHOLD=98
-
+      - THRESHOLD=94
```
wilfwilson: This is... unfortunate. Is there a deep reason why you can't get better code coverage in `isomorph.gi` via `isomorph.tst`, or is it just that you've not got round to it?

If you're not going to get the coverage up, I'd prefer you to put this back up to `98` anyway, and instead put `isomorph.tst` into `.covignore`.

doc/isomorph.xml
```diff
@@ -41,7 +41,7 @@
 
 <#GAPDoc Label="BlissAutomorphismGroup">
 <ManSection>
-  <Attr Name="BlissAutomorphismGroup" Arg="digraph[, colours]"/>
+  <Attr Name="BlissAutomorphismGroup" Arg="digraph[, vertex_colours, edge_colours]"/>
```
wilfwilson: Does this have to be used with either both optional arguments or none, or is just vertex colours okay? In which case:
```suggestion
  <Attr Name="BlissAutomorphismGroup" Arg="digraph[, vertex_colours[, edge_colours]]"/>
```

doc/isomorph.xml
```diff
@@ -268,23 +206,93 @@ true
...
+      <Item>
+        <A>edge_colours</A> must be fail or a list of <C>n</C> lists of integers
+        of the same shape as <C>OutNeighbours(digraph)</C>, where
+        <A>edge_colours</A><C>[i][j]</C> is the colour of the edge
+        <C>OutNeighbours(digraph)[i][j]</C>, using the colours <C>[1 .. k]</C>
+        for some <C>k &lt;= n</C>;
+      </Item>
+    </List>
+
+    Gving <A>vert_colours</A> [<A>edge_colours</A>] as <C>fail</C> is equivalent
```
wilfwilson:
```suggestion
    Giving <A>vert_colours</A> [<A>edge_colours</A>] as <C>fail</C> is equivalent
```

doc/isomorph.xml
```diff
@@ -54,6 +54,9 @@
     the value of either <C>BlissAutomorphismGroup</C> or  <Ref
       Attr="NautyAutomorphismGroup"/>. These groups are, of course, equal but
     their generating sets may differ. <P/>
+    The attribute <Ref Attr="AutomorphismGroup" Label="for a digraph,
+      homogeneous list, and list" /> returns the value of
+    <C>BlissAutomorphismGroup</C> as it is not implemented for &nauty;
```
wilfwilson: Perhaps say that that is where you can find the requirements of the optional arguments to this function?

Also, the three-argument version of `BlissAutomorphismGroup` is not an attribute, so I think it should be documented more like:
```suggestion
  <Attr Name="BlissAutomorphismGroup" Arg="digraph"/>
  <Oper Name="BlissAutomorphismGroup" Arg="digraph, vertex_colours, edge_colours"/>
```
I presume - or something like that (perhaps with a two-argument version, if applicable, as mentioned above). Perhaps they will need labels.

doc/isomorph.xml
```diff
@@ -54,6 +54,9 @@
     the value of either <C>BlissAutomorphismGroup</C> or  <Ref
       Attr="NautyAutomorphismGroup"/>. These groups are, of course, equal but
     their generating sets may differ. <P/>
+    The attribute <Ref Attr="AutomorphismGroup" Label="for a digraph,
```
wilfwilson: You can't have line-breaks in the label bit of a `Ref` tag - GAP doc gives warnings.

doc/isomorph.xml
```diff
@@ -268,23 +206,93 @@ true
...
+    for more information about isomorphisms of coloured digraphs. <P/>
+
+    If <A>digraph</A> is not a multidigraph then the automorphism group is
+    returned as a group of permutations on the set of vertices of
+    <A>digraph</A>. <P/>
+
+    If <A>digraph</A> is a multidigraph then the automorphism group is returned
+    as the direct product of a group of permutations on the set of vertices of
+    <A>digraph</A> with a group of permutations on the set of edges of
+    <A>digraph</A>. These groups can be accessed using <Ref Attr="Projection"/>
```
wilfwilson: All references to `Projection` should probably be
```
<Ref Oper="Projection" Label="for a domain and a positive integer" BookName="ref"/>
```

doc/isomorph.xml
```diff
@@ -268,23 +206,93 @@ true
...
+    This operation computes the automorphism group of a vertex- and/or
+    edge-coloured digraph.
+    A coloured digraph can be specified by its underlying digraph
+    <A>digraph</A> and colourings <A>vert_colours</A>, <A>edge_colours</A>.
+    Let <C>n</C> be the number of vertices of <A>digraph</A>. The colourings
+    must have the following forms:
+
+    <List>
+      <Item>
+        <A>vert_colours</A> must be <C>fail</C> or a list of <C>n</C> integers,
```
wilfwilson:
```suggestion
        <A>vert_colours</A> must be <K>fail</K> or a list of <C>n</C> integers,
```

doc/isomorph.xml
```diff
@@ -268,23 +206,93 @@ true
...
+        where <A>vert_colours</A><C>[i]</C> is the colour of vertex <C>i</C>,
+        using the colours <C>[1 .. m]</C> for some <C>m &lt;= n</C>;
+      </Item>
+      <Item>
+        <A>edge_colours</A> must be fail or a list of <C>n</C> lists of integers
```
wilfwilson:
```suggestion
        <A>edge_colours</A> must be <K>fail</K> or a list of <C>n</C> lists of integers
```

.travis.yml
james-d-mitchell: I'm not sure what is going on here, but @flsmith agreed to try to remove this change yesterday, if I remember correctly.

flsmith: Caused by my confusion about nauty and now fixed

doc/isomorph.xml
flsmith: Just vertex colour is okay, fixed

doc/isomorph.xml
flsmith: Agreed and fixed

PR #194, opened 2019-04-10
Added DigraphMycielskian function

MTWhyte:
This pull request adds a function which takes a symmetric digraph, and returns its Mycielskian.

The Mycielskian of a symmetric digraph is a larger symmetric digraph constructed from it, which has greater chromatic number. For more information on the construction, see [here](https://en.wikipedia.org/wiki/Mycielskian).

### DIFF
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -1241,6 +1241,32 @@ gap> DigraphLoops(gr);
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="DigraphMycielskian">
+<ManSection>
+  <Attr Name="DigraphMycielskian" Arg="digraph"/>
+  <Returns>A digraph.</Returns>
+  <Description>
+    If <A>digraph</A> is a symmetric digraph, then <C>DigraphMycielskian</C> returns the Mycielskian of <A>digraph</A>. If <A>digraph</A> is not symmetric, then an error message is returned.<P/>
+    
+   The Mycielskian of a symmetric digraph is a larger symmetric digraph constructed from it, which has a larger chromatic number. For further information, see <URL> https://en.wikipedia.org/wiki/Mycielskian </URL>. <P/>
+
+    <Example><![CDATA[
+gap> D := Digraph([[2], [1]]);
+<immutable digraph with 2 vertices, 2 edges>
+gap> ChromaticNumber(D);
+2
+gap> D := DigraphMycielskian(D);
+<immutable digraph with 5 vertices, 10 edges>
+gap> ChromaticNumbmvimer(D);
+3
+gap> D := DigraphMycielskian(D);
+<immutable digraph with 11 vertices, 40 edges>
+gap> ChromaticNumber(D);
+4]]></Example>
+  </Description>
+</ManSection>
+<#/GAPDoc>
+
 <#GAPDoc Label="DigraphDegeneracy">
 <ManSection>
   <Attr Name="DigraphDegeneracy" Arg="digraph"/>
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -1246,12 +1246,12 @@ gap> DigraphLoops(gr);
   <Attr Name="DigraphMycielskian" Arg="digraph"/>
   <Returns>A digraph.</Returns>
   <Description>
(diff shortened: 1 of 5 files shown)


doc/attr.xml
```diff
@@ -1241,6 +1241,32 @@ gap> DigraphLoops(gr);
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="DigraphMycielskian">
+<ManSection>
+  <Attr Name="DigraphMycielskian" Arg="digraph"/>
+  <Returns>A digraph.</Returns>
+  <Description>
+    If <A>digraph</A> is a symmetric digraph, then <C>DigraphMycielskian</C> returns the Mycielskian of <A>digraph</A>. If <A>digraph</A> is not symmetric, then an error message is returned.<P/>
```
wilfwilson: Could you please delete the second sentence here? We (seem to) have adopted the convention in this package that we don't include such sentences, describing than an error is given when the argument doesn't match the requirements.

doc/attr.xml
```diff
@@ -1241,6 +1241,32 @@ gap> DigraphLoops(gr);
...
+    
+   The Mycielskian of a symmetric digraph is a larger symmetric digraph constructed from it, which has a larger chromatic number. For further information, see <URL> https://en.wikipedia.org/wiki/Mycielskian </URL>. <P/>
```
wilfwilson: Could you please have the `<URL>` tags tight against the URL, i.e.
`<URL>https://en.wikipedia.org/wiki/Mycielskian</URL>`

doc/attr.xml
```diff
@@ -1241,6 +1241,32 @@ gap> DigraphLoops(gr);
...
+
+    <Example><![CDATA[
+gap> D := Digraph([[2], [1]]);
```
wilfwilson: Could you change this to `CycleDigraph(2)`; it's the same digraph, but perhaps makes it slightly easier for a reader to understand what graph it is 🙂

gap/attr.gi
```diff
@@ -1336,6 +1336,42 @@ end);
 # It is the backend to IsBipartiteDigraph, Bicomponents, and DigraphColouring
 # for a 2-colouring
 
+InstallMethod(DigraphMycielskian, "for a symmetric digraph",
```
wilfwilson: I think we really need a different version of this for mutable and immutable digraphs; the one for immutable digraphs would look like this:
```
if not IsSymmetricDigraph(D) or IsMultiDigraph(D) then
  (blah)
fi;
return MakeImmutable(DigraphMycielskian(DigraphMutableCopy(D)));
```
And the mutable version would be the version that exists at the moment.

As things stand at the moment, for an immutable digraph, the current function makes a new copy in memory of the existing digraph every time you add a vertex or an edge, which happens a large number of times and harms performance.

tst/standard/attr.tst
```diff
@@ -1175,6 +1175,28 @@ gap> for i in [1 .. 50] do
...
+gap> D := DigraphMycielskian(D);
+<immutable digraph with 17 vertices, 112 edges>
+gap> ChromaticNumber(D);
+5
+gap> D1 := Digraph([[], [3], [2]]);
+<immutable digraph with 3 vertices, 2 edges>
+gap> D2 := Digraph([[], [3, 4], [2, 5], [2, 6], [3, 6], [4, 5, 7], [6]]);
+<immutable digraph with 7 vertices, 12 edges>
+gap> IsIsomorphicDigraph(DigraphMycielskian(D1), D2);
+true
```
wilfwilson: Could you also add a couple of tests of running this on a mutable digraph?

gap/attr.gi
```diff
@@ -1336,6 +1336,42 @@ end);
...
+[IsDigraph],
+function(D)
+  local n, i, j;
+  if not IsSymmetricDigraph(D) or IsMultiDigraph(D) then
```
wilfwilson: What does/should this do to digraphs that have loops?

MTWhyte: That's a good point! If we take a graph with loops like `D := Digraph([[1, 2], [1]])`, then DigraphMycielskian(D) has a repeated edge, which I don't want. I'll fix this.

wilfwilson:
Thanks for the changes @MTWhyte

PR #198, opened 2019-06-05
Use MakeString instead of C_NEW_STRING

fingolfin:
The latter is kinda deprecated (I'll add a comment to the GAP header `stringobj.h` making that explicit.

This PR could also use `MakeImmString` instead of `MakeString`, to produce an immutable string; but I played it safe, and made sure to not change the semantics of the code.

### DIFF
--- a/src/homos.c
+++ b/src/homos.c
@@ -1926,8 +1926,7 @@ Obj FuncHomomorphismDigraphsFinder(Obj self, Obj args) {
     }
   }
 
-  Obj str;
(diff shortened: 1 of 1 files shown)


james-d-mitchell:
Thanks @fingolfin this seems like an eminently sensible change.

wilfwilson:
Thanks @fingolfin, I've merged. `C_NEW_STRING` was introduced only a few months ago in #160, and had already been refactored out of `master`. Unless you think it's urgent, I won't hurry out a release.

fingolfin:
Not urgent

PR #200, opened 2019-06-05
Fix DIGRAPHS_Bipartite for digraphs with isolated vertices

### LINKED ISSUE #199: Bug with IsBipartiteDigraph
There is a bug with the function IsBipartiteDigraph, which gives an error when applied as follows:

~~~~
gap> D := Digraph([[4, 6, 8], [], [], [], [7], [2], [], [], [8], []]);
<immutable digraph with 10 vertices, 6 edges>
gap> IsBipartiteDigraph(D);
Error, the argument does not describe a transformation at /Users/murraywhyte/gap/lib/trans.gi:111 called from
Transformation( colours
 ) at /Users/murraywhyte/gap/pkg/Digraphs/gap/attr.gi:1435 called from
DIGRAPHS_Bipartite( D ) at /Users/murraywhyte/gap/pkg/Digraphs/gap/prop.gi:390 called from
<function "unknown">( <arguments> )
 called from read-eval loop at *stdin*:20
brk>
~~~~

wilfwilson:
This code was recently refactored; the problem is that the assignment of colours to vertices didn't give a colour to isolated vertices. This bug was only ever present in `master`, it wasn't released.

Resolves #199.

### DIFF
--- a/doc/grahom.xml
+++ b/doc/grahom.xml
@@ -410,7 +410,7 @@ gap> DigraphColouring(ChainDigraph(10), 1);
 fail
 gap> D := ChainDigraph(10);;
 gap> t := DigraphColouring(D, 2);
-Transformation( [ 2, 1, 2, 1, 2, 1, 2, 1, 2, 1 ] )
+Transformation( [ 1, 2, 1, 2, 1, 2, 1, 2, 1, 2 ] )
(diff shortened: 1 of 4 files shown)


james-d-mitchell:
Thanks @wilfwilson

Issue #205, opened 2019-06-17
configure option to use external bliss and planarity


isuruf:

james-d-mitchell:
For reference, the suggestion is to do something like what was done for Semigroups in:


https://github.com/gap-packages/Semigroups/pull/590
https://github.com/gap-packages/Semigroups/pull/593
https://github.com/gap-packages/Semigroups/pull/595

james-d-mitchell:
Hmm, so, one problem is that the released version of bliss had some bugs in it, which we resolved in the version that's included with Digraphs. So, if we use the system bliss these problems will still exist, I think the same was true for libplanarity (although to a lesser extent). I'm not sure how to resolve this as neither bliss nor libplanarity have public repos where issues can be reported, as far as I am aware.

isuruf:
We can patch bliss and libplanarity with the fixes here for the system version. (extern issue is already fixed with planarity system copy I'm using)

james-d-mitchell:
Ok awesome, I'll work on a PR to use the system planarity just now.

james-d-mitchell:
@isuruf if you make a bliss conda feedstock, then I can apply the patch we used in Digraphs.

isuruf:
Fixed by https://github.com/gap-packages/Digraphs/pull/207. Thanks @james-d-mitchell

isuruf:
#207 only fixes planarity.

dimpase:
We should merge bliss fixes from https://github.com/mkoeppe/bliss and the ones here, to create a "canonical" version.

isuruf:
@james-d-mitchell, we also have https://github.com/conda-forge/bliss-feedstock which uses  https://github.com/mkoeppe/bliss

james-d-mitchell:
Agreed @dimpase it'd be good to have a merged version of bliss with all existing fixes included.

wilfwilson:
@isuruf added the `--with-external-bliss` option in #225, which was included in the v1.0.0 release, so I'll close this now.

PR #207, opened 2019-06-17
System planarity

james-d-mitchell:
This PR is similar to:

gap-packages/Semigroups#590
gap-packages/Semigroups#593
gap-packages/Semigroups#595

and partially resolves Issue #205. Thanks to @embray for help with this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,6 +79,28 @@ matrix:
       - GAP=master
       - PACKAGES=required
       - THRESHOLD=98
+    
+    - env:
+      - SUITE=external-planarity
+      - ABI=64
+      - GAP=stable-4.10
+      - PACKAGES=required
+      - EXTRA_PKG_FLAGS="--with-external-planarity"
+      - LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$HOME/miniconda2/lib
+      - CFLAGS="$CFLAGS -I$HOME/miniconda2/include -O2 -g"
+      - LDFLAGS="$LDFLAGS -L$HOME/miniconda2/lib"
+
+before_install:
+  - if [ "$SUITE" == "external-planarity" ]; then
+      wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh;
+      chmod +x miniconda.sh;
+      ./miniconda.sh -b;
+      export PATH=/home/travis/miniconda2/bin:$PATH;
+      conda update --yes conda;
+      conda config --add channels conda-forge;
+      conda install --yes planarity;
+      conda install --yes readline;
+    fi;
 
 install:
   # GAP and some packages require special flags for compilation in 32-bit mode
--- a/Makefile.am
+++ b/Makefile.am
@@ -7,9 +7,11 @@
 
 ACLOCAL_AMFLAGS = -I m4
 
-SUBDIRS = @PLANARITY_SUITE_DIR@
+if WITH_INCLUDED_PLANARITY 
+  SUBDIRS = @PLANARITY_SUITE_DIR@
+  PLANAR_INCLUDE = -I$(top_srcdir)/@PLANARITY_SUITE_DIR@/
+endif
 
-PLANAR_INCLUDE   = -I$(top_srcdir)/@PLANARITY_SUITE_DIR@/
 DIGRAPHS_INCLUDE = -I$(top_srcdir)/extern/
 
 BINARCHDIR = bin/$(GAPARCH)
@@ -72,7 +74,9 @@ digraphs_la_CFLAGS += $(WARNING_CFLAGS)
 
 digraphs_la_LDFLAGS  =  $(GAP_LDFLAGS) -module -avoid-version
 
+if WITH_INCLUDED_PLANARITY
 digraphs_la_LIBADD = @PLANARITY_SUITE_DIR@/libplanarity.la
+endif
 
 if SYS_IS_CYGWIN
 digraphs_la_LDFLAGS += -no-undefined -version-info 0:0:0 -Wl,$(GAPROOT)/bin/$(GAPARCH)/gap.dll
@@ -80,10 +84,14 @@ endif
 
 all-local: digraphs.la
 	$(mkdir_p) $(top_srcdir)/$(BINARCHDIR) $(top_srcdir)/bin/lib
+if WITH_INCLUDED_PLANARITY
 	cp -RL @PLANARITY_SUITE_DIR@/.libs/*  $(top_srcdir)/bin/lib/
+endif
 if SYS_IS_CYGWIN
 	cp .libs/digraphs.dll $(GAPINSTALLLIB)/digraphs.so
+if WITH_INCLUDED_PLANARITY
 	cp bin/lib/cygplanarity-0.dll $(GAPROOT)/.libs
+endif
 else
 	cp .libs/digraphs.so $(GAPINSTALLLIB)/
 endif
--- a/configure.ac
+++ b/configure.ac
@@ -15,11 +15,6 @@ AC_CONFIG_HEADER([src/_pkgconfig.h:cnf/pkgconfig.h.in])
 AC_CONFIG_MACRO_DIR([m4])
 AC_CONFIG_AUX_DIR([cnf])
 
-AC_CONFIG_SUBDIRS([extern/edge-addition-planarity-suite-Version_3.0.0.5])
-
-PLANARITY_SUITE_DIR=extern/edge-addition-planarity-suite-Version_3.0.0.5
-AC_SUBST(PLANARITY_SUITE_DIR)
-
 dnl ##
 dnl ## Get canonical host info
 dnl ##
@@ -90,6 +85,8 @@ AM_CONDITIONAL([ENABLE_STATS], [test "x$enable_stats" = xyes])
 
 AX_CODE_COVERAGE()
 
+# Enable compiler warnings 
+
 AC_ARG_ENABLE([compile-warnings],
     [AS_HELP_STRING([--enable-compile-warnings], [enable compiler warnings])],
     [enable_compile_warnings=yes],
@@ -110,6 +107,16 @@ AS_IF([test "x$enable_compile_warnings" != "xno"],
        -pedantic dnl
        -Wold-style-cast])
    AC_SUBST([WARNING_CFLAGS])])
+
+# Check for system planarity suite
+
+AX_CHECK_PLANARITY()
+
+if test "x$with_external_planarity" = xno ; then
+  AC_CONFIG_SUBDIRS([extern/edge-addition-planarity-suite-Version_3.0.0.5])
+  PLANARITY_SUITE_DIR=extern/edge-addition-planarity-suite-Version_3.0.0.5
+  AC_SUBST(PLANARITY_SUITE_DIR)
+fi
 
 dnl compiler builtins
 AC_DEFUN([CHECK_COMPILER_BUILTIN],
--- a/m4/ax_check_planarity.m4
+++ b/m4/ax_check_planarity.m4
@@ -1,0 +1,33 @@
+dnl handle planarity checks
+dnl
+dnl if --with-external-planarity is supplied,
+dnl use it if it is known to pkg-config and is new enough;
+dnl otherwise use the included version
+dnl
+AC_DEFUN([AX_CHECK_PLANARITY], [
+  AC_ARG_WITH([external-planarity],
+	      [AC_HELP_STRING([--with-external-planarity],
+			      [use external planarity])],
+              [],
+              [with_external_planarity=no])
+  AC_MSG_CHECKING([whether to use external planarity])
+  AC_MSG_RESULT([$with_external_planarity])
+  if test "x$with_external_planarity" = xyes ; then
+        AC_LANG_PUSH([C])
+        AC_CHECK_LIB([planarity], 
+                     [gp_InitGraph], 
+                     [],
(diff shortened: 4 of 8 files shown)


james-d-mitchell:
The motivation behind this is to make GAP and in particular Digraphs easier to use with conda.

wilfwilson:
This will need rebasing and a couple of tweaks to the Travis setup (it is not expecting an 'external-libsemigroups' suite yet).

james-d-mitchell:
You mean an "external-planarity" suite, right? I don't find any mention of libsemigroups

wilfwilson:
Yeah sorry, brainfart!

wilfwilson:
You should also change `scripts/travis-test.sh:34` to say:
```
elif [ "$SUITE" == "test" ] || [ "$SUITE" == "external-planarity" ] ; then
```
then I think it should all work.

james-d-mitchell:
Hmm it is still not working, can you tell what I did wrong?

wilfwilson:
Yes, in `travis-build-dependencies.sh`:
```
[ "$SUITE" != "lint" ] && ["$SUITE" != "external-planarity"]
```
There should be whitespace padding the square brackets, like with the linting bit:
```
[ "$SUITE" != "lint" ] && [ "$SUITE" != "external-planarity" ]
```

james-d-mitchell:
Thanks! Fixed.

james-d-mitchell:
Not fixed, what am I doing wrong now?

.travis.yml
```diff
@@ -79,6 +79,27 @@ matrix:
...
       - THRESHOLD=98
+    
+    - env:
+      - SUITE=external-planarity
+      - ABI=64
+      - GAP=stable-4.10
+      - PACKAGES=required
+      - EXTRA_PKG_FLAGS="--with-external-planarity"
+      - LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$HOME/miniconda2/lib
+      - CFLAGS=$CFLAGS:-I$HOME/miniconda2/include
```
isuruf: This needs to be `CFLAGS="$CFLAGS -I$HOME/miniconda2/include"`

james-d-mitchell:
@wilfwilson ready to go?

wilfwilson:
Do you want your new travis job to run the tests, or are you happy just to check that it compiles? Currently it's just installing everything and then finishing - see https://github.com/gap-packages/Digraphs/pull/207#issuecomment-503002151

james-d-mitchell:
Ah sorry I misread your earlier comment, yes, I think it should also run the tests. Let me amend just now.

wilfwilson:
Everything seems to work properly now, which is great.

However, annoyingly, the new job takes 33 minutes - more than twice as long as the next longest job https://travis-ci.org/gap-packages/Digraphs/jobs/547604896

Can you think why it would be so much slower? Our Travis setup already takes an annoying long time, and I'm not super happy about making it 33 minutes slower. If it's just necessarily going to be slower, then maybe we should only run the standard tests or something, but perhaps there's something wrong if it's so slow.

Anyway, I was wondering what your thoughts are about this.

embray:
That's strange that it's so much slower.  My first guess might have been that just installing miniconda and the needed dependencies was adding more time, but I didn't think it would be that much.  Indeed looking at [this log](https://travis-ci.org/gap-packages/Digraphs/jobs/547604896) it only adds about 2 minutes. while most of the time was spent in the

    $ scripts/travis-build-digraphs.sh && scripts/travis-test.sh

step.  Not clear why it would be so much slower.  Can that be reproduced locally?

embray:
I see now looking at the logs that many of the GAP tests performed, in particular the Digraphs tests, are just actually running several times slower, often as much as 5 or 6 times slower, on the --external-planarity configuration than otherwise.

Could it have something to do with the planarity package on conda?  Or even just building/running GAP in conda in general?  Maybe there are some dependencies (e.g. a different BLAS?) that would improve matters?  I just don't know what would cause GAP to run so much slower or if that's a normal, reproducible thing in conda...

embray:
Well, good news is I was able to reproduce the problem:  Tests were significantly slower in a GAP+diagraphs built and running in a conda environment, by roughly the same factors as we're seeing on Travis.

Bad news is that means it's not just some CI oddity, and is a real problem, and one that's very non-obvious.

Is there some good way I can benchmark GAP itself and/or the GC to see if there's a deeper problem here outside the digraphs package?

embray:
Just as one point of comparison, the first run of the test suite with bliss results, in my non-conda build:

```
total      8461 ms (1409 ms GC) and 1.05GB allocated
              0 failures in 15 files
```

while in my conda build:

```
total     39102 ms (3608 ms GC) and 1.05GB allocated
              0 failures in 15 files
```

james-d-mitchell:
Hmm, I agree this is nuts, and we shouldn't merge this PR until this has been resolved. I will investigate when I get the chance, but I can't do it immediately, so any input from anyone else will be gratefully received.

embray:
If there is a reason to move forward on this sooner-rather-than-later, at least this PR provides the necessary way forward.  Also it would make a big difference just to drop `DigraphsTestExtreme();` in this case.  That said, for the time being, that was be just brushing the problem under the rug.

embray:
FWIW these are the libraries I have loaded in the non-conda build:

```
0x00007f437f527f40  0x00007f437f57c42f  Yes (*)     /usr/lib/x86_64-linux-gnu/libgmp.so.10
0x00007f437f307e00  0x00007f437f317bf8  Yes (*)     /lib/x86_64-linux-gnu/libz.so.1
0x00007f437f0d30d0  0x00007f437f0f3ee5  Yes         /lib/x86_64-linux-gnu/libreadline.so.6
0x00007f437edbf610  0x00007f437ee2e056  Yes         /lib/x86_64-linux-gnu/libm.so.6
0x00007f437ebb6ed0  0x00007f437ebb79ce  Yes         /lib/x86_64-linux-gnu/libdl.so.2
0x00007f437e9b3f10  0x00007f437e9b4804  Yes         /lib/x86_64-linux-gnu/libutil.so.1
0x00007f437e609520  0x00007f437e751e03  Yes         /lib/x86_64-linux-gnu/libc.so.6
0x00007f437e3cd3d0  0x00007f437e3d9028  Yes (*)     /lib/x86_64-linux-gnu/libtinfo.so.5
0x00007f437f793b00  0x00007f437f7ae5b0  Yes         /lib64/ld-linux-x86-64.so.2
0x00007f437e1b38a0  0x00007f437e1ba0e8  Yes         /home/embray/src/gap-system/gap-noconda.tmp/pkg/io-4.5.1/bin/x86_64-pc-linux-gnu-default64-kv3/io.so
0x00007f437df90d90  0x00007f437dfa5fd2  Yes (*)     /home/embray/src/gap-system/gap-noconda.tmp/pkg/nautytraces/bin/x86_64-pc-linux-gnu-default64-kv3/NautyTracesInterface.so
0x00007f437dd87550  0x00007f437dd8bd88  Yes         /home/embray/src/gap-system/gap-noconda.tmp/pkg/orb-4.8.2/bin/x86_64-pc-linux-gnu-default64-kv3/orb.so
0x00007f437db45110  0x00007f437db6b95d  Yes         /home/embray/src/gap-system/gap-noconda.tmp/pkg/digraphs/bin/x86_64-pc-linux-gnu-default64-kv3/digraphs.so
0x00007f437d913300  0x00007f437d934714  Yes (*)     /home/embray/src/gap-system/gap-noconda.tmp/pkg/digraphs/extern/edge-addition-planarity-suite-Version_3.0.0.5/.libs/libplanarity.so.0
0x00007f437d662620  0x00007f437d6c5803  Yes (*)     /usr/lib/x86_64-linux-gnu/libstdc++.so.6
0x00007f437d3f3ab0  0x00007f437d403985  Yes (*)     /lib/x86_64-linux-gnu/libgcc_s.so.1
```

And these are the ones I have on the conda build:

```
0x00007fbe7e81ff40  0x00007fbe7e87442f  Yes (*)     /usr/lib/x86_64-linux-gnu/libgmp.so.10
0x00007fbe7e5fc400  0x00007fbe7e61017c  Yes (*)     /home/embray/src/gap-system/miniconda2/lib/libz.so.1
0x00007fbe7ec67040  0x00007fbe7ec8b75b  Yes (*)     /home/embray/src/gap-system/miniconda2/lib/libreadline.so.7
0x00007fbe7e2f9610  0x00007fbe7e368056  Yes         /lib/x86_64-linux-gnu/libm.so.6
0x00007fbe7e0f0ed0  0x00007fbe7e0f19ce  Yes         /lib/x86_64-linux-gnu/libdl.so.2
0x00007fbe7deedf10  0x00007fbe7deee804  Yes         /lib/x86_64-linux-gnu/libutil.so.1
0x00007fbe7db43520  0x00007fbe7dc8be03  Yes         /lib/x86_64-linux-gnu/libc.so.6
0x00007fbe7ec20040  0x00007fbe7ec3728d  Yes (*)     /home/embray/src/gap-system/miniconda2/lib/./libtinfo.so.6
0x00007fbe7ea8bb00  0x00007fbe7eaa65b0  Yes         /lib64/ld-linux-x86-64.so.2
0x00007fbe7d78b8d0  0x00007fbe7d79444f  Yes (*)     /home/embray/src/gap-system/gap/pkg/io-4.5.1/bin/x86_64-pc-linux-gnu-default64-kv3/io.so
0x00007fbe7d565cd0  0x00007fbe7d57c62c  Yes (*)     /home/embray/src/gap-system/gap/pkg/nautytraces/bin/x86_64-pc-linux-gnu-default64-kv3/NautyTracesInterface.so
0x00007fbe7d359550  0x00007fbe7d35fcf1  Yes (*)     /home/embray/src/gap-system/gap/pkg/orb-4.8.2/bin/x86_64-pc-linux-gnu-default64-kv3/orb.so
0x00007fbe7d10a050  0x00007fbe7d13ea3d  Yes         /home/embray/src/gap-system/gap/pkg/digraphs/bin/x86_64-pc-linux-gnu-default64-kv3/digraphs.so
0x00007fbe7ebf0040  0x00007fbe7ec03a37  Yes         /home/embray/src/gap-system/miniconda2/lib/libplanarity.so.0
0x00007fbe7d033d90  0x00007fbe7d0ab211  Yes         /home/embray/src/gap-system/miniconda2/lib/libstdc++.so.6
0x00007fbe7ebd62d8  0x00007fbe7ebe1b79  Yes         /home/embray/src/gap-system/miniconda2/lib/libgcc_s.so.1
```

This is an (admittedly old) Ubuntu 14.04 machine; though clearly the problem is reproducible on Ubuntu 16.04 as well.

So a question I would have is, of the libraries coming from conda, which are most likely to impact GAP performance so severely (if indeed some library is the problem)?

james-d-mitchell:
What is libtinfo? That seems to be the only difference (in the library names at least).

embray:
libtinfo is for providing terminal information, and is used by readline.  Seems unlikely?

james-d-mitchell:
It is possible that one of the executables provided by conda is not compiled with optimisations enabled? Another possibility is that we are misusing the planarity code in Digraphs.

embray:
I tried running with `LD_PRELOAD="/lib/x86_64-linux-gnu/libgcc_s.so.1:/usr/lib/x86_64-linux-gnu/libstdc++.so.6:/lib/x86_64-linux-gnu/libz.so.1:/lib/x86_64-linux-gnu/libreadline.so.6:/home/embray/src/gap-system/gap-noconda.tmp/pkg/digraphs/extern/edge-addition-planarity-suite-Version_3.0.0.5/.libs/libplanarity.so.0`

the result being that the *only* library provided by conda being used is readline+tinfo.  But even then I'm running gap with `-E` (disable readline) and it doesn't seem to make a difference.

Indeed, I'm guessing compiler optimizations or something must be the next question...?

embray:
Ok problem partially solved, though I don't know why.  I re-configured my conda GAP with:

```
CCACHE_DISABLE=1 ./configure CFLAGS="$CFLAGS -O2"
```

not sure if disabling ccache had anything to do with it (probably not) but I included it for good measure.  

And that fixed it.

So for *some reason* it's building in conda with optimizations disabled.  Not sure why. I'm not using the compiler provided by conda nor do I have any environment variables set that should impact that.

embray:
It seems that (I don't know if this is a normal autoconf behavior or something in GAP's `configure`), but if I have some custom `CFLAGS` set in my environment, then it *won't* set the default of `CFLAGS="-g -O2"`.  Otherwise it sets that to the default.

ChrisJefferson:
It's normal autoconf behaviour to let you define your own CFLAGS if you want to.

james-d-mitchell:
Maybe the issue is that I'm setting `CFLAGS="$CFLAGS -I$HOME/miniconda2/include"` in `.travis.yml` and so everything compiled after that is not getting optimisation flags??

embray:
Yep; I apparently never realized this but by default if you define your own CFLAGS they are not *appended* to the default (which are `-g -O2`, but rather replace them).

embray:
@james-d-mitchell Just try changing it to `CFLAGS="$CFLAGS -O2 -g"`.  There are ways we can update the configure script for GAP to ensure that this gets done as well, but it would be a start.

Also, didn't know you were leaving so soon; sorry I didn't get to say bye!

james-d-mitchell:
Thanks @embray I've done that, let's see how it works out. 

I looked for you to say bye before I left, but I think you might have been sleeping, anyway, see you soon, and thanks for all your help!

wilfwilson:
It seems to take basically the expected time now. Happy for me to merge @james-d-mitchell?

james-d-mitchell:
Yup, sorry for the extreme number of updates required!

Issue #206, opened 2019-06-17
DIGRAPHS_Dir is unsafe


wilfwilson:
(This applies to the Semigroups package too.)

Currently, `DIGRAPHS_Dir` uses `PackageInfo("digraphs")[1]!.InstallationPath`. This gives you the installation path of *some* Digraphs installation, but not necessarily the one that is loaded. 

`PackageInfo("digraphs")` is a list, each entry of which is a record that corresponds to *some* Digraphs that GAP knows about. These will be listed in `PackageInfo("digraphs")` in some order, but that order does not need to correspond to the loaded one.

@ChrisJefferson managed to have two versions of the package installed, he loaded the second one, but then that meant that `DIGRAPHS_Dir` gave the installation path of the wrong version.

Instead we should use `GAPInfo.PackagesLoaded.digraphs` 🙂

wilfwilson:
Resolved by #209.

wilfwilson:
Resolved in v0.15.4.

PR #209, opened 2019-06-18
utils: fix DIGRAPHS_Dir

james-d-mitchell:
Resolves Issue #206.

### DIFF
--- a/gap/utils.gi
+++ b/gap/utils.gi
@@ -118,7 +118,7 @@ end);
 
 InstallGlobalFunction(DIGRAPHS_Dir,
 function()
-  return PackageInfo("digraphs")[1]!.InstallationPath;
+  return GAPInfo.PackagesLoaded.digraphs[1];
 end);
 
 #############################################################################


wilfwilson:
Could you restart the AppVeyor test? This PR surely can't have affected it, so it's probably just a random failure.

james-d-mitchell:
I'm not so sure, did you look at the appveyor log?

james-d-mitchell:
I restarted in any case...

wilfwilson:
I looked at the log, but couldn't immediately make sense of it. In any case, if it recurs we can look more closely :)

james-d-mitchell:
Failed again

wilfwilson:
Yep, really annoying. It's only when it gets to the extreme tests, after a lot of reading (using `DIGRAPHS_Dir` has already been done.

wilfwilson:
I think that the AppVeyor failure is unrelated to this PR. It's failing here:
```
########> Diff in /home/appveyor/gap/pkg/digraphs/tst/extreme/attr.tst:31
# Input is:
gr := ReadDigraphs(Concatenation(DIGRAPHS_Dir(),
                                    "/data/symmetric-closure.ds6.gz"),
                      DigraphFromDiSparse6String,
                      1);
# Expected output:
<digraph with 46656 vertices, 120245 edges>
# But found:
Error, no method found! For debugging hints type ?Recovery from NoMethodFound
Error, no 1st choice method found for `DigraphFromDiSparse6String' on 1 argume\
nts
The 1st argument is 'fail' which might point to an earlier problem

########

```
But a few lines before it is successfully doing:
```
gap> gr := ReadDigraphs(Concatenation(DIGRAPHS_Dir(),
>                                     "/digraphs-lib/extreme.d6.gz"), 1);
```
And (surely) there's a problem with `DIGRAPHS_Dir()` in the first bit if and only if there's a problem with it in the second bit.

So I think this must be one of those annoying GAP-Windows problems which we battle with (GAP on Windows is weird with forking, which happens when you read a file I think).

I think poking it once a twice more will hopefully get AppVeyor to pass (I'm running it https://ci.appveyor.com/project/wilfwilson/digraphs/builds/25364607 at the moment) - if it does I propose merging.

wilfwilson:
I'm confident that this will pass now.

james-d-mitchell:
Thanks @wilfwilson

PR #211, opened 2019-06-18
Improve DigraphOddGirth

MTWhyte:
This pull request makes a minor tweak to the function `DigraphOddGirth`, avoiding unnecessary computation.

### DIFF
--- a/tst/standard/attr.tst
+++ b/tst/standard/attr.tst
@@ -1183,7 +1183,7 @@ gap> for i in [2 .. 200] do
 gap> D := CycleDigraph(IsMutableDigraph, 7);
 <mutable digraph with 7 vertices, 7 edges>
 gap> for i in [1 .. 20] do
->   D := DigraphDisjointUnion(D, G);
+>   DigraphDisjointUnion(D, G);
 > od;
 gap> DigraphOddGirth(D);
 7
--- a/tst/standard/attr.tst
+++ b/tst/standard/attr.tst
@@ -1174,14 +1174,14 @@ gap> for i in [1 .. 50] do
 > od;
 gap> DigraphOddGirth(gr);
 3
-gap> G := Digraph([[]]);
-<immutable digraph with 1 vertex, 0 edges>
+gap> G := Digraph(IsMutableDigraph, [[]]);
+<mutable digraph with 1 vertex, 0 edges>
 gap> for i in [2 .. 200] do
->   G := DigraphAddVertex(G, i);
->   G := DigraphAddEdges(G, [[1, i], [i, 1]]);
+>   DigraphAddVertex(G, i);
+>   DigraphAddEdges(G, [[1, i], [i, 1]]);
 > od;
-gap> D := CycleDigraph(7);
-<immutable digraph with 7 vertices, 7 edges>
+gap> D := CycleDigraph(IsMutableDigraph, 7);
+<mutable digraph with 7 vertices, 7 edges>
 gap> for i in [1 .. 20] do
 >   D := DigraphDisjointUnion(D, G);
 > od;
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
Great, thanks @MTWhyte, can you add a test that is faster with the changes in this PR, than before?

MTWhyte:
Have thought of an example for the tests which takes half a second with the changes, and about 25 seconds without. Will push in due course.

tst/standard/attr.tst
```diff
@@ -1174,6 +1174,19 @@ gap> for i in [1 .. 50] do
...
+gap> for i in [2 .. 200] do
+>   G := DigraphAddVertex(G, i);
+>   G := DigraphAddEdges(G, [[1, i], [i, 1]]);
+> od;
+gap> D := CycleDigraph(7);
+<immutable digraph with 7 vertices, 7 edges>
+gap> for i in [1 .. 20] do
+>   D := DigraphDisjointUnion(D, G);
+> od;
+gap> DigraphOddGirth(D);
```
wilfwilson: Could you please do all of this test with mutable digraphs instead? It will remove a lot of copying I think and therefore quicker (and also using our cool new functionality 🙂): `G := Digraph(IsMutable[[]]);`, `D := DigraphMutableCopy(CycleDigraph(7));`, etc.

tst/standard/attr.tst
```diff
@@ -1174,14 +1174,14 @@ gap> for i in [1 .. 50] do
...
->   G := DigraphAddEdges(G, [[1, i], [i, 1]]);
+>   DigraphAddVertex(G, i);
+>   DigraphAddEdges(G, [[1, i], [i, 1]]);
 > od;
-gap> D := CycleDigraph(7);
-<immutable digraph with 7 vertices, 7 edges>
+gap> D := CycleDigraph(IsMutableDigraph, 7);
+<mutable digraph with 7 vertices, 7 edges>
 gap> for i in [1 .. 20] do
 >   D := DigraphDisjointUnion(D, G);
```
MTWhyte: This slipped by me because I wasn't sure how `DigraphDisjointUnion` works for mutable digraphs, but it seems like it alters the digraph in the first argument, and leaves the digraph in the second argument alone - I'll update this line accordingly. Would it be useful if I add a mutable example to the `DigraphDisjointUnion` documentation?

MTWhyte:
@wilfwilson that's interesting. I think the bug is because `InducedSubdigraph` modifies the digraph if it is mutable, so I can fix it pretty easily.

In looking in to this, I've actually found a possible bug in `InducedSubdigraph`, in that it modifies its argument digraph even if it returns an error. For example, in the following, I accidentally omitted the list brackets in the second argument:

~~~
gap> D := RandomDigraph(IsMutableDigraph, 10, 0.5);
<mutable digraph with 10 vertices, 57 edges>
gap> InducedSubdigraph(D, [4, 5]);
Error, List Elements: <list>[4] must have an assigned value in
  DigraphVertexLabels( D
 ) at /Users/murraywhyte/gap/pkg/Digraphs/gap/oper.gi:827 called from
<function "unknown">( <arguments> )
 called from read-eval loop at *stdin*:16
brk> quit;
gap> D;
<mutable digraph with 2 vertices, 4 edges>
~~~

MTWhyte:
I've fixed the bug, albeit in a slightly less nice way than I'd planned. Originally, I was going to change the structure to use `DigraphRemoveVertices` on a mutable digraph, in place of `InducedSubdigraph`, but this didn't make some examples run any faster, and in fact stopped the most recently added test from running in any reasonable time. I'll push my changes now.

wilfwilson:
I've pushed an update - I noticed that you only need to make an immutable copy if you've already passed the first few checks and the digraph is not strongly connected. If the tests pass I'll merge.

wilfwilson:
Thanks @MTWhyte.

PR #214, opened 2019-06-21
Fix issue 213

james-d-mitchell:
I think this will fix Issue 213.


james-d-mitchell:
Nope, messed up the tests, should be fixed now.

PR #217, opened 2019-06-21
Add IsDigraphCore property

MTWhyte:
This pull request adds a method for determining whether or not a digraph is a core. A digraph _D_ is a core if and only if every endomorphism on _D_ is an automorphism on _D_.

### DIFF
--- a/gap/prop.gi
+++ b/gap/prop.gi
@@ -596,15 +596,16 @@ function(D)
   local hook, proper_endo_found, N;
 
   N := DigraphNrVertices(D);
-  if DigraphHasLoops(D) and N > 1 then
+  if (DigraphHasLoops(D) or IsEmptyDigraph(D)) and N > 1 then
     return false;
   elif IsCompleteDigraph(D) then
     return true;
-  elif IsBipartiteDigraph(D) and N > 2 then
+  elif IsSymmetricDigraph(D) and IsBipartiteDigraph(D) and N > 2 then
     return false;
   fi;
-  # The core of a digraph with loops is a vertex with a loop, and the core of a
-  # bipartite digraph is the complete digraph on 2 vertices.
+  # The core of a digraph with loops is a vertex with a loop, of an empty
+  # digraph is a vertex, and of a non-empty, symmetric bipartite digraph is the
+  # complete digraph on 2 vertices.
 
   proper_endo_found := false;
   hook := function(unneded_argument, T)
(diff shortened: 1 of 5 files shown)


gap/prop.gi
```diff
@@ -589,3 +589,48 @@ end);
...
+                             hook,                      # hook
+                             fail,                      # user_param
+                             infinity,                  # max results
+                             fail,                      # hint
+                             false,                     # injective
+                             DigraphVertices(D),        # image
+                             [],                        # partial_map
+                             fail,                      # colors1
+                             fail);                     # colors2
+  if not proper_endo_found then
```
wilfwilson: These five lines can be replaced with
`return not proper_endo_found;`.

gap/prop.gi
```diff
@@ -589,3 +589,48 @@ end);
...
+  N := DigraphNrVertices(D);
+  if DigraphHasLoops(D) and N > 1 then
+    return false;
+  elif IsCompleteDigraph(D) then
+    return true;
+  elif IsBipartiteDigraph(D) and N > 2 then
+    return false;
+  fi;
+  # The core of a digraph with loops is a vertex with a loop, and the core of a
+  # bipartite digraph is the complete digraph on 2 vertices.
```
wilfwilson: A non-empty bipartite digraph. If there’s no edges the core is the empty digraph with one vertex.

MTWhyte:
Thanks for your suggestions - I've implemented them now. Before I push again, I'm thinking about the empty digraph with no vertices. We have

~~~
gap> D := EmptyDigraph(0);
<immutable digraph with 0 vertices, 0 edges>
gap> DigraphHomomorphism(D, D);
fail
~~~

and so by the above `D` here shouldn't be a core. On the other hand, if `D` isn't a core, then what would its core be, given that it has no proper subdigraphs? What do you think?

wilfwilson:
Good idea about the digraph with no vertices.

I think that the digraph with 0 vertices is a core. If you think of a function on a set as a special kind of binary relation on that set, then the unique function on the digraph with 0 vertices is simply the empty relation, i.e. the empty function, which vacuously is an automorphism. So everything is consistent :)

I think it's fair enough that `DigraphHomomorphism(D, D);` fails here, for technical reasons rather than mathematical reasons, because it's trick to implement an 'empty function' in GAP.

MTWhyte:
Great, I'll push my changes now

PR #218, opened 2019-06-26
Ignore: Testing range-source digraph bug

flsmith:
This PR only exists to get Travis to run some tests.

I'm examining strange behaviour seen in https://travis-ci.org/gap-packages/Digraphs/jobs/550844430


wilfwilson:
You should turn on Travis on your own fork, i.e. make Travis run the tests when you push to your own forks of things, not just `gap-packages`.

You can turn stuff on here https://travis-ci.org/account/repositories. Here for instance are all of the jobs that Travis runs for me on my own fork of Digraphs: https://travis-ci.org/wilfwilson/Digraphs/builds

Means I don't have to wait for other `gap-packages` jobs to be finished before the jobs start running on my stuff (and there can be a lot of waiting).

flsmith:
Ah, I didn't realise it was so easy.

Thanks!

PR #220, opened 2019-06-26
Display

james-d-mitchell:
This PR improves the error messages in `Splash` if the specified, or none of the default, viewers is available. It also stops using the architecture to decide about the list of `VizViewers`.

### DIFF
--- a/tst/standard/display.tst
+++ b/tst/standard/display.tst
@@ -205,7 +205,8 @@ node [shape=Mrecord, height=0.5, fixedsize=true]ranksep=1;
 gap> Splash(DotDigraph(RandomDigraph(10)), rec(viewer := 1));
 Error, the option `viewer` must be a string, not an integer,
 gap> Splash(DotDigraph(RandomDigraph(10)), rec(viewer := "asdfasfa"));
-Error, the viewer "asdfasfa" specified in the option `viewer` is not available,
+Error, the viewer "asdfasfa" specified in the option `viewer` is not available\
+,
 
 #  DIGRAPHS_UnbindVariables
 gap> Unbind(adj);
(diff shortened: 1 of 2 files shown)


tst/standard/display.tst
```diff
@@ -201,6 +201,12 @@ node [shape=Mrecord, height=0.5, fixedsize=true]ranksep=1;
 2 -> 1
 }
 
+# Splash
+gap> Splash(DotDigraph(RandomDigraph(10)), rec(viewer := 1));
+Error, the option `viewer` must be a string, not an integer,
+gap> Splash(DotDigraph(RandomDigraph(10)), rec(viewer := "shit"));
```
wilfwilson: interesting

ChrisJefferson: Also, there is a fairly popular program called 'fuck' ( https://github.com/nvbn/thefuck ) which I believe installs a command called 'shit', so some people will have that program installed (if you are going to look for a runnable program with that name).

james-d-mitchell: So, should I use a string that can't be misconstrued as a program?? Or is this good to go? @wilfwilson

wilfwilson: I'd prefer that (don't mind if you still want to have 'shit' as part of it), but I don't feel strongly about it.

wilfwilson: @james-d-mitchell Okay fine, but maybe write "shite" instead to avoid the conflict potential that Chris mentioned above?
```suggestion
gap> Splash(DotDigraph(RandomDigraph(10)), rec(viewer := "shite"));
```

james-d-mitchell:
@wilfwilson sorry I forgot about this completely. I changed away from using an actual word. Hopefully this is ok!

wilfwilson:
Thanks. The test now has a diff because of the line break, just do the `rec(rewriteToFile := true)` thing and it should be fine.

wilfwilson:
I just made the change, to get it over with.

PR #221, opened 2019-07-14
Add DigraphCore method

MTWhyte:
This pull request adds a method `DigraphCore`, which takes a digraph and returns its core.

The core of a digraph _D_ is the (unique up to isomorphism) minimal subdigraph which is a homomorphic image of _D_.

The method works by searching for endomorphisms of the argument digraph, alternating between looking for ones of the smallest and largest possible rank. If one of the smallest possible rank is found, then the procedure stops. If one of the largest possible rank is found, then unnecessary parts of the digraph are removed from consideration.

In the case where the digraph is non-connected, the method computes the core of each connected component and then combines them in the minimal possible way, rather than searching for the core of the digraph as a whole. This works significantly faster on the disjoint union of cycle digraphs in the tests.

### DIFF
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -1467,12 +1467,12 @@ gap> HamiltonianPath(g);
     <Returns>A list of positive integers.</Returns>
     <Description>
       If <A>D</A> is a digraph, then <C>DigraphCore</C> returns a list of vertices
-      corresponding to the <C>core</C> of <A>D</A>. In particular, the subdigraph of <A>D</A>
-      induced by this list is isomorphic to the core of <A>D</A>.<P/>
+      corresponding to the <C>core</C> of <A>D</A>. In particular, the subdigraph
+      of <A>D</A> induced by this list is isomorphic to the core of <A>D</A>.<P/>
 
-    The <E>core</E> of a digraph <C>D</C> is the minimal subdigraph <E>C</E> of <C>D</C> which
-    is a homomorphic image of <C>D</C>. The core of a digraph <C>D</C> is unique up to
-    isomorphism.
+    The <E>core</E> of a digraph <C>D</C> is the minimal subdigraph <A>C</A> of
+    <C>D</C> which is a homomorphic image of <C>D</C>. The core of a digraph 
+    is unique up to isomorphism.
       <Example><![CDATA[
 gap> D := DigraphSymmetricClosure(CycleDigraph(8));
 <immutable digraph with 8 vertices, 16 edges>
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -1461,6 +1461,35 @@ gap> HamiltonianPath(g);
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="DigraphCore">
+  <ManSection>
+    <Oper Name = "DigraphCore" Arg = "D"/>
+    <Returns>A list of positive integers.</Returns>
+    <Description>
+      If <A>D</A> is a digraph, then <C>DigraphCore</C> returns a list of vertices
+      corresponding to the <C>core</C> of <A>D</A>. In particular, the subdigraph of <A>D</A>
+      induced by this list is isomorphic to the core of <A>D</A>.<P/>
+
+    The <E>core</E> of a digraph <C>D</C> is the minimal subdigraph <E>C</E> of <C>D</C> which
+    is a homomorphic image of <C>D</C>. The core of a digraph <C>D</C> is unique up to
+    isomorphism.
+      <Example><![CDATA[
+gap> D := DigraphSymmetricClosure(CycleDigraph(8));
+<immutable digraph with 8 vertices, 16 edges>
+gap> DigraphCore(D);
+[ 1, 2 ]
+gap> D := PetersenGraph();
+<immutable digraph with 10 vertices, 30 edges>
+gap> DigraphCore(D);
+[ 1 .. 10 ]
+gap> D := Digraph([[3], [3], [4], [5], [2]]);
+<immutable digraph with 5 vertices, 5 edges>
+gap> DigraphCore(D);
+[ 2, 3, 4, 5 ]]]></Example>
+    </Description>
+  </ManSection>
+<#/GAPDoc>
+
 <#GAPDoc Label="MaximalAntiSymmetricSubdigraph">
 <ManSection>
   <Attr Name="MaximalAntiSymmetricSubdigraph" Arg="digraph"/>
--- a/gap/attr.gi
+++ b/gap/attr.gi
@@ -1687,6 +1687,138 @@ function(D)
   return fail;
 end);
 
+InstallMethod(DigraphCore, "for a digraph",
+[IsDigraph],
+function(digraph)
+  local N, lo, topdown, bottomup, hom, lo_var, image,
+  comps, comp, cores, G, vert_labels, D, in_core, n, m, L;
+  if not IsImmutableDigraph(digraph) then
+    digraph := DigraphImmutableCopy(digraph);
+  fi;
+  N := DigraphNrVertices(digraph);
+  if IsEmptyDigraph(digraph) then
+    if N >= 1 then
+      return [1];
+    else
+      return [];
+    fi;
+  fi;
+  digraph := ReducedDigraph(digraph);  # isolated verts are not in core
+  if DigraphHasLoops(digraph) then
+    return DigraphLoops(digraph){[1]};
+  elif not IsConnectedDigraph(digraph) then
+    comps := DigraphConnectedComponents(digraph).comps;
+    DigraphVertexLabels(digraph);
+    cores := [];
+    for comp in comps do
+      D := InducedSubdigraph(digraph, comp);
+      D := InducedSubdigraph(D,
+           Positions(BlistList(DigraphVertexLabels(D), DigraphCore(D)), true));
+      Add(cores, D);
+    od;
+    L := Length(cores);
+    in_core := List([1 .. L], x -> true);
+    n := 1;
+    while n <= L do
+      if n <> L then
+        m := n + 1;
+      else
+        m := 1;
+      fi;
+      while m <> n and in_core[n] do
+        if in_core[m] and DigraphHomomorphism(cores[m], cores[n]) <> fail then
+          in_core[m] := false;
+        fi;
+        if m < L then
+          m := m + 1;
+        else
+          m := 1;
+        fi;
+      od;
+      n := n + 1;
+    od;
+    cores := ListBlist(cores, in_core);
+    vert_labels := [];
+    for G in cores do
+      vert_labels := Concatenation(vert_labels, DigraphVertexLabels(G));
+    od;
+    return vert_labels;
+  elif IsCompleteDigraph(digraph) then
+    return [1 .. N];
+  elif IsSymmetricDigraph(digraph) and IsBipartiteDigraph(digraph) then
+    return DigraphEdges(digraph)[1];
+  fi;
+
+  if IsDigraphCore(digraph) then
+    return DigraphVertexLabels(digraph);
+  fi;
+
+  lo := function(D)  # lower bound on core size
+    local oddgirth;
+    oddgirth := DigraphOddGirth(D);
+    if oddgirth <> infinity then
+      return oddgirth;
+    fi;
+    return 3;
+  end;
+
+  hom      := [];
+  lo_var := lo(digraph);
+  bottomup := lo_var;
+  N := DigraphNrVertices(digraph);
+  topdown  := N;
+
+  while topdown >= bottomup do
+    HomomorphismDigraphsFinder(digraph,                   # domain copy
+                               digraph,                   # range copy
+                               fail,                      # hook
+                               hom,                       # user_param
+                               1,                         # max_results
+                               bottomup,                  # hint (i.e. rank)
+                               false,                     # injective
+                               DigraphVertices(digraph),  # image
+                               [],                        # partial_map
+                               fail,                      # colors1
+                               fail);                     # colors2
+
+    if Length(hom) = 1 then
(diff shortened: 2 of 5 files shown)


doc/attr.xml
```diff
@@ -1461,6 +1461,35 @@ gap> HamiltonianPath(g);
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="DigraphCore">
+  <ManSection>
+    <Oper Name = "DigraphCore" Arg = "D"/>
+    <Returns>A list of positive integers.</Returns>
+    <Description>
+      If <A>D</A> is a digraph, then <C>DigraphCore</C> returns a list of vertices
```
wilfwilson: What you've implemented is the list of vertex labels corresponding to the core. I think you actually want to be returning the list of vertices, which is what the documentation suggestions. Vertex labels have no requirements on them, they're not even required to be distinct, which makes them not very useful for being things like the output of computations. For instance, look at this:
```
gap> gr := CycleDigraph(4);;
gap> SetDigraphVertexLabels(gr, [fail, fail, fail, fail]);
gap> DigraphCore(gr);
[ fail, fail, fail, fail ]
```
Not very helpful! There might be a good reason to use non-unique vertex labels, but if you do so this function isn't very useful.

gap/attr.gi
```diff
@@ -1687,6 +1687,138 @@ function(D)
...
+function(digraph)
+  local N, lo, topdown, bottomup, hom, lo_var, image,
+  comps, comp, cores, G, vert_labels, D, in_core, n, m, L;
+  if not IsImmutableDigraph(digraph) then
+    digraph := DigraphImmutableCopy(digraph);
+  fi;
+  N := DigraphNrVertices(digraph);
+  if IsEmptyDigraph(digraph) then
+    if N >= 1 then
+      return [1];
```
wilfwilson: And here it is returning vertex numbers rather than vertex labels (as it should) - just to point out the inconsistency.

gap/attr.gi
```diff
@@ -1687,6 +1687,138 @@ function(D)
...
+    od;
+    return vert_labels;
+  elif IsCompleteDigraph(digraph) then
+    return [1 .. N];
+  elif IsSymmetricDigraph(digraph) and IsBipartiteDigraph(digraph) then
+    return DigraphEdges(digraph)[1];
+  fi;
+
+  if IsDigraphCore(digraph) then
+    return DigraphVertexLabels(digraph);
```
wilfwilson:
```suggestion
    return DigraphVertices(digraph);
```

gap/attr.gi
```diff
@@ -1687,6 +1687,138 @@ function(D)
...
+                               fail);                     # colors2
+    if Length(hom) = 1 then
+      image    := ImageSetOfTransformation(hom[1], N);
+      digraph  := InducedSubdigraph(digraph, image);
+      N        := DigraphNrVertices(digraph);
+      Unbind(hom[1]);
+      lo_var := lo(digraph);
+    fi;
+
+    if topdown - 1 < N then
```
wilfwilson: Here (I think?) you can just do:
```
topdown := Minimum(topdown - 1, N);
bottom := Maximum(bottomup + 1, lo_var);
```

gap/attr.gi
```diff
@@ -1687,6 +1687,138 @@ function(D)
...
+                               1,                         # max_results
+                               bottomup,                  # hint (i.e. rank)
+                               false,                     # injective
+                               DigraphVertices(digraph),  # image
+                               [],                        # partial_map
+                               fail,                      # colors1
+                               fail);                     # colors2
+
+    if Length(hom) = 1 then
+      image    := ImageSetOfTransformation(hom[1], N);
```
wilfwilson:
```suggestion
      return ImageSetOfTransformation(hom[1], N);
```

gap/attr.gi
```diff
@@ -1687,6 +1687,138 @@ function(D)
...
+  lo := function(D)  # lower bound on core size
+    local oddgirth;
+    oddgirth := DigraphOddGirth(D);
+    if oddgirth <> infinity then
+      return oddgirth;
+    fi;
+    return 3;
+  end;
+
+  hom      := [];
```
wilfwilson: These assignments in total are neither in their usual positions (lots of spaces between variable names `hom` and `topdown` and the subsequent `:=`), nor are they all lined up nicely. I don't mind which one you choose, but I'd like it if you chose one of those two strategies rather than just random.

gap/attr.gi
```diff
@@ -1687,6 +1687,138 @@ function(D)
...
+    od;
+    cores := ListBlist(cores, in_core);
+    vert_labels := [];
+    for G in cores do
+      vert_labels := Concatenation(vert_labels, DigraphVertexLabels(G));
+    od;
+    return vert_labels;
+  elif IsCompleteDigraph(digraph) then
+    return [1 .. N];
+  elif IsSymmetricDigraph(digraph) and IsBipartiteDigraph(digraph) then
```
wilfwilson: I think this part of the `if` should come before the `elif not IsConnectedDigraph(digraph) then` bit, because if you have a non-connected symmetric bipartite digraph, which is easy to check, then this method would be super quick, whereas the non-connected bit of code for constructing the core could conceivably take a bit longer.

gap/attr.gi
```diff
@@ -1687,6 +1687,138 @@ function(D)
...
+    return DigraphEdges(digraph)[1];
```
wilfwilson: This is a potentially very slow way to get a single edge of a digraph - we always want to avoid constructing the list/set of edges if possible, it can be slow to produce and take up a lot of memory.

It would take more code, but you could do something like:
```
i := First(DigraphVertices(digraph), i -> OutDegreeOfVertex(digraph, i) > 0);
return [i, OutNeighboursOfVertex(i, 1)];
```
instead.

gap/attr.gi
```diff
@@ -1687,6 +1687,138 @@ function(D)
...
+                               bottomup,                  # hint (i.e. rank)
+                               false,                     # injective
+                               DigraphVertices(digraph),  # image
+                               [],                        # partial_map
+                               fail,                      # colors1
+                               fail);                     # colors2
+
+    if Length(hom) = 1 then
+      image    := ImageSetOfTransformation(hom[1], N);
+      return DigraphVertexLabels(digraph){image};
```
wilfwilson:
```suggestion
```

gap/attr.gi
```diff
@@ -1687,6 +1687,138 @@ function(D)
...
+      n := n + 1;
+    od;
+    cores := ListBlist(cores, in_core);
+    vert_labels := [];
+    for G in cores do
+      vert_labels := Concatenation(vert_labels, DigraphVertexLabels(G));
+    od;
+    return vert_labels;
+  elif IsCompleteDigraph(digraph) then
+    return [1 .. N];
```
wilfwilson:
```suggestion
    return DigraphVertices(digraph);
```

gap/attr.gi
```diff
@@ -1687,6 +1687,138 @@ function(D)
...
+  if DigraphHasLoops(digraph) then
+    return DigraphLoops(digraph){[1]};
+  elif not IsConnectedDigraph(digraph) then
+    comps := DigraphConnectedComponents(digraph).comps;
+    DigraphVertexLabels(digraph);
+    cores := [];
+    for comp in comps do
+      D := InducedSubdigraph(digraph, comp);
+      D := InducedSubdigraph(D,
+           Positions(BlistList(DigraphVertexLabels(D), DigraphCore(D)), true));
```
wilfwilson: I don't understand what this line is supposed to do.

gap/attr.gi
```diff
@@ -1687,6 +1687,138 @@ function(D)
...
+    else
+      return [];
+    fi;
+  fi;
+  digraph := ReducedDigraph(digraph);  # isolated verts are not in core
+  if DigraphHasLoops(digraph) then
+    return DigraphLoops(digraph){[1]};
+  elif not IsConnectedDigraph(digraph) then
+    comps := DigraphConnectedComponents(digraph).comps;
+    DigraphVertexLabels(digraph);
```
wilfwilson: If the vertex labels are already set, which they might be, this doesn't do anything. Do you perhaps mean to do:
```suggestion
    SetDigraphVertexLabels(digraph, [1 .. N]);
```

gap/attr.gi
```diff
@@ -1687,6 +1687,138 @@ function(D)
...
+    cores := [];
+    for comp in comps do
+      D := InducedSubdigraph(digraph, comp);
+      D := InducedSubdigraph(D,
+           Positions(BlistList(DigraphVertexLabels(D), DigraphCore(D)), true));
+      Add(cores, D);
+    od;
+    L := Length(cores);
+    in_core := List([1 .. L], x -> true);
```
wilfwilson:
```suggestion
    in_core := ListWithIdenticalEntries(L, true);
```

gap/attr.gi
```diff
@@ -1687,6 +1687,138 @@ function(D)
...
+        if m < L then
+          m := m + 1;
+        else
+          m := 1;
+        fi;
+      od;
+      n := n + 1;
+    od;
+    cores := ListBlist(cores, in_core);
+    vert_labels := [];
```
wilfwilson:
```suggestion
    return Union(List(cores, DigraphVertexLabels));
```

gap/attr.gi
MTWhyte: Thanks for your comments - I'd just been using `DigraphVertexLabels` as a way of retrieving prior vertex numbers, without really understanding how the vertices of a digraph can be labelled in whatever way we like.

This line creates the induced subdigraph of connected component `D` corresponding to its core - the reason it's not just `D := InducedSubdigraph(D, DigraphCore(D))` is to do with the vertex labelling. Since I've mistakenly been thinking of vertex labels and the list of vertices interchangeably, when I deal with the connected components here, `DigraphCore(D)` isn't necessarily a subset of the list of vertices of `D`, and so instead this line avoids running in to this problem by finding the positions of the elements of `DigraphCore(D)` within `DigraphVertexLabels(D)`. However, this will change after I've sorted out the vertex label misunderstanding in the rest of the function, and the line will probably end up much simpler.

james-d-mitchell: This'd be better if it was:
~~~~
D := InducedSubdigraph(D, DigraphVertexLabels(D){DigraphCore(D)};
~~~~
if that's really what you want to do.

gap/attr.gi
MTWhyte: I think it might be better to run `SetDigraphVertexLabels(digraph, DigraphVertices(digraph))` right at the beginning of the function, and then keep the `return DigraphVertexLabels(digraph)` parts. Otherwise, since I run `ReducedDigraph(digraph)` right at the beginning of the function, the way the vertices with no in or out neighbours play in to the vertex labelling are forgotten about. As in,
~~~
gap> D := Digraph([[], [3], [2]]);
<immutable digraph with 3 vertices, 2 edges>
gap> DigraphCore(D);
[ 1, 2 ]
~~~
when I really would want `[2, 3]` returned.

To fix my inconsistency with talking about vertex labelling and numbering, I'll remove the cases where I returned vertices with returning the vertex label. Sorry this has taken me a while, by the way - I have been away for a bit.

gap/attr.gi
MTWhyte: I agree that using the minimum and maximum functions is more readable, but I remember @james-d-mitchell suggested that sometimes they can be less efficient than using conditionals?

wilfwilson: Possibly, but I think any difference in speed would be negligible, whereas I think the `Minimum` etc significantly benefits readability and future maintenance.

MTWhyte:
I've fixed the vertex labels vs vertices problem, and I'll push my changes now. At the beginning, of my function, I think it's necessary to make an immutable copy of the argument `digraph`, regardless of whether or not `digraph` is mutable, otherwise changes to the vertex labels of `digraph` aren't just local to the function. Is there a better way around this I don't know about?

gap/attr.gi
```diff
@@ -1687,6 +1687,127 @@ function(D)
...
+    else
+      return [];
+    fi;
+  fi;
+  SetDigraphVertexLabels(digraph, [1 .. N]);
+  digraph := ReducedDigraph(digraph);  # isolated verts are not in core
+  N       := DigraphNrVertices(digraph);
+  if DigraphHasLoops(digraph) then
+    i := First(DigraphVertices(digraph), i -> i in OutNeighbours(digraph)[i]);
+    return [DigraphVertexLabels(digraph)[i]];
```
wilfwilson:
```suggestion
    return [DigraphVertexLabel(digraph, i)];
```
Same kind of philosophy here. It also fits in nicely with what you've done below, with `OutDegreeOfVertex` and `OutNeighboursOfVertex` 🙂

gap/attr.gi
```diff
@@ -1687,6 +1687,127 @@ function(D)
...
+      return [1];
+    else
+      return [];
+    fi;
+  fi;
+  SetDigraphVertexLabels(digraph, [1 .. N]);
+  digraph := ReducedDigraph(digraph);  # isolated verts are not in core
+  N       := DigraphNrVertices(digraph);
+  if DigraphHasLoops(digraph) then
+    i := First(DigraphVertices(digraph), i -> i in OutNeighbours(digraph)[i]);
```
wilfwilson:
```suggestion
    i := First(DigraphVertices(digraph), i -> i in OutNeighboursOfVertex(digraph, i));
```
I suggest this change for maximum abstraction with little overhead. For a possible future representation of digraph, you might be able to get the out-neighbours of a vertex without having to construct and hold in memory the whole list of `OutNeighbours`, which is what `OutNeighbours(digraph)` does.

MTWhyte: Great, I'll push the changes now 🙂

wilfwilson:
Thanks for this. Let's see whether @james-d-mitchell has any further suggestions, and merge if not.

Issue #222, opened 2019-07-14
MonomorphismsDigraphs returns non-monomorphisms


wilfwilson:
The following was reported by Gordon Royle to James Mitchell via email:
```
gap> D1 := DigraphFromGraph6String("E}hO");
<digraph with 6 vertices, 18 edges>
gap> D2 := DigraphFromGraph6String("E}h_");
<digraph with 6 vertices, 18 edges>
gap> mono := MonomorphismsDigraphs(D1, D2);
[ Transformation( [ 1, 2, 3, 3 ] ), Transformation( [ 1, 3, 2, 2, 4 ] ),
  Transformation( [ 2, 1, 3, 3, 6, 5 ] ),
  Transformation( [ 2, 3, 1, 1, 4, 5 ] ),
  Transformation( [ 3, 1, 2, 2, 6, 4 ] ),
  Transformation( [ 3, 2, 1, 1, 5, 4 ] ) ]
gap> ForAll(mono, f -> IsDigraphHomomorphism(D1, D2, f));
true
gap> ForAny(mono, f -> IsDigraphMonomorphism(D1, D2, f));
false
```
`MonomorphismsDigraphs` is returning transformations that, although homomorphisms, do not define monomorphisms. (They're clearly not injective!).

james-d-mitchell:
Thanks @wilfwilson I believe I have a fix for this, I’ll make Pr tomorrow

wilfwilson:
Great news.

wilfwilson:
Resolved in v0.15.4.

PR #223, opened 2019-07-15
Fix issue 222

james-d-mitchell:
This PR fixes issue 222. Previously the monomorphism finding code did not properly ensure that the found homomorphisms were injective. This PR fixes this by removing all points in the image of a partial monomorphism from the set of vertices that can be chosen for the image of the next point.

### DIFF
--- a/tst/standard/grahom.tst
+++ b/tst/standard/grahom.tst
@@ -2101,6 +2101,14 @@ gap> HomomorphismDigraphsFinder(D1, D2, fail, [], 1,
 > fail, 2, [1, 2, 3], fail, fail, fail);
 [  ]
 
+# Issue 222
+gap> D1 := DigraphFromGraph6String("E}hO");
+<digraph with 6 vertices, 18 edges>
+gap> D2 := DigraphFromGraph6String("E}h_");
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
Resolved Issue #222

tst/standard/grahom.tst
```diff
@@ -2101,6 +2101,18 @@ gap> HomomorphismDigraphsFinder(D1, D2, fail, [], 1,
...
 [  ]
 
+# Issue 222
+gap> D1 := DigraphFromGraph6String("E}hO");
+<digraph with 6 vertices, 18 edges>
+gap> D2 := DigraphFromGraph6String("E}h_");
+<digraph with 6 vertices, 18 edges>
+gap> mono := MonomorphismsDigraphs(D1, D2);
+[  ]
+gap> ForAll(mono, f -> IsDigraphHomomorphism(D1, D2, f));
```
wilfwilson: I recommend you get rid of these two `ForAll` calls, they're a bit redundant.

james-d-mitchell: Done

Issue #226, opened 2019-07-25
Implement digraph products and related operations


reiniscirpons:
There are many notions of digraph product. Furthermore, there is a notion of primality and unique factorization wrt these products, and there are efficient algorithms for computing some of them (see article at the end for a full exposition).

I think it would be nice to implement:
- `DigraphCartesianProduct(G1, G2, ...)` - for digraphs compute the Cartesian product of all of them.
- `DigraphDirectProduct(G1, G2, ...)` - for digraphs compute the Direct product of all of them.
- `IsDigraphCartesianProductPrime(G)` - for a digraph decide whether its prime wrt the Cartesian product.
- `IsDigraphDirectProductPrime(G)` - for a digraph decide whether its prime wrt the Direct product.
- `DigraphCartesianSkeleton(G)` - for a digraph, compute its Cartesian skeleton (this is useful in direct product factorization)
- `DigraphCartesianPrimeFactorization(G)` - for a digraph, compute its prime factors wrt the Cartesian product.
- `DigraphDirectPrimeFactorization(G)` - for a digraph, compute its prime factors wrt the Direct product.

-----
Some theory:
For digraphs `G` and `H` there are multiple possible ways of defining a product digraph. Two of the more important ones are:

1.  Cartesian product `G ☐ H`. The set of vertices is `V(G) x V(H)`, there is an edge between `(x, x')` and `(y, y')` iff either `(x, y) ∈ E(G)` and `x' = y'` or `x = y` and `(x',y') ∈ E(H)`. 

2. Direct product `G x H`. The set of vertices is `V(G) x V(H)`, there is an edge between `(x, x')` and `(y, y')` iff `(x, y) ∈ E(G)` and `(x',y') ∈ E(H)`. 

Here `V(G), E(G)` is the set of edges and vertices of `G` respectively.

For more information see
[Hammack R.H. (2018) Digraphs Products. In: Bang-Jensen J., Gutin G. (eds) Classes of Directed Graphs. Springer Monographs in Mathematics. Springer, Cham](https://link.springer.com/chapter/10.1007/978-3-319-71840-8_10)

james-d-mitchell:
Good idea @reiniscirpons! Any possibility of a PR in this direction??

PR #228, opened 2019-08-16
Add DigraphCartesianProduct and DigraphDirectProduct methods

reiniscirpons:
This PR implements two types of digraph product as discussed in #226 for a collection of digraphs. `DigraphCartesianProduct` computes the [Cartesian product](https://en.wikipedia.org/wiki/Cartesian_product_of_graphs) of a collection of digraphs and `DigraphDirectProduct` computes the [direct product](https://en.wikipedia.org/wiki/Tensor_product_of_graphs) of a collection of digraphs. It also includes two new attributes `DigraphCartesianProductProjections` and `DigraphDirectProductProjections` for storing projections onto the coordinates of vertices.

Since the vertex set of `DigraphCartesianProduct(G, H)` is the Cartesian product of the vertex sets of `G` and `H`, but the `Digraphs` package does not permit a vertex set to contain pairs of integers, only integers themselves, then we encode a pair `(i, j)` as `i + n * (j-1)` where `n = DigraphNrVertices(G)`.

If `D = DigraphCartesianProduct(gr_1, gr_2, ...)` then `DigraphCartesianProductProjections(D)` returns a list of transformations, where the `i`-th transformation corresponds to a projection onto a copy of `gr_i` contained within `D`. The projections are useful as they allow for us to reconstruct (under some circumstances) the graphs used for creating the product, and also provide for a way of regaining the information of the coordinates of each vertex (which were lost when we decided to encode `(i, j)` as `i + n * (j-1)`). 

To make sure my code works with mutable as well as immutable digraphs, I used DigraphJoin as a basis, changing only the moving parts responsible for creating the digraph and leaving the more technical bits untouched. It seems like this has worked, but I would appreciate if someone told me whether everything is sound with regards to mutability.

Not exactly sure about the time complexity. For two graphs I think it should be `O(E_1 + E_2)` for Cartesian product and `O(E_1 * E_2)` for Direct product, where `E_1, E_2` is the number of edges of each digraph.

### DIFF
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -1785,15 +1785,18 @@ gap> ChromaticNumber(D);
     <C>DigraphCartesianProductProjections</C> returns a list <C>proj</C>
     such that <C>proj[i]</C> is the projection onto the <C>i</C>-th
     coordinate of the product.
+    <P/>
 
     A projection is an idempotent endomorphism of <A>digraph</A>. If
     <C>gr1, gr2, ...</C> are all loopless digraphs, then the restriction
     of <A>digraph</A> onto the image of <C>proj[i]</C> is isomorphic to
     <C>gr_i</C>.
+    <P/>
 
-    Currently this attribute is only set upon creating a graph via
+    Currently this attribute is only set upon creating a digraph via
     <C>DigraphCartesianProduct</C> and there is no way of calculating
-    it for general graphs.
+    it for other digraphs.
+    <P/>
 
     For more information see <Ref Oper="DigraphCartesianProduct"/> 
 
@@ -1826,14 +1829,17 @@ true
     <C>DigraphDirectProductProjections</C> returns a list <C>proj</C>
     such that <C>proj[i]</C> is the projection onto the <C>i</C>-th
     coordinate of the product.
+    <P/>
 
     A projection is an idempotent endomorphism of <A>digraph</A>. If
     <C>gr1, gr2, ...</C> are all loopless digraphs, then the projection 
     of <A>digraph</A> via <C>proj[i]</C> is isomorphic to <C>gr_i</C>.
+    <P/>
 
-    Currently this attribute is only set upon creating a graph via
+    Currently this attribute is only set upon creating a digraph via
     <C>DigraphDirectProduct</C> and there is no way of calculating
-    it for general graphs.
+    it for other digraphs.
+    <P/>
 
     For more information see <Ref Oper="DigraphDirectProduct"/> 
 
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -1774,3 +1774,83 @@ gap> ChromaticNumber(D);
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="DigraphCartesianProductProjections">
+<ManSection>
+  <Attr Name="DigraphCartesianProductProjections" Arg="digraph"/>
+  <Returns>A list of transformations.</Returns>
+  <Description>
+    If <A>digraph</A> is a Cartesian product digraph, 
+    <C>digraph = DigraphCartesianProduct(gr_1, gr_2, ... )</C>, then
+    <C>DigraphCartesianProductProjections</C> returns a list <C>proj</C>
+    such that <C>proj[i]</C> is the projection onto the <C>i</C>-th
+    coordinate of the product.
+
+    A projection is an idempotent endomorphism of <A>digraph</A>. If
+    <C>gr1, gr2, ...</C> are all loopless digraphs, then the restriction
+    of <A>digraph</A> onto the image of <C>proj[i]</C> is isomorphic to
+    <C>gr_i</C>.
+
+    Currently this attribute is only set upon creating a graph via
+    <C>DigraphCartesianProduct</C> and there is no way of calculating
+    it for general graphs.
+
+    For more information see <Ref Oper="DigraphCartesianProduct"/> 
+
+    <Example><![CDATA[
+gap> D := DigraphCartesianProduct(ChainDigraph(3), CycleDigraph(4),
+> Digraph([[2], [2]]));;
+gap> HasDigraphCartesianProductProjections(D);
+true
+gap> proj := DigraphCartesianProductProjections(D);; Length(proj);
+3
+gap> IsIdempotent(proj[2]);
+true
+gap> RankOfTransformation(proj[3]);
+2
+gap> S := ImageSetOfTransformation(proj[2]);;
+gap> IsIsomorphicDigraph(CycleDigraph(4), InducedSubdigraph(D, S));
+true
+]]></Example>
+  </Description>
+</ManSection>
+<#/GAPDoc>
+
+<#GAPDoc Label="DigraphDirectProductProjections">
+<ManSection>
+  <Attr Name="DigraphDirectProductProjections" Arg="digraph"/>
+  <Returns>A list of transformations.</Returns>
+  <Description>
+    If <A>digraph</A> is a direct product digraph, 
+    <C>digraph = DigraphDirectProduct(gr_1, gr_2, ... )</C>, then
+    <C>DigraphDirectProductProjections</C> returns a list <C>proj</C>
+    such that <C>proj[i]</C> is the projection onto the <C>i</C>-th
+    coordinate of the product.
+
+    A projection is an idempotent endomorphism of <A>digraph</A>. If
+    <C>gr1, gr2, ...</C> are all loopless digraphs, then the projection 
+    of <A>digraph</A> via <C>proj[i]</C> is isomorphic to <C>gr_i</C>.
+
+    Currently this attribute is only set upon creating a graph via
+    <C>DigraphDirectProduct</C> and there is no way of calculating
+    it for general graphs.
+
+    For more information see <Ref Oper="DigraphDirectProduct"/> 
+
+    <Example><![CDATA[
+gap> D := DigraphDirectProduct(ChainDigraph(3), CycleDigraph(4),
+> Digraph([[2], [2]]));;
+gap> HasDigraphDirectProductProjections(D);
+true
+gap> proj := DigraphDirectProductProjections(D);; Length(proj);
+3
+gap> IsIdempotent(proj[2]);
+true
+gap> RankOfTransformation(proj[3]);
+2
+gap> P := DigraphRemoveAllMultipleEdges(ReducedDigraph(OnDigraphs(D, proj[2])));; gap> IsIsomorphicDigraph(CycleDigraph(4), P);
+true
+]]></Example>
+  </Description>
+</ManSection>
+<#/GAPDoc>
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -1788,14 +1788,14 @@ gap> ChromaticNumber(D);
     <P/>
 
     A projection is an idempotent endomorphism of <A>digraph</A>. If
-    <C>gr1, gr2, ...</C> are all loopless digraphs, then the restriction
-    of <A>digraph</A> onto the image of <C>proj[i]</C> is isomorphic to
-    <C>gr_i</C>.
+    <C>gr1, gr2, ...</C> are all loopless digraphs, then the induced 
+    subdigraph of <A>digraph</A> on the image of <C>proj[i]</C> is 
+    isomorphic to <C>gr_i</C>.
     <P/>
 
-    Currently this attribute is only set upon creating a digraph via
-    <C>DigraphCartesianProduct</C> and there is no way of calculating
-    it for other digraphs.
+    Currently this attribute is only set upon creating an immutable 
+    digraph via <C>DigraphCartesianProduct</C> and there is no way 
+    of calculating it for other digraphs.
     <P/>
 
     For more information see <Ref Oper="DigraphCartesianProduct"/> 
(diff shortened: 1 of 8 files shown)


wilfwilson:
Hi @reiniscirpons, I've been making some fairly major changes to the Digraphs package as part of introducing mutable digraphs and working towards Digraph v1.0. Unfortunately that means that I've created conflicts with this PR, so this will need to be rebased.

(However, I'm not finished making the big changes, so you should probably wait until next week before resolving the conflicts.)

wilfwilson:
Should be safe to rebase and resolve conflicts now.

doc/attr.xml
```diff
@@ -1774,3 +1774,85 @@ gap> ChromaticNumber(D);
...
+<#GAPDoc Label="DigraphCartesianProductProjections">
+<ManSection>
+  <Attr Name="DigraphCartesianProductProjections" Arg="digraph"/>
+  <Returns>A list of transformations.</Returns>
+  <Description>
+    If <A>digraph</A> is a Cartesian product digraph, 
+    <C>digraph = DigraphCartesianProduct(gr_1, gr_2, ... )</C>, then
+    <C>DigraphCartesianProductProjections</C> returns a list <C>proj</C>
+    such that <C>proj[i]</C> is the projection onto the <C>i</C>-th
+    coordinate of the product.
```
wilfwilson: Please put paragraph breaks after the things you want paragraphs, so that they compile into separate paragraphs (and in the documentation for direct product, too).
```suggestion
    coordinate of the product.
    <P/>
```

doc/attr.xml
```diff
@@ -1774,3 +1774,85 @@ gap> ChromaticNumber(D);
...
+
+    A projection is an idempotent endomorphism of <A>digraph</A>. If
+    <C>gr1, gr2, ...</C> are all loopless digraphs, then the restriction
+    of <A>digraph</A> onto the image of <C>proj[i]</C> is isomorphic to
+    <C>gr_i</C>.
+
+    Currently this attribute is only set upon creating a graph via
```
wilfwilson:
```suggestion
    Currently this attribute is only set upon creating a digraph via
```

doc/attr.xml
```diff
@@ -1774,3 +1774,85 @@ gap> ChromaticNumber(D);
...
+    <C>DigraphCartesianProduct</C> and there is no way of calculating
+    it for general graphs.
```
wilfwilson: The phrase "general graphs" might suggest that there's some technical meaning, so I suggest the following.
```suggestion
    it for other digraphs.
```

doc/attr.xml
wilfwilson: This means that `DigraphCartesianProductProjections` can only ever do anything for immutable digraphs. In particular, the projections are immediately thrown away upon creation for mutable digraphs. I think this should be mentioned at least. Do you have any alternative suggestions @james-d-mitchell?

doc/oper.xml
```diff
@@ -962,6 +962,96 @@ gap> DigraphJoin(gr, gr2);
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="DigraphCartesianProduct">
+<ManSection>
+  <Func Name="DigraphCartesianProduct" Arg="gr1, gr2, ..." Label="for an arbitrary 
+   number of digraphs"/>
```
wilfwilson:
```suggestion
   non-zero number of digraphs"/>
```

doc/oper.xml
```diff
@@ -962,6 +962,96 @@ gap> DigraphJoin(gr, gr2);
...
+    In the first form, if <A>gr1</A>, <A>gr2</A>, etc. are digraphs, then
+    <C>DigraphCartesianProduct</C> returns their Cartesian product.
+
+    In the second form, if <A>list</A> is a non-empty list of digraphs,
+    then <C>DigraphCartesianProduct</C> returns the Cartesian product of 
+    the digraphs contained in the list. <P/>
+
+    The Cartesian product of two digraphs <C>G</C>, <C>H</C> is a digraph with 
+    vertex set <C>Cartesian(DigraphVertices(G), DigraphVertices(H))</C> and there
+    is an edge connecting <C>[u, u']</C> with <C>[v, v']</C> iff <C> u = v </C>
```
wilfwilson: Can you make this paragraph more specific about the direction of edges in all cases. Would `u is connected to v` mean that there is an edge `[u,v]`, or is it okay for the edge to go the other way `[v,u]`? From reading this, it's not clearly.

For instance, do you mean?
```suggestion
    is an edge from <C>[u, u']</C> to <C>[v, v']</C> iff <C> u = v </C> 
```

doc/oper.xml
```diff
@@ -962,6 +962,96 @@ gap> DigraphJoin(gr, gr2);
...
+  <Description>
+    In the first form, if <A>gr1</A>, <A>gr2</A>, etc. are digraphs, then
+    <C>DigraphCartesianProduct</C> returns their Cartesian product.
+
+    In the second form, if <A>list</A> is a non-empty list of digraphs,
+    then <C>DigraphCartesianProduct</C> returns the Cartesian product of 
+    the digraphs contained in the list. <P/>
+
+    The Cartesian product of two digraphs <C>G</C>, <C>H</C> is a digraph with 
+    vertex set <C>Cartesian(DigraphVertices(G), DigraphVertices(H))</C> and there
```
wilfwilson: Mathematically, fine, but since the vertex set of a digraph is a list of positive integers in the Digraphs package, you also need to say how this mathematical definition actually relates to what happens in the package.

doc/attr.xml
```diff
@@ -1774,3 +1774,85 @@ gap> ChromaticNumber(D);
...
+  <Description>
+    If <A>digraph</A> is a Cartesian product digraph, 
+    <C>digraph = DigraphCartesianProduct(gr_1, gr_2, ... )</C>, then
+    <C>DigraphCartesianProductProjections</C> returns a list <C>proj</C>
+    such that <C>proj[i]</C> is the projection onto the <C>i</C>-th
+    coordinate of the product.
+
+    A projection is an idempotent endomorphism of <A>digraph</A>. If
+    <C>gr1, gr2, ...</C> are all loopless digraphs, then the restriction
+    of <A>digraph</A> onto the image of <C>proj[i]</C> is isomorphic to
```
wilfwilson: Sorry, I can't work out immediately exactly what
```
the restriction of <A>digraph</A> onto the image of <C>proj[i]</C> is isomorphic to <C>gr_i</C>.
```
means. Could you try and explain, and then maybe we can work out a clearer way of phrasing it.

doc/attr.xml
```diff
@@ -1774,3 +1774,91 @@ gap> ChromaticNumber(D);
...
+    coordinate of the product.
+    <P/>
+
+    A projection is an idempotent endomorphism of <A>digraph</A>. If
+    <C>gr1, gr2, ...</C> are all loopless digraphs, then the restriction
+    of <A>digraph</A> onto the image of <C>proj[i]</C> is isomorphic to
+    <C>gr_i</C>.
+    <P/>
+
+    Currently this attribute is only set upon creating a digraph via
```
wilfwilson:
```suggestion
    Currently this attribute is only set upon creating an immutable digraph via
```

doc/attr.xml
```diff
@@ -1774,3 +1774,91 @@ gap> ChromaticNumber(D);
...
+    such that <C>proj[i]</C> is the projection onto the <C>i</C>-th
+    coordinate of the product.
+    <P/>
+
+    A projection is an idempotent endomorphism of <A>digraph</A>. If
+    <C>gr1, gr2, ...</C> are all loopless digraphs, then the projection 
+    of <A>digraph</A> via <C>proj[i]</C> is isomorphic to <C>gr_i</C>.
+    <P/>
+
+    Currently this attribute is only set upon creating a digraph via
```
wilfwilson:
```suggestion
    Currently this attribute is only set upon creating an immutable digraph via
```

doc/attr.xml
wilfwilson: I think it's just best to draw attention that this only works for immutable digraphs, as I will suggest in a concrete suggestion.

doc/attr.xml
wilfwilson: Ah @reiniscirpons, do you mean `the induced subdigraph of <A>digraph</A>` on the image of <C>proj[i]</C>`? If so, then please update it to use `induced subdigraph`.

doc/attr.xml
```diff
@@ -1774,3 +1774,91 @@ gap> ChromaticNumber(D);
...
+    If <A>digraph</A> is a direct product digraph, 
+    <C>digraph = DigraphDirectProduct(gr_1, gr_2, ... )</C>, then
+    <C>DigraphDirectProductProjections</C> returns a list <C>proj</C>
+    such that <C>proj[i]</C> is the projection onto the <C>i</C>-th
+    coordinate of the product.
+    <P/>
+
+    A projection is an idempotent endomorphism of <A>digraph</A>. If
+    <C>gr1, gr2, ...</C> are all loopless digraphs, then the projection 
+    of <A>digraph</A> via <C>proj[i]</C> is isomorphic to <C>gr_i</C>.
```
wilfwilson: Please use the same language as above, with induced subdigraph, if that's what it actually is.

doc/oper.xml
```diff
@@ -962,6 +962,105 @@ gap> DigraphJoin(gr, gr2);
...
+
+<#GAPDoc Label="DigraphDirectProduct">
+<ManSection>
+  <Func Name="DigraphDirectProduct" Arg="gr1, gr2, ..." Label="for an arbitrary 
+   non-zero number of digraphs"/>
+  <Func Name="DigraphDirectProduct" Arg="list" Label="for a list of digraphs"/>
+  <Returns>A digraph.</Returns>
+  <Description>
+    In the first form, if <A>gr1</A>, <A>gr2</A>, etc. are digraphs, then
+    <C>DigraphDirectProduct</C> returns their direct product.
```
wilfwilson:
```suggestion
    <C>DigraphDirectProduct</C> returns a digraph isomorphic to their direct product.
```

doc/oper.xml
```diff
@@ -962,6 +962,105 @@ gap> DigraphJoin(gr, gr2);
...
+    <P/>
+
+    In the second form, if <A>list</A> is a non-empty list of digraphs,
+    then <C>DigraphDirectProduct</C> returns the direct product of
```
wilfwilson:
```suggestion
    then <C>DigraphDirectProduct</C> returns a digraph isomorphic to the direct product of 
```

doc/oper.xml
```diff
@@ -962,6 +962,105 @@ gap> DigraphJoin(gr, gr2);
...
 
+<#GAPDoc Label="DigraphCartesianProduct">
+<ManSection>
+  <Func Name="DigraphCartesianProduct" Arg="gr1, gr2, ..." Label="for an arbitrary 
+   non-zero number of digraphs"/>
+  <Func Name="DigraphCartesianProduct" Arg="list" Label="for a list of digraphs"/>
+  <Returns>A digraph.</Returns>
+  <Description>
+    In the first form, if <A>gr1</A>, <A>gr2</A>, etc. are digraphs, then
+    <C>DigraphCartesianProduct</C> returns their Cartesian product.
```
wilfwilson:
```suggestion
    <C>DigraphCartesianProduct</C> returns a digraph isomorphic to their Cartesian product.
```

doc/oper.xml
```diff
@@ -962,6 +962,105 @@ gap> DigraphJoin(gr, gr2);
...
+    <P/>
+
+    In the second form, if <A>list</A> is a non-empty list of digraphs,
+    then <C>DigraphCartesianProduct</C> returns the Cartesian product of
```
wilfwilson:
```suggestion
    then <C>DigraphCartesianProduct</C> returns a digraph isomorphic to the Cartesian product of 
```

doc/oper.xml
```diff
@@ -962,6 +962,105 @@ gap> DigraphJoin(gr, gr2);
...
+    vertex set <C>[1 .. DigraphNrVertices(G)*DigraphNrVertices(H)]</C> instead. The vertex
+    pair can be somewhat regained by using <Ref Attr="DigraphDirectProductProjections"/>.
+    <P/>
+ 
+    As the direct product is associative, the direct product of a collection 
+    of digraphs <A>gr1</A>, <A>gr2</A>, <C>...</C> is computed in the obvious 
+    fashion. 
+    <P/>
+
+    Note that previously set <Ref Oper="DigraphVertexLabels"/> will be lost.
```
wilfwilson: This sentence doesn't seem correct, having looked at the code. Please describe what the code does - you do actually set the vertex labels! And this can typically be used to work out the correspondence between vertices of the individual digraphs, and the vertices of the product - so perhaps this corrected sentence, and `"The vertex pair can be somewhat regained by using <Ref Attr="DigraphDirectProductProjections"/>."` can be combined somehow.

Analogous comment for the Cartesian product.

doc/attr.xml
reiniscirpons: Hey, for Cartesian product your comment about induced subdigraph was spot on - that was exactly what I was trying to say. For Direct product what I actually mean is that the image of the digraph under the projection is isomorphic to the original digraph. I have made changes accordingly

wilfwilson:
Whoops, it seems that GitHub has some key commands that I accidentally pressed, which meant that I accidentally merged this before the tests had finished. Thankfully I'm sure they were going to pass anyway!

PR #229, opened 2019-08-20
Make digraphs immutable via MakeImmutable

wilfwilson:
Using the function `PostMakeImmutable` to do the important stuff previously done in `MakeImmutableDigraph`.


james-d-mitchell:
Haha, awesome, if only I'd known about `PostMakeImmutable` before! Shouldn't you also delete `IsValidDigraph`?

wilfwilson:
Yes, I'll delete `IsValidDigraph` once the tests pass here. Just keeping it in for now as a sanity check.

PR #243, opened 2019-09-16
Make gap/cliques.gi always return immutable lists (of lists)

### LINKED ISSUE #240: strange mutablity in clique outputs
Some of the clique functions in the master branch, such as DigraphMaximalCliques, return a mutable list of immutable lists, this is a bit strange.

For example:
```gap
gap> D := Digraph(IsMutableDigraph,
> [[2, 3], [1, 3], [1, 2, 4], [3, 5, 6], [4, 6], [4, 5]]);;
gap> cliques := DigraphMaximalCliques(D);
[ [ 1, 2, 3 ], [ 4, 5, 6 ], [ 3, 4 ] ]
gap> IsMutable(cliques);
true
gap> IsMutable(cliques[1]);
false
```
the list [ [ 1, 2, 3 ], [ 4, 5, 6 ], [ 3, 4 ] ] is mutable but the lists [ 1, 2, 3 ], [ 4, 5, 6 ] and [ 3, 4 ] are immutable.

wilfwilson:
The problem with #240 was that some cliques were created using the function `Orb`, which gives immutable objects, but otherwise the cliques tended to be mutable.

Now everything is always immutable.

Fixes #240.

### DIFF
--- a/gap/cliques.gi
+++ b/gap/cliques.gi
@@ -114,10 +114,10 @@ function(arg)
   fi;
   arg[1] := DigraphMutableCopyIfMutable(arg[1]);
   arg[1] := DigraphDual(DigraphRemoveAllMultipleEdges(arg[1]));
-  return CallFuncList(DIGRAPHS_Clique,
-                      Concatenation([true],
-                                    [arg[1]],
-                                    arg{[2 .. Length(arg)]}));
+  return MakeImmutable(CallFuncList(DIGRAPHS_Clique,
+                                    Concatenation([true],
+                                                  [arg[1]],
+                                                  arg{[2 .. Length(arg)]})));
 end);
 
 InstallGlobalFunction(DigraphIndependentSet,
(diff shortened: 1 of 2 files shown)


wilfwilson:
Or @james-d-mitchell do we always want to return mutable lists of mutable lists? (excepting that immutability is forced for attributes)

james-d-mitchell:
Good question, I'm not sure, I think we should just make a policy and then make sure we follow it (which I don't think we've done until now). Wouldn't the best policy be to do whatever is the least work (computationally)? Return an immutable list of immutable lists, and then the user can copy that if they want? Rather than us copying it for them, which is maybe unnecessary?

wilfwilson:
In all cases, even for mutable digraphs? I'll give it a think, but I think that's fine.

james-d-mitchell:
Yes, that's what I'm suggesting. It isn't possible to convert in-place an immutable to a mutable object in GAP, so a copy is always necessary to do this, but you can convert a mutable to an immutable in-place, so isn't the cheapest thing to do this?

wilfwilson:
Yep I think so.

james-d-mitchell:
Cheap = good in this context I think, but not when buying a washing machine.

wilfwilson:
I've made the change @james-d-mitchell. This will make it much easier to document too.

james-d-mitchell:
Thanks @wilfwilson, I expect that this will come up in other places than here too!

PR #253, opened 2019-09-25
Update orbits doc for mutable digraphs

le27:


wilfwilson:
Can you please rebase this on the latest commit of the master branch? It seems that this PR contains a bunch of other changes, not just orbits stuff.

wilfwilson:
Thanks!

doc/orbits.xml
```diff
@@ -13,11 +13,11 @@
   <Attr Name="DigraphGroup" Arg="digraph"/>
   <Returns>A permutation group.</Returns>
   <Description>
-    If <A>digraph</A> was created knowing a subgroup of its automorphism group,
-    then this group is stored in the attribute <C>DigraphGroup</C>. If
-    <A>digraph</A> is not created knowing a subgroup of its automorphism group,
-    then <C>DigraphGroup</C> returns the entire automorphism group of
-    <A>digraph</A>.
+    If <A>digraph</A> was immutable and created knowing a subgroup of
```
james-d-mitchell: "was immutable and" -> "is immutable and was"

doc/orbits.xml
```diff
@@ -13,11 +13,11 @@
...
+    its automorphism group, then this group is stored in the attribute
+    <C>DigraphGroup</C>. If <A>digraph</A> is mutable, or was not created
+    knowing a subgroup of its automorphism group, then <C>DigraphGroup</C>
+    returns the entire automorphism group of <A>digraph</A>.
```
james-d-mitchell: Add "Note that if digraph is mutable, then the automorphism group is recomputed every time this function is called."

doc/orbits.xml
```diff
@@ -123,21 +138,27 @@ gap> 5 / gens[sch[5]];
 <ManSection>
   <Attr Name="DigraphOrbitReps" Arg="digraph"/>
   <Returns>
-    A list of integers.
+    An immutable list of integers.
   </Returns>
```
james-d-mitchell: Wasn't there some discussion about returning mutable lists if that was an option, because mutable lists can always be made immutable if desirable, and but an immutable list has to be copied to be made mutable?

wilfwilson: This is an attribute though, so the value has to be mutable.

doc/orbits.xml
```diff
@@ -171,18 +198,22 @@ Group(())
 <ManSection>
   <Attr Name="DigraphOrbits" Arg="digraph"/>
   <Returns>
-    A list of lists of integers.
+    An immutable list of immutable lists of integers.
```
wilfwilson: The sublists in an immutable list are necessarily immutable (something obviously isn't immutable if it contains mutable things). So the second immutable is redundant here.

PR #262, opened 2019-09-25
DigraphFromDigraph6String ability to return mutable digraphs

MTWhyte:
This pull request adds the functionality for `DigraphFromDigraph6String` to return mutable digraphs.

### DIFF
--- a/gap/io.gi
+++ b/gap/io.gi
@@ -795,7 +795,8 @@ InstallMethod(DigraphFromDigraph6String, "for a function and a string",
 InstallMethod(DigraphFromDigraph6String, "for a string",
 [IsString], s -> DigraphFromDigraph6String(IsImmutableDigraph, s));
 
-BindGlobal("DIGRAPHS_FromSparse6String",
+InstallMethod(DigraphFromSparse6StringCons, "for IsMutableDigraph and a string",
+[IsMutableDigraph, IsString],
 function(func, s)
   local list, n, start, blist, pos, num, bpos, k, range, source, len, v, i,
   finish, x, j;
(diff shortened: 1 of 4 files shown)


james-d-mitchell:
Looks good! What about the other functions in the same file??

gap/io.gi
```diff
@@ -761,17 +762,28 @@ function(func, s)
...
 
   if legacy then  # source and range are reversed
-    return func(rec(DigraphNrVertices := n,
-                    DigraphSource     := range,
-                    DigraphRange      := source));
-  fi;
-  return func(rec(DigraphNrVertices := n,
-                  DigraphRange      := range,
-                  DigraphSource     := source));
+    return DigraphNC(IsMutableDigraph,
```
wilfwilson: I don't have any idea why this code was originally written like this, but you should check with at least @james-d-mitchell that we're losing anything important by removing the ability of `DIGRAPHS_FromDigraph6String` to take an arbitrary "constructor", and forcing `DigraphNC`.

gap/io.gi
```diff
@@ -761,17 +762,28 @@ function(func, s)
...
 end);
 
+InstallMethod(DigraphFromDigraph6StringCons,
+"for IsImmutableDigraph and a string",
+[IsImmutableDigraph, IsString],
+{filt, s} -> MakeImmutable(DigraphFromDigraph6StringCons(IsMutableDigraph, s)));
+
+InstallMethod(DigraphFromDigraph6String, "for a function and a string",
+[IsFunction, IsString],
+{func, s} -> DigraphFromDigraph6StringCons(func, s));
```
wilfwilson:
```suggestion
DigraphFromDigraph6StringCons);
```

gap/io.gi
MTWhyte: Yeah, we're definitely ok to do this - getting rid of `DIGRAPHS_FromDigraph6String` was actually James's suggestion.

MTWhyte:
Great, I'll change the others too!

wilfwilson:
Great, thanks a lot. Really handy.

gap/io.gi
james-d-mitchell: I concur, it was my idea, I think this is the best solution.

gap/io.gi
james-d-mitchell: Nice

wilfwilson:
@MTWhyte do you mean you'll do it in this PR, or in a separate PR?

MTWhyte:
@wilfwilson which do you think would be better?

wilfwilson:
I think you may as well do it all in this PR, but I don’t mind if you want to do it separately.

MTWhyte:
I've now updated the 'friends' too, namely `DigraphFromGraph6String`, `DigraphFromSparse6String`, `DigraphFromDiSparse6String` and `DigraphFromPlainTextString`.

PR #267, opened 2019-10-28
Speed up `OnDigraphs` for a digraph and a permutation

ChrisJefferson:
This came up from some code which was heavily using digraphs. this almost doubled the speed of the algorithm, as it turns out these GAP-level checks were dominating the run-time of OnDigraphs.

* Instead of checking if the permutation fixes the vertices, check
if it is the identity (which is much cheaper, although catches
less cases)

* Do not check explictly if the permutation maps the vertices
  of the graph to itself, do the mapping then check afterwards.

### DIFF
--- a/gap/oper.gi
+++ b/gap/oper.gi
@@ -997,17 +997,15 @@ InstallMethod(OnDigraphs, "for a mutable digraph by out-neighbours and a perm",
 [IsMutableDigraph and IsDigraphByOutNeighboursRep, IsPerm],
 function(D, p)
   local out, permed;
-  if p = () then
+  if SmallestMovedPoint(p) > DigraphNrVertices(D) then
     return D;
+  elif ForAny(DigraphVertices(D), i -> i ^ p > DigraphNrVertices(D)) then
+    ErrorNoReturn("the 2nd argument <p> must be a permutation that permutes ",
+                  "the vertices of the digraph <D> that is the 1st argument.");
   fi;
 
   out := D!.OutNeighbours;
   permed := Permuted(out, p);
-  if Length(permed) > DigraphNrVertices(D) then
-    ErrorNoReturn("the 2nd argument <p> must be a permutation that permutes ",
-                  "the vertices of the digraph <D> that is the 1st argument,");
-  fi;
-
   out{DigraphVertices(D)} := permed;
   Apply(out, x -> OnTuples(x, p));
   ClearDigraphEdgeLabels(D);
@@ -1017,7 +1015,7 @@ end);
 InstallMethod(OnDigraphs, "for a immutable digraph and a perm",
 [IsImmutableDigraph, IsPerm],
 function(D, p)
-  if p = () then
+  if SmallestMovedPoint(p) > DigraphNrVertices(D) then
     return D;
   fi;
   return MakeImmutable(OnDigraphs(DigraphMutableCopy(D), p));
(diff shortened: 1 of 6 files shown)


gap/oper.gi
```diff
@@ -608,15 +608,19 @@ end);
 InstallMethod(OnDigraphs, "for a mutable digraph by out-neighbours and a perm",
 [IsMutableDigraph and IsDigraphByOutNeighboursRep, IsPerm],
 function(D, p)
-  local out;
-  if ForAll(DigraphVertices(D), i -> i ^ p = i) then
+  local out, permed;
+  if p = () then
```
wilfwilson: In one sense this is worse: previously we were testing whether `p` acts as the identity on `DigraphVertices(D)`, we can be the case even if `p` is not the identity permutation. But I don't mind too much about losing that case.

Is `p = ()` optimised somehow? Do all permutations inherently know whether or not they are the identity, for instance?

gap/oper.gi
```diff
@@ -608,15 +608,19 @@ end);
...
     return D;
-  elif ForAny(DigraphVertices(D), i -> i ^ p > DigraphNrVertices(D)) then
+  fi;
+
+  out := D!.OutNeighbours;
+  permed := Permuted(out, p);
+  if Length(permed) > DigraphNrVertices(D) then
```
wilfwilson: If `p`  is invalid, `Permuted` simply won't work properly (I think) and cause its own `Error` to happen, which on the one hand is good because it means that the method prevents something improper happen, but it does mean that we lose the ability to give a proper error message.

wilfwilson:
Another solution is that we could create some `NC` versions of functions that are particularly performance-critical, including `OnDigraphsNC`.

gap/oper.gi
ChrisJefferson: P=() isn't specifically optimised, but it will almost always be very fast, espically if (as is common) the permutation moves the largest point it is defined on, as that is checked first.

ChrisJefferson: Just to say, we could use `SmallestMovedPoint` is we really care about saving this optimisation. That is still much faster than the old code, and would keep the same answer.

wilfwilson:
@ChrisJefferson I know this is a bit late (sorry) but I intend to come up with a resolution for this soon that keeps the performance improvement.

james-d-mitchell:
I'd propose improving the performance of `OnDigraphs` at gap days in Brussels in 2025, if we have the time.

reiniscirpons:
I tried my best to incorporate the improved methods, but I just do not see the timing improvements claimed. I settled on implementing a separate `OnDigraphsNC` method which skips the more expensive `ForAll` and `ForEach` checks and only does the `p = ()` optimization. I then replaced internal usage of the `OnDigraphs` method with `OnDigraphsNC`. Here are some runtime stats:

|Test |Old method time, ms|Proposed new `OnDigraphsNC` method time, ms| Proposed new `OnDigraphs` method time, ms
|-|-|-|-|
| 1000 random digraphs with 100 vertices, `IsIsomorphicDigraph` | 703 | **689** | - |
| 829 random topologically sortable digraphs with 100 vertices, `DigraphTransitiveReduction` | 65 | **52** | - |
|100 random transitive digraphs with 1000 vertices, `IsIsomorphicDigraph` | **5476** | 5374 | - |
|100 random transitive digraphs with 1000 vertices, `DigraphTransitiveReduction` | 24020 | **23947** | - |
|100 random digraphs with 1000 vertices, `OnDigraphs` or `OnDigraphsNC` respectively | 1415 | **1403** | 1487 |

There seem to be some minor improvement (within margin of error in some cases) with the proposed `OnDigraphsNC` method, but it slows down `OnDigraphs` by a bit.

james-d-mitchell:
Thanks @reiniscirpons looks good

PR #268, opened 2019-11-06
Add DigraphDijkstraS and DigraphDijkstraST

mariatsalakou:
This does the same as #128 with tests and documentation


james-d-mitchell:
It looks like the datastructures package doesn't work with load and save workspace in 32-bit mode for some reason.

PR #275, opened 2019-11-28
Release candidate for v1.0.2

wilfwilson:


wilfwilson:
I'm not going to let the tests run since:
1. They're duplicated (the tests run both because `rc-*` is a branch on this repo, and so every new push gets tested, and the tests because this is a PR).
2. This branch only differs from `stable-1.0` by one line in `README.md`, and so the tests past here iff they pass there.

PR #225, opened 2019-07-20
Add --with-external-bliss option to unvendor bliss

isuruf:
cc @james-d-mitchell, @kiwifb, @dimpase

PR #278, opened 2020-01-16
Make bliss return the size of the automorphism group

ChrisJefferson:
Bliss previously returned the size of the automorphism group
as a double, which is not useful for GAP. Instead of trying to
build the size of the group during search, store a list of numbers
which must be multiplied together and return that to GAP.

This PR isn't ready yet (various tests fail). The main problem is it's going to be annoying to make the tests work with, and without, an external bliss. I was wondering why that option is there, and who is using it?


james-d-mitchell:
Thanks @ChrisJefferson, I think the sage guys are using their own bliss instead of the one vendored by Digraphs. Agreed that it is annoying to have to tests working in two different cases. Is one possibility to submit a patch to 

https://github.com/mkoeppe/bliss

which is I think where Sage gets its Bliss from. Maybe @nthiery or @embray can offer more details of the rationale behind this if necessary.

james-d-mitchell:
Ah sorry actually it looks like it me that wanted to have a GAP conda package and that @isuruf wrote the code for external bliss in PR #225.

wilfwilson:
Anyway, since the tests are passing, I'd be happy for this to be merged. Everyone else agree?

In which case I recommend a squash and merge, to get rid of the two commits that I added.

isuruf:
I'm not sure I understand the need for a list. Why not is the gmp class BigNum and convert that to GAP?

ChrisJefferson:
Because I'd have to figure out how to get bliss to use the same gmp library as GAP, get all the headers linked together, and also the gmp object in bliss would be allocated with malloc and we'd have to get it moved into gasman memory. I imagine all these thighs are possible, but I'm not positive how to do them correctly.

Sent from Nine<http://www.9folders.com/>
________________________________

ChrisJefferson:
I might just pull in @fingolfin, just to see if he thinks making gmp variables in bliss and then getting them turned in to GAP numbers is reasonable / easy.

isuruf:
I'm okay with you using a list here, but would like it if there's a way for distros to use GMP. cc @fbissey

isuruf:
As for GMP, you only need to care about getting the headers correctly. You don't have to link because gap itself will be pulling in GMP before you in import semigroups in GAP.

fingolfin:
I'm not fully informed about the context of this and what are possible options and what not; but let me point out that we added `MakeObjInt` in GAP 4.10 specifically to convert GMP `mpz_t` to GAP integers. This function does not require that the same GMP is used. If you need to support older GAP versions, it would be safe to copy the implementation of `MakeObjInt` and use if inside a suitable `#if` (As for GAP <= 4.9 of course you know the internal GAP format won't change; but in theory, it might in future GAP versions; hence us `MakeObjInt` there).

If using `MakeObjInt` is possible and somebody tells me why, I am happy to think about other options.

james-d-mitchell:
I'm also happy with this.

fingolfin:
(Just to clarify, I am not objecting to this PR or taking any stance, just replying to @ChrisJefferson's question :-)

ChrisJefferson:
@fingolfin : Thanks, I didn't think you were, I was just worried about trying to use GMP "normally" and then trying to get that number into GAP.

Personally, I do think this is a bit of a horrible hack, but bliss is basically abandoned at this point. If anyone else wanted to neaten this up for general use, I'd be happy to help do that, but for my purposes this is "good enough".

PR #283, opened 2020-01-20
Allow specifying vertex colours in IsDigraphHomomorphism etc.

flsmith:
This PR adds the ability to specify vertex colours in the operations
```
IsDigraphHomomorphism
IsDigraphEpimorphism
IsDigraphMonomorphism
IsDigraphEndomorphism
IsDigraphAutomorphism
```

Issue #279, opened 2020-01-19
Allow not calculating the automorphism group before finding homomorphisms?


flsmith:
There are cases where the user knows that calculating the automorphism group of a digraph will take far longer than looking for homomorphisms. For example, if you look for embeddings between these two digraphs with the specified colouring(of which there are none) it takes about a minute to just compute the automorphism group of the target graph, while it really only needs to check one additional vertex!

```
gap> source_colours := [1 .. 5];;
gap> colours := [ 1, 2, 3, 3, 4, 4, 4, 4, 4, 4, 4, 3, 5, 4, 5, 4, 5, 4, 5, 4, 5, 4, 5,
>   4, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 6, 6, 6, 6, 6, 6, 6, 6, 6,
>   6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6,
>   6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6,
>   6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6,
>   6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6,
>   6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 4, 5, 4, 5,
>   4, 5, 4, 5 ];;
gap> D := DigraphFromDiSparse6String(".~?Ad__B`oLa?Jrg@_K@_K@_K@_Md_O?__?_SK?CM?SO?cX?k^?[dBKiBSnB[wBc\\fo]gW^h__iG`aLV@|_A|dCTnAlxA]BC[JpGLqwc");
<immutable digraph with 165 vertices, 38 edges>
gap> C := DigraphDisjointUnion(ChainDigraph(3), ChainDigraph(2));;
gap> partial_map := [1, , 15, 158, 2];;
gap> order := [1, 3, 4, 5, 2];;
gap> Size(HomomorphismDigraphsFinder(source, 
                                     D,
                                     fail,
                                     [],
[...]

PR #285, opened 2020-01-22
Allow user to specify automorphisms in HomomorphismDigraphsFinder

flsmith:
This PR allows the user to specify a subgroup of the automorphism group to `HomomorphismDigraphsFinder`. Representatives of homomorphisms are then returned up to right multiplication by elements of the automorphism subgroup.

This PR also fixes a bug where non-embeddings could be returned when the embedding option was chosen.

This PR deals with #279 and is based on #283

### DIFF
--- a/doc/grahom.xml
+++ b/doc/grahom.xml
@@ -10,7 +10,7 @@
 
 <#GAPDoc Label="HomomorphismDigraphsFinder">
 <ManSection>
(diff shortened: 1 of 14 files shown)


james-d-mitchell:
I'm completely in favour of this, obviously the CI isn't passing and should be fixed first. @flsmith

Issue #155, opened 2018-12-06
Implement maximum (bipartite) matchings


wilfwilson:
https://en.wikipedia.org/wiki/Matching_(graph_theory)

reiniscirpons:
I am currently working on this. Since matching is something that only makes sense for an undirected graph, I will only implement methods for symmetric digraphs.

Currently am implementing the [Blossom algorithm](https://en.wikipedia.org/wiki/Blossom_algorithm) which finds a maximum matching in `O(E*V^2)` time, where `V` is the number of vertices and `E` is number of edges.

Afterwards might upgrade to the Micali-Vazirani algorithm which runs in `O(E*sqrt(V))` time and is the most efficient known algorithm for the problem. Currently thinking of using notes from [this article](https://arxiv.org/abs/1210.4594).

james-d-mitchell:
Resolved in v1.2.0

PR #294, opened 2020-02-21
Add DigraphMaximumMatching

reiniscirpons:
This PR adds the following attributes and property:  
* `DigraphMaximalMatching` - for a digraph, this attribute returns a maximal matching of the digraph.
* `DigraphMaximumMatching` - for a digraph, this attribute returns a maximum matching of the digraph.
* `IsMaximumMatching` - for a digraph, and list of pairs of vertices, returns true if the list is a valid maximum matching of the digraph and false otherwise.  

See issue #155.  

### Definitions:
A matching of a graph is a subset of the edges such that no pair of edges are incident at a vertex. A matching is *maximal* if it cannot be extended by adding an edge. A matching is *maximum* if it has maximum cardinality among all possible matchings. Note that the Digraphs package definition of matching allows the inclusion of loops in a matching.  

### Complexity: 
If what follows let `G` be a digraph with `n` vertices and `m`edges.  
`DigraphMaximalMatching` uses a greedy algorithm, which performs very fast in practice (about `O(n)`).   
`DigraphMaximumMatching`: 
* To handle loops we first matches all of the loop vertices to themselves (we can prove this is always optimal). Loops are then removed by looking at the subdigraph induced by the loopless vertices. This is done so that we could use existing algorithms for maximum matching.
* Since edge direction does not actually matter in a matching, we look at the symmetric closure of the induced graph. 
* We use a greedy matching algorithm to get an initial matching. This is a heuristic which improves most matching algorithms.
* If the induced graph is bipartite, we use the [Hopcroft-Karp algorithm](https://en.wikipedia.org/wiki/Hopcroft%E2%80%93Karp_algorithm) which runs in time `O(sqrt(n)*m)`
* Otherwise we use a general graph matching algorithm which is a derivative of the [one by Gabow](http://cobra.ee.ntu.edu.tw/~khr/khr/p221-gabow.pdf) [1]  and runs in about time `O(n*m*log(n))`, although I am not sure of the exact complexity.
* Afterwards we assemble all of the pieces together to get a final matching of the original graph, and convert it to the list of vertex pairs format that Digraphs uses.  

`IsMaximumMatching`: Here we simply compare the length of the given matching to the one given by `DigraphMaximumMatching`. So takes the same amount of time.

### Improvements:
There are many things that could be improved in the future, if the need arises, for example
#### Improving the general matching algorithm
The current general graph matching algorithm is not the best. In particular the algorithm of [Micali and Vazirani](https://ieeexplore.ieee.org/document/4567800) [2] runs in time `O(sqrt(n)*m)`, and is also faster in practice. The algorithm is also adaptive, and so would exhibit even better performance for graphs close to certain special classes of graphs, see [[3]](https://arxiv.org/pdf/1904.11244). However it is harder to implement. 
#### Improving the heuristics
Currently the only heuristic used is the greedy initial matching.  I implemented a different initial matching - the dynamic mindegree algorithm, however it has complexity `O(m)` (versus the about `O(n)` that the greedy matching does), and so only provides a performance benefit for `n>2500` or so, and makes the overall runtime slower for `n` less than that, so I decided not to include it in the final PR. Maybe worth looking into.

An outline for different heuristics is given in [[4]](https://core.ac.uk/download/pdf/84869336.pdf), [5] and [[6]](http://www.ii.uib.no/~fredrikm/fredrik/papers/JEA2010.pdf).
#### Implement the whole thing in C
The implementation doesn't really use any Gap-specific things, so implementing it in C would probably be possible and would probably make things faster.

#### Improve IsMaximumMatching
We dont actually need to compute the maximum matching of the graph to decide whether or not a given matching is maximum, we can also simply run a single phase of the algorithm to see if we could find an augmenting path. If not then the given matching is maximum, otherwise it is not. This was tricky to implement without duplicating code so I didnt.

### References:
[1] Harold N. Gabow. 1976. An Efficient Implementation of Edmonds’ Algorithm for Maximum Matching on Graphs. J. ACM 23, 2 (April 1976), 221–234. DOI: https://doi.org/10.1145/321941.321942
[2]  S. Micali and V. V. Vazirani, "An O(v|v| c |E|) algoithm for finding maximum matching in general graphs," 21st Annual Symposium on Foundations of Computer Science (sfcs 1980), Syracuse, NY, USA, 1980, pp. 17-27. doi: 10.1109/SFCS.1980.12
[3] Falko Hegerfeld, & Stefan Kratsch. (2019). On adaptive algorithms for maximum matching.  https://https://arxiv.org/abs/1904.11244
[4] Huang, M., & Stein, C. (2017). Extending Search Phases in the Micali-Vazirani Algorithm. SEA.
[5] Kececioglu, J.D., & Pecqueur, A.J. (1998). Computing Maximum-Cardinality Matchings in Sparse General Graphs. Algorithm Engineering.
[6] Johannes Langguth, Fredrik Manne, and Peter Sanders. 2010. Heuristic initialization for bipartite matching problems. J. Exp. Algorithmics 15, Article 1.3 (March 2010), 22 pages. DOI:https://doi.org/10.1145/1671970.1712656

### DIFF
--- a/doc/oper.xml
+++ b/doc/oper.xml
@@ -1665,13 +1665,13 @@ true
     of <A>digraph</A>.  Note that this definition allows a matching to contain
     loops.  See <Ref Prop="DigraphHasLoops" />.  The matching <C>M</C> is
     <E>maximal</E> if it is contained in no larger matching of the digraph,
-    is <E>maximum</E> if it has the greatest cardinality among all mathings and
+    is <E>maximum</E> if it has the greatest cardinality among all matchings and
     is <E>perfect</E> if every vertex of the digraph is incident to an edge in
     the matching.  Every maximum or perfect matching is maximal. Note however
     that for digraphs with loops, not every perfect matching will be maximum.
 
     <Example><![CDATA[
-gap> D := Digraph([[2], [1], [2, 3, 4], [3, 5], [1]]);
+gap> D := Digraph([[1, 2], [1, 2], [2, 3, 4], [3, 5], [1]]);
 <immutable digraph with 5 vertices, 8 edges>
 gap> IsMatching(D, [[2, 1], [3, 2]]);
 false
@@ -1680,13 +1680,18 @@ gap> IsMatching(D, edges);
 true
 gap> IsMaximalMatching(D, edges);
 false
-gap> edges := [[5, 1], [3, 3]];;
+gap> edges := [[2, 1], [3, 4]];;
 gap> IsMaximalMatching(D, edges);
 true
 gap> IsPerfectMatching(D, edges);
 false
 gap> edges := [[1, 2], [3, 3], [4, 5]];;
 gap> IsPerfectMatching(D, edges);
+true
+gap> IsMaximumMatching(D, edges);
+false
+gap> edges := [[1, 1], [2, 2], [3, 3], [4, 5]];;
+gap> IsMaximumMatching(D, edges);
 true
 ]]></Example>
   </Description>
--- a/doc/oper.xml
+++ b/doc/oper.xml
@@ -1653,10 +1653,12 @@ true
   <Description>
     If <A>digraph</A> is a digraph and <A>list</A> is a list of pairs of
     vertices of <A>digraph</A>, then <C>IsMatching</C> returns <K>true</K> if
-    <A>list</A> is a matching of <A>digraph</A>.  The operations
-    <C>IsMaximalMatching</C>, <C>IsMaximumMatching</C> and <C>IsPerfectMatching</C> 
-    return <K>true</K> if <A>list</A> is a maximal, maximum or perfect, matching 
-    of <A>digraph</A>, respectively.  Otherwise, these operations return <K>false</K>.
+    <A>list</A> is a matching of <A>digraph</A>.  The operation
+    <C>IsMaximalMatching</C> returns <K>true</K> if <A>list</A> is a maximal 
+    matching, <C>IsMaximumMatching</C> returns <K>true</K> if <A>list</A> is a
+    maximum matching and <C>IsPerfectMatching</C> returns <K>true</K> if 
+    <A>list</A> is a perfect, matching of <A>digraph</A>, respectively.  
+    Otherwise, each of these operations return <K>false</K>.
     <P/>
 
     A <E>matching</E> <C>M</C> of a digraph <A>digraph</A> is a subset of the
@@ -1667,8 +1669,8 @@ true
     <E>maximal</E> if it is contained in no larger matching of the digraph,
     is <E>maximum</E> if it has the greatest cardinality among all matchings and
     is <E>perfect</E> if every vertex of the digraph is incident to an edge in
-    the matching.  Every maximum or perfect matching is maximal. Note however
-    that for digraphs with loops, not every perfect matching will be maximum.
+    the matching.  Every maximum or perfect matching is maximal. Note, however,
+    that that not every perfect matching of digraphs with loops is maximum.
 
     <Example><![CDATA[
 gap> D := Digraph([[1, 2], [1, 2], [2, 3, 4], [3, 5], [1]]);
--- a/doc/oper.xml
+++ b/doc/oper.xml
@@ -1672,7 +1672,7 @@ true
 
     <Example><![CDATA[
 gap> D := Digraph([[1, 2], [1, 2], [2, 3, 4], [3, 5], [1]]);
-<immutable digraph with 5 vertices, 8 edges>
+<immutable digraph with 5 vertices, 10 edges>
 gap> IsMatching(D, [[2, 1], [3, 2]]);
 false
 gap> edges := [[3, 2]];;
--- a/doc/oper.xml
+++ b/doc/oper.xml
@@ -1647,24 +1647,28 @@ true
 <ManSection>
   <Oper Name="IsMatching" Arg="digraph, list"/>
   <Oper Name="IsMaximalMatching" Arg="digraph, list"/>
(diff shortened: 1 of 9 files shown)


doc/oper.xml
```diff
@@ -1647,43 +1647,52 @@ true
...
   <Returns><K>true</K> or <K>false</K>.</Returns>
   <Description>
     If <A>digraph</A> is a digraph and <A>list</A> is a list of pairs of
     vertices of <A>digraph</A>, then <C>IsMatching</C> returns <K>true</K> if
     <A>list</A> is a matching of <A>digraph</A>.  The operations
-    <C>IsMaximalMatching</C> and <C>IsPerfectMatching</C> return <K>true</K> if
-    <A>list</A> is a maximal, or perfect, matching of <A>digraph</A>,
-    respectively.  Otherwise, these operations return <K>false</K>. <P/>
+    <C>IsMaximalMatching</C>, <C>IsMaximumMatching</C> and <C>IsPerfectMatching</C> 
+    return <K>true</K> if <A>list</A> is a maximal, maximum or perfect, matching
```
flsmith: commata overload

doc/oper.xml
```diff
@@ -1647,43 +1647,52 @@ true
...
     edges of <A>digraph</A>, i.e. <C>DigraphEdges(<A>digraph</A>)</C>, such
     that no pair of distinct edges in <C>M</C> are incident to the same vertex
     of <A>digraph</A>.  Note that this definition allows a matching to contain
     loops.  See <Ref Prop="DigraphHasLoops" />.  The matching <C>M</C> is
-    <E>maximal</E> if it is contained in no larger matching of the digraph, and
+    <E>maximal</E> if it is contained in no larger matching of the digraph,
+    is <E>maximum</E> if it has the greatest cardinality among all matchings and
     is <E>perfect</E> if every vertex of the digraph is incident to an edge in
-    the matching.  Every perfect matching is maximal.
+    the matching.  Every maximum or perfect matching is maximal. Note however
```
flsmith: Maybe
"Note, however, that for digraphs with loops, not every perfect matching will be maximum."
or
"Note, however, that not every perfect matching of digraphs with loops is maximum."

flsmith: Also this is just a bit upsetting in general, but if that's how it's already defined then not sure what can be done about it

gap/attr.gi
```diff
@@ -2044,3 +2044,230 @@ InstallMethod(DigraphMycielskian,
...
+              return true;
+            fi;
+          fi;
+        fi;
+      fi;
+    od;
+    return false;
+  end;
+
+  time := List(DigraphVertices(D), x -> 0);
```
flsmith: `ListWithIdenticalEntries`?

gap/attr.gi
```diff
@@ -2044,3 +2044,230 @@ InstallMethod(DigraphMycielskian,
...
+  return mate;
+end);
+
+BindGlobal("DIGRAPHS_MateToMatching",
+function(D, mate)
+  local u, M;
+  M := [];
+  for u in DigraphVertices(D) do
+    if u <= mate[u] then
+      if IsDigraphEdge(D, u, mate[u]) then Add(M, [u, mate[u]]);
```
flsmith: Usually we would break these lines

gap/attr.gi
```diff
@@ -2044,3 +2044,230 @@ InstallMethod(DigraphMycielskian,
...
+      elif IsDigraphEdge(D, mate[u], u) then Add(M, [mate[u], u]);
+      fi;
+    fi;
+  od;
+  return Unique(M);
```
flsmith: `Set` is more efficient than `Unique`

tst/standard/attr.tst
```diff
@@ -2210,6 +2210,42 @@ gap> P := DigraphRemoveAllMultipleEdges(ReducedDigraph(OnDigraphs(D, proj[2])));
...
+<immutable digraph with 8 vertices, 10 edges>
+gap> DigraphMaximumMatching(D);
+[ [ 1, 5 ], [ 2, 6 ], [ 3, 7 ], [ 4, 8 ] ]
+gap> D := Digraph(IsMutable, [[2, 3], [1, 4], [2, 4], [5], [3, 5]]);
+<mutable digraph with 5 vertices, 9 edges>
+gap> M := DigraphMaximumMatching(D);; IsMaximalMatching(D, M);
+true
+gap> Length(M);
+3
+
```
flsmith: Maybe a few more tests for both of these? Simply because the algorithm is mildly complicated and I didn't try to verify that it was doing the right thing, so some extra attempts at breaking it would be welcome.

doc/oper.xml
reiniscirpons: Is the new formulation any better?

tst/standard/attr.tst
reiniscirpons: I added more tests to maximum matching - some large graphs with known unique maximum matchings and other large graphs with known maximum matching sizes. I didn't add more tests for maximal matching though, because we already test it against a random graph of size 100, which seems hard enough to just pass by chance.

tst/standard/attr.tst
```diff
@@ -2210,6 +2210,217 @@ gap> P := DigraphRemoveAllMultipleEdges(ReducedDigraph(OnDigraphs(D, proj[2])));
 gap> IsIsomorphicDigraph(H, P);
 true
 
+# DigraphMaximalMatching
+gap> D := RandomDigraph(100);;
+gap> M := DigraphMaximalMatching(D);; IsMaximalMatching(D, M);
```
james-d-mitchell: It's probably better to create the graph at random on your computer, and then use its string representation explicitly here. Tests with randomness like this are often a bad idea.

james-d-mitchell:
Thanks @reiniscirpons !

PR #298, opened 2020-03-09
Try to fix travis

james-d-mitchell:


james-d-mitchell:
Seems like maybe there's nothing to fix after all...

Issue #297, opened 2020-03-06
Digraphs not compiling in Centos 7


enriqueartal:
In Centos 7, gcc 485, Digraphs fail to compile with this error:

~~~~
gmake[1]: Entering directory `/usr/local/gap/pkg/digraphs-1.1.1'
  CC       src/digraphs_la-digraphs.lo
src/digraphs.c: In function ‘FuncOutNeighboursFromSourceRange’:
src/digraphs.c:102:3: error: ‘for’ loop initial declarations are only allowed in C99 mode
for (Int i = 1; i <= n; i++) {
^
src/digraphs.c:102:3: note: use option -std=c99 or -std=gnu99 to compile your code
src/digraphs.c:109:12: error: redefinition of ‘i’
for (Int i = 1; i <= LEN_LIST(src); i++) {
^
src/digraphs.c:102:12: note: previous definition of ‘i’ was here
for (Int i = 1; i <= n; i++) {
^
src/digraphs.c:109:3: error: ‘for’ loop initial declarations are only allowed in C99 mode
for (Int i = 1; i <= LEN_LIST(src); i++) {
^
src/digraphs.c: In function ‘DigraphNrEdges’:
src/digraphs.c:126:5: error: ‘for’ loop initial declarations are only allowed in C99 mode
for (Int i = 1; i <= n; i++) {
^
src/digraphs.c: In function ‘FuncDIGRAPH_TRANS_REDUCTION’:
src/digraphs.c:502:3: error: ‘for’ loop initial declarations are only allowed in C99 mode
for (UInt i = 1; i <= n; i++) {
^
src/digraphs.c:515:13: error: redefinition of ‘i’
for (UInt i = 1; i <= n; i++) {
^
src/digraphs.c:502:13: note: previous definition of ‘i’ was here
for (UInt i = 1; i <= n; i++) {
^
src/digraphs.c:515:3: error: ‘for’ loop initial declarations are only allowed in C99 mode
for (UInt i = 1; i <= n; i++) {
^
src/digraphs.c:573:11: error: ‘for’ loop initial declarations are only allowed in C99 mode
for (UInt m = 0; m < n; m++) {
^
src/digraphs.c: In function ‘buildBlissDigraph’:
src/digraphs.c:1613:7: error: ‘for’ loop initial declarations are only allowed in C99 mode
for (Int l = 1; l <= len; l++) {
^
src/digraphs.c: In function ‘MultiplyList’:
[...]

PR #299, opened 2020-03-16
build: check for c99

james-d-mitchell:


james-d-mitchell:
I think this resolves Issue #297

PR #304, opened 2020-04-08
docker: remove hard-wired GAP version

james-d-mitchell:


james-d-mitchell:
@alex-konovalov if you'd be kind enough to double check that this change makes sense, I'd be very grateful.

alex-konovalov:
@james-d-mitchell where this file is used - to demonstrate the package on Binder?

james-d-mitchell:
hi @alex-konovalov yes that's the purpose of the file

PR #307, opened 2020-04-15
IsDigraphFolding

le27:
A folding from a multidrgraph to another is a pair of maps (one between the vertex sets and one between the edges sets) which is a digraph homomorphism and with the further property that it induces a bijection between the spaces of bi-infinite paths through the digraphs.

The function takes as input two digraphs and two transformations and checks if the given transformations define a folding.

### DIFF
--- a/gap/grahom.gi
+++ b/gap/grahom.gi
@@ -652,6 +652,176 @@ function(src, ran, x, cols1, cols2)
     and DigraphsRespectsColouring(src, ran, x, cols1, cols2);
 end);
 
+InstallMethod(IsDigraphFolding, "for a digraph, a digraph, and a transformation",
+[IsDigraph, IsDigraph, IsTransformation, IsTransformation],
+function(domain, codomain, vertexmap, edgemap)
+  local temp, badedgepairs, optiontree, newoptions, currentpath,
+        goodconfigurations, newoptionstemp, newoption, codomedges,
+        domedges, domedgesbyinvertex, domedgesbyoutvertex, edgepath,
+        configurationstocheck, currentconfiguration, i, goodedgepairs,
+        newpair, codomedgesbyinvertex, codomedgesbyoutvertex,
+        newconfiguration, currentpair, failedpairs, pathcoverededges,
+        coverededges, loopable, forwardcoverededges;
+
+#checking functions have the correct domain and codomain
+  for i in [DigraphNrVertices(domain) + 1 .. DegreeOfTransformation(vertexmap)] do
+    if not i^vertexmap = i then
+      return false;
+    fi;
+  od;
+  for i in [DigraphNrEdges(domain) + 1 .. DegreeOfTransformation(edgemap)] do
+    if not i^edgemap = i then
+      return false;
+    fi;
+  od;
+  if not IsSubset([1 .. DigraphNrVertices(codomain)],
+                  ImageSetOfTransformation(vertexmap, DigraphNrVertices(domain))) then
+    return false;
+  fi;
+  if not IsSubset([1 .. DigraphNrEdges(codomain)],
+                  ImageSetOfTransformation(edgemap, DigraphNrEdges(domain))) then
+    return false;
+  fi;
+
+#checking that the function is a multidigraph homomorphism
+  for i in [1 .. DigraphNrEdges(domain)] do
+    temp := DigraphEdges(domain)[i];
+    if not DigraphEdges(codomain)[i^edgemap] = [temp[1]^vertexmap, temp[2]^vertexmap] then
+      return false;
+    fi;
+  od;
+
+  domedges := DigraphEdges(domain);
+  codomedges := DigraphEdges(codomain);
+
+  domedgesbyinvertex := List(DigraphVertices(domain),
+                             x -> Filtered([1 .. Size(domedges)],
+                                           y -> domedges[y][1] = x));
+  codomedgesbyinvertex := List(DigraphVertices(codomain),
+                               x -> Filtered([1 .. Size(codomedges)],
+                                             y -> codomedges[y][1] = x));
+  domedgesbyoutvertex := List(DigraphVertices(domain),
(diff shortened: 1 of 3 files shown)


james-d-mitchell:
@le27 thanks for the contribution! It seems like linting failed (should be easy to fix) and that the tests you've written don't cover enough of the lines in the code. On another less technical note, what algorithm are you implementing? I think you came up with this yourself, right? Can you do a short write up of the algorithm, and append it to this PR?

le27:
While I came up with the algorithm, I should mention that it is similar to the algorithm collin came up with, for closely related problem of checking bijectivity of a transducer. Here's a writeup of the algorithm:
[Folding_Algorithm.pdf](https://github.com/gap-packages/Digraphs/files/4486722/Folding_Algorithm.pdf)

gap/grahom.gi
```diff
@@ -652,6 +652,200 @@ function(src, ran, x, cols1, cols2)
...
+  #checking that the function is a multidigraph homomorphism 
+  for i in [1 .. DigraphNrEdges(domain)] do
+    temp := DigraphEdges(domain)[i];
+    if not DigraphEdges(codomain)[i ^ edgemap] =
+           [temp[1] ^ vertexmap, temp[2] ^ vertexmap] then
+      return false;
+    fi;
+  od;
+
+  domedges := DigraphEdges(domain);
```
flsmith, line 696: Could define these above and use them in the homomorphism check.

gap/grahom.gi
```diff
@@ -652,6 +652,200 @@ function(src, ran, x, cols1, cols2)
...
+                                                         codomedges[x][2],
+                                                         y)));
+  coverededges := Filtered(forwardcoverededges,
+                  x -> ForAny(loopable, y -> IsReachable(codomain, y,
+                                                         codomedges[x][1])));
+
+  goodconfigurations := [];
+  configurationstocheck := List(coverededges,
+                                x -> [x, Filtered([1 .. Size(domedges)],
+                                                   y -> y ^ edgemap = x)]);
```
flsmith, line 804: I think I may be misunderstanding something here, but in the pdf you say 

> We start by checking which edges of the domain are involved in biinfinite paths as these edges can
> be ignored throughout

and I think these correspond to the `coverededges`, but here the `coverededges` come from the codomain, and they are not ignored?

gap/grahom.gi
```diff
@@ -652,6 +652,200 @@ function(src, ran, x, cols1, cols2)
...
+  while not configurationstocheck = [] do
```
flsmith, line 805: again, should this just be a loop through `configurationstocheck`?

gap/grahom.gi
```diff
@@ -652,6 +652,200 @@ function(src, ran, x, cols1, cols2)
...
+      AddSet(goodconfigurations, StructuralCopy(currentconfiguration));
+      newoptionstemp := Set(List(currentconfiguration[2],
+                                 x -> domedges[x][2]));
+      newoptionstemp := Union(List(newoptionstemp,
+                                   x -> domedgesbyinvertex[x]));
+      newoptions := List(Filtered(codomedgesbyinvertex[codomedges
+                         [currentconfiguration[1]][2]],
+                         z -> z in forwardcoverededges),
+                         x -> [x, Filtered(newoptionstemp,
+                                           y -> y ^ edgemap = x)]);
```
flsmith, line 821: this is intensely confusing to read

flsmith: Why not use `Intersection` instead of `Filtered`?

gap/grahom.gi
```diff
@@ -652,6 +652,200 @@ function(src, ran, x, cols1, cols2)
...
+      newoptions := Filtered(newoptions, x -> not x in goodconfigurations);
```
flsmith, line 822: `Difference`

gap/grahom.gi
```diff
@@ -652,6 +652,200 @@ function(src, ran, x, cols1, cols2)
...
+      #finding an unwritable path 
+      if ForAny(newoptions, x-> x[2] = []) then
+        return [false, "sur"];
```
flsmith, line 825: Was this for debugging? It seems odd.

gap/grahom.gi
```diff
@@ -652,6 +652,200 @@ function(src, ran, x, cols1, cols2)
...
+        currentpath[Size(currentpath)] := currentpath[Size(currentpath)] + 1;
+        while currentpath <> [] and currentpath[Size(currentpath)] >
+                                    Size(optiontree[Size(currentpath)]) do
+          Remove(currentpath);
+          Remove(optiontree);
+          if Size(currentpath) > 0 then
+            currentpath[Size(currentpath)] := currentpath[Size(currentpath)]
+                                              + 1;
+          fi;
+        od;
```
flsmith, line 838: same comment as above for backtracking

gap/grahom.gi
```diff
@@ -652,6 +652,200 @@ function(src, ran, x, cols1, cols2)
...
+
+  domedgesbyinvertex := List(DigraphVertices(domain),
+                             x -> Filtered([1 .. Size(domedges)],
+                                           y -> domedges[y][1] = x));
+  codomedgesbyinvertex := List(DigraphVertices(codomain),
+                               x -> Filtered([1 .. Size(codomedges)],
+                                             y -> codomedges[y][1] = x));
+  domedgesbyoutvertex := List(DigraphVertices(domain),
+                              x -> Filtered([1 .. Size(domedges)],
+                                            y -> domedges[y][2] = x));
```
flsmith, line 707: It would be more efficient to loop through the edges than filter them, although this probably won't make much difference to the overall efficiency.

wilfwilson:
I get the impression that this was accidentally merged... is that the case @le27 @james-d-mitchell? (Note that it causes more Travis failures than were also existing). If so, shall we manually revert the commits, and get Luke to open a new PR, in which he can hopefully address @flsmith's comments?

PR #308, opened 2020-04-19
Change #include "src/compiled.h" to just "compiled.h"

fingolfin:
This will allow future GAP versions to simplify their GAP_CPPFLAGS,
which currently need to contain `-I$(abs_srcdir)` just to support this
old use case. Also, this in turn makes it easier to have package
compatible with future installed version of GAP, which have headers in
PREFIX/include/gap/ so there is no `src` there (unless downstream
packagers jump through hoops to ensure that, e.g. by adding a special
symlink).

Note: we could also #include "gap_all.h" but that would require GAP >= 4.11.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -47,7 +47,7 @@ LT_INIT([disable-static dlopen win32-dll])
 dnl ##
 dnl ## Locate the GAP root dir
 dnl ##
-AC_FIND_GAP
+FIND_GAP
 
 dnl ##
 dnl ## Detect Windows resp. Cygwin
--- a/m4/find_gap.m4
+++ b/m4/find_gap.m4
@@ -3,7 +3,7 @@
 # Can be configured using --with-gaproot=...
 #######################################################################
 
-AC_DEFUN([AC_FIND_GAP],
+AC_DEFUN([FIND_GAP],
 [
   AC_LANG_PUSH([C])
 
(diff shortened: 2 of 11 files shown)


james-d-mitchell:
Thanks @fingolfin, in principle I'm happy to accept this, except that it breaks backwards compatibility (as can be seen in the failing travis-ci tests). If this could be updated to remain compatible with older versions of GAP, then I'd be happy to merge.

fingolfin:
Yeah I noticed the failures, that's of course bad. I hope I can fix it by adding to the CPPFLAGS; but I need to work on teaching related stuff first. I'll try to update this PR (and similar ones elsewhere) later this week.

wilfwilson:
For the record, adding [the corresponding commit to NautyTracesInterface](https://github.com/gap-packages/NautyTracesInterface/commit/3dbd4fbe1fc0d699d0442ef670b3884f46bfba8e) broke its own compatibility 4.9, which is why all of our Travis tests are now failing their `stable-4.9` jobs, even without this PR.

fingolfin:
Once https://github.com/gap-packages/NautyTracesInterface/pull/35 is merged, also commit 7ce857744a13f8e01572c7a649bc576c685f9565 can be reverted

wilfwilson:
Great :)

PR #309, opened 2020-04-28
CHANGELOG: mention that datastructures is now required

wilfwilson:
I always think it's nice to mention such things.


james-d-mitchell:
Thanks @wilfwilson happy to merge, there are probably some other places that this should be mentioned too. Such as in Chapter 1, and in the README, that's my mistake! I'll make a PR now, unless you'd like to add it to this one?

wilfwilson:
Please add that in a separate PR 🙂

PR #313, opened 2020-04-29
C implementation of the Bron Kerbosch 

jjonusas:
Translated the gap implementation of Bron Kerbosch to C.


src/cliques.c
```diff
@@ -0,0 +1,628 @@
...
+
+static Obj clique_hook_collect(void* user_param,
+                               const BitArray* clique,
+                               const uint16_t nr) {
+  UInt   i;
+  Obj    c;
+
+  if (TNUM_OBJ((Obj) user_param) == T_PLIST_EMPTY) {
+    RetypeBag(user_param, T_PLIST);
+  }
```
fingolfin, line 95: This shouldn't be necessary (and I would recommend avoiding it!) as you use `ASS_LIST` below (good!)

src/cliques.c
```diff
@@ -0,0 +1,628 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+static Obj GAP_FUNC;       // Variable to hold a GAP level hook function
+
+static Obj (*HOOK)(void*,  // HOOK function applied to every homo found
+                   const BitArray*,
+                   const uint16_t);
+static void* USER_PARAM;   // A USER_PARAM for the hook
+
+static jmp_buf OUTOFHERE;  // So we can jump out of the deepest
```
fingolfin, line 72: Just to remark: this code won't work in a multi threaded setting due to all these global variables :-(.

The usual solution I'd use would be to put these into a simple struct, and then pass around a pointer to that struct to all functions that need it. At the top level, a member of that struct can be instantiated on the stack.

This can actually even be beneficial in terms of performance, as on many architectures, accessing global variables is slower than accessing a plain pointer.

src/cliques.c
```diff
@@ -0,0 +1,628 @@
...
+  uint16_t nr = GRAPH->nr_vertices;
+  BitArray* try = get_conditions(TRY, 0);
+  BitArray* ban = get_conditions(BAN, 0);
+
+  if (depth > 0 && !max && ( size == 0 || size == depth)) {
+    // We are not looking for maximal cliques
+    HOOK(USER_PARAM, CLIQUE, nr);
+    *nr_found += 1;
+    if (*nr_found >= limit) {
+      longjmp(OUTOFHERE, 1);
```
fingolfin, line 294: longjmp & setjmp are tricky beasts. I'd recommend to avoid them if possible. And here it really is possible, and quite easy: just change `BronKerbosch` to return an int; e.g. `0` for success, `1` for failure (or the other way around, doesn't matter). Then instead of doing a longjmp here and below, just `return FAILURE`; change all other `return;` to `return SUCCESS;`. And finally, in the recursive calls, just do `if (FAILURE == BronKerbosch(...)) return FAILURE;`

src/cliques.c
```diff
@@ -0,0 +1,628 @@
...
+    for (Int i = 1; i <= LEN_LIST(exclude_obj); ++i) {
+      if (!ISB_LIST(exclude_obj, i)) {
+        ErrorQuit("the 6th argument <exclude> must be a dense list,", 0L, 0L);
+      } else if (!IS_POS_INT(ELM_LIST(exclude_obj, i))) {
+        ErrorQuit("the 6th argument <exclude> must only contain positive "
+                  "integers, but found %s in position %d,",
+                  (Int) TNAM_OBJ(ELM_LIST(exclude_obj, i)),
+                  i);
+      } else if (INT_INTOBJ(ELM_LIST(exclude_obj, i)) 
+                 > DigraphNrVertices(digraph_obj)) {
```
fingolfin, line 496: This is not safe: `IS_POS_INT` will return true also for values like `2^1000` which are not intobj; and then the above command can either accidentally accept an invalid value (although that's unlikely), or crash (if in debug mode).

You could either use `IS_POS_INTOBJ` above (but then also adjust the error message, e.g. to say `... positive small integers` (the term "small integers" is defined precisely in the GAP manual. Well, at least since GAP 4.11 it is... ). Or else adjust the check here, e.g. by inserting an `IS_INTOBJ` or `IS_POS_INTOBJ` call before the `INT_INTOBJ`.

src/cliques.c
```diff
@@ -0,0 +1,628 @@
...
+    ErrorQuit("the 7th argument <max> must true or false, not %s,",
+              (Int) TNAM_OBJ(max_obj),
+              0L);
+  }
+  if (!IS_INTOBJ(size_obj) && size_obj != Fail) {
+    ErrorQuit("the 8th argument <size> must be an integer "
+              "or fail, not %s,",
+              (Int) TNAM_OBJ(size_obj),
+              0L);
+  } else if (IS_INTOBJ(size_obj) && INT_INTOBJ(size_obj) <= 0) {
```
fingolfin, line 518: Just use `IS_POS_INTOBJ` -- or even better, combined this check with the previous one, and say `must be a positive small integer` in the error message?

src/cliques.c
```diff
@@ -0,0 +1,628 @@
...
+    if (CALL_1ARGS(IsPermGroup, aut_grp_obj) != True) {
+      ErrorQuit(
+          "the 9th argument <aut_grp> must be a permutation group "
+          "or fail, not %s,",
+          (Int) TNAM_OBJ(aut_grp_obj),
+          0L);
+    }
+    Obj gens = CALL_1ARGS(GeneratorsOfGroup, aut_grp_obj);
+    DIGRAPHS_ASSERT(IS_LIST(gens));
+    DIGRAPHS_ASSERT(LEN_LIST(gens) > 0);
```
fingolfin, line 535: So I realize this is unlikely to ever matter in practice, but I wonder: do you really want to crash if `gens` is bad data? Why not `ErrorQuit` here like for everything else, too? Or for that matter, simply remove the assertions, the first call to `LEN_LIST` will trigger an `Error` of its own if `gens` is not a list.

src/cliques.c
james-d-mitchell: Thanks for the comments @fingolfin, much of the kernel module of Digraphs is not thread-safe, an issue that I'm aware of, but haven't really seen any reason to change until now.

fingolfin: In practice I also don't care that much about thread safety, it is just usually the quickest way to convince people of my true desire: to avoid global variables as much as possible. They tend to cause all kinds of design issues down the road, and often make it much harder to reason about code that uses them. But here at least they are static and restricted to one relatively small source file.

Anyway, all that is just my opinion, and obviously that shouldn't count for much here, as I am not a stakeholder in this project :-). I merely stumbled over this PR and left some well-meant quick comments. Mind you, overall this code looks perfectly fine

jjonusas: Thanks @fingolfin, the comments are much appreciated!

flsmith:
I noticed while valgrinding this that you don't free the memory allocated by `new_bit_array`. 
There's also an assertion failure in debug mode:
``# line 474 of 629 (75%)gap: src/conditions.h:85: clear_conditions: Assertion `nr2 != 0` failed.``
when `CliqueNumber` is called on `NullDigraph(0)`.

Once these are resolved this looks good to merge.

flsmith:
I've pushed fixes for the assertion/memory leaks to [flsmith:cliques](https://github.com/flsmith/Digraphs/tree/cliques).

jjonusas:
@flsmith thanks for the fixes!

flsmith:
Merged, thanks @jjonusas and sorry for the delay!

PR #319, opened 2020-05-13
Added Dijkstra version for digraph with edges weights

mariatsalakou:


gap/oper.gi
```diff
@@ -1365,6 +1365,37 @@ InstallMethod(DigraphDijkstra, "for a digraph, and a vertex",
 [IsDigraph, IsPosInt],
 {digraph, source} -> DIGRAPHS_DijkstraST(digraph, source, fail));
 
+InstallMethod(DigraphDijkstraSTWeights, "for a digraph, a vertex, a vertex, and a list of weights",
```
james-d-mitchell, line 1368: I suggest using your new version `DigraphDijkstraSTWeights` as the code of `DIGRAPHS_DijkstraST`, with one change: the argument `weights` could be a function like:
~~~~
weights := function(u, v)
  return list_of_weights[u][v];
end;
~~~~
if a list of weights is given, and if no list of weights is given then we could do:
~~~~
weights := function(u, v)
  return 1;
end;
~~~~

gap/oper.gi
```diff
@@ -1365,6 +1365,37 @@ InstallMethod(DigraphDijkstra, "for a digraph, and a vertex",
...
+    od;
+
+    dist[source] := 0;
+    Push(queue, [0, source]);
+
+    while not IsEmpty(queue) do
+        u := Pop(queue);
+        u := u[2];
+        for v in OutNeighbours(digraph)[u] do
+            alt := dist[u] + weights[u][v];
```
james-d-mitchell, line 1389: would become `weights(u, v)` instead of `weights[u][v]`

gap/oper.gi
```diff
@@ -1365,6 +1365,37 @@ InstallMethod(DigraphDijkstra, "for a digraph, and a vertex",
 [IsDigraph, IsPosInt],
 {digraph, source} -> DIGRAPHS_DijkstraST(digraph, source, fail));
```
james-d-mitchell, line 1366:
~~~~
{digraph, source} -> DIGRAPHS_DijkstraST(digraph, source, fail))
~~~~
would become
~~~~
{digraph, source} -> DIGRAPHS_DijkstraST(digraph, source, fail, {u, v} -> 1))
~~~~

gap/oper.gi
```diff
@@ -1365,6 +1365,37 @@ InstallMethod(DigraphDijkstra, "for a digraph, and a vertex",
...
 
+InstallMethod(DigraphDijkstraSTWeights, "for a digraph, a vertex, a vertex, and a list of weights",
+[IsDigraph, IsPosInt, IsPosInt, IsList],
+function(digraph, source, target, weights)
+    local dist, prev, queue, u, v, alt;
```
james-d-mitchell, line 1371: Check that the `weights` have the same "shape" as `OutNeighbours(digraph)`

PR #324, opened 2020-06-06
Digraph orientation

james-d-mitchell:


Peter-Ing:
This Digraph on nodes 1,2 
[ [ 1, 1 ], [ 1, 2 ], [ 1, 2 ], [ 2, 2 ] ]

james-d-mitchell:
This requires some work, @Peter-Ing has pointed out some cases where it doesn't work as intended, and at the very least the manual entry is not correct.

james-d-mitchell:
I'm going to close this for the time being until I figure out what is wrong, and how to rectify it.

PR #325, opened 2020-06-09
Introduce StrongOrientation, IsBridgelessDigraph, and Bridges

james-d-mitchell:
This PR replaces #324, and is now hopefully correct.


james-d-mitchell:
@peter-ing, can you also please check that this is good for your purposes?

Peter-Ing:
> @Peter-Ing, can you also please check that this is good for your purposes?
Looking through I believe this should behave as the previous version for my purposes and so is good.

james-d-mitchell:
Thanks for being willing to referee @wilfwilson !

gap/attr.gi
```diff
@@ -17,85 +17,152 @@ InstallMethod(OutNeighbours, "for a digraph by out-neighbours",
...
-      v     := stack[depth][1];
-      w     := nbs[v][stack[depth][2]];
-      if v <> 1 and low[w] >= num[v] and not points_seen[v] then
-        points_seen[v] := true;
-        Add(points, v);
+  # number of nodes encountered in the search so far
+  counter := 0;
+
+  # the order in which the nodes are visited, -1 indicates "not yet visited".
+  pre := ListWithIdenticalEntries(N, -1);
```
MTWhyte, line 54: I was following the reference given on Line 18, and it refers to this ordering as 'Num'. Since broadly the variable names are in keeping with the reference, might it be better to call this list `num` instead of `pre` for ease of comparison?

doc/attr.xml
```diff
@@ -1068,6 +1067,95 @@ gap> ArticulationPoints(D);
...
+fail
+]]></Example>
+  </Description>
+</ManSection>
+<#/GAPDoc>
+
+<#GAPDoc Label="Bridges">
+<ManSection>
+  <Oper Name="Bridges" Arg="D"/>
+  <Attr Name="BridgesAttr" Arg="D"/>
```
wilfwilson, line 1116:
```suggestion
  <Attr Name="Bridges" Arg="D"/>
```

doc/attr.xml
```diff
@@ -1068,6 +1067,95 @@ gap> ArticulationPoints(D);
...
+<ManSection>
+  <Oper Name="StrongOrientation" Arg="D"/>
+  <Attr Name="StrongOrientationAttr" Arg="D"/>
+  <Returns>A digraph or <K>fail</K>.</Returns>
+  <Description>
+    A <E>strong orientation</E> of a connected symmetric digraph <A>D</A> (if
+    it exists) is a strongly connected subdigraph <C>C</C> of <A>D</A> such
+    that for every edge <C>[u, v]</C> of <A>D</A> either <C>[u, v]</C> or
+    <C>[v, u]</C> is an edge of <C>C</C> but not both.  By Robbin's Theorem,
+    only bridgeless digraphs (see <Ref Prop="IsBridgelessDigraph"/>) admit a
```
wilfwilson, line 1080: Is this an if and only if? If so, I think this should be said more explicitly.

gap/attr.gi
```diff
@@ -17,85 +17,152 @@ InstallMethod(OutNeighbours, "for a digraph by out-neighbours",
...
+function(D)
+  return DIGRAPHS_ArticulationPointsBridgesStrongOrientation(D)[3];
+end);
+
+InstallMethodThatReturnsDigraph(StrongOrientation,
+"for a digraph by out-neighbours",
+[IsDigraphByOutNeighboursRep],
+function(D)
+  if not IsSymmetricDigraph(D) then
+    ErrorNoReturn("not yet implemented");
```
wilfwilson, line 163: Is there any prospect of a notion of `StrongOrientation` for non-symmetric digraphs? This error message would suggest that there is a notion, we just haven't going round to programming it yet. Which might set the wrong expectation.

doc/attr.xml
```diff
@@ -1068,6 +1067,95 @@ gap> ArticulationPoints(D);
...
+    strong orientation. 
+    <P/>
+
+    This operation returns a strong orientation of the digraph <A>D</A> if 
+    <A>D</A> is symmetric and <A>D</A> admits a strong orientation. If <A>D</A>
+    is symmetric but does not admit a strong orientation, then <K>fail</K> is
+    returned. If <A>D</A> is not symmetric, then an error is given. 
+    <P/>
+
```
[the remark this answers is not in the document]
james-d-mitchell: Nice thanks will do!

gap/attr.gi
james-d-mitchell: Maybe I should update the reference, which only explains how to compute the articulation points, there's another reference for computing bridges (which involves adding 3 lines to the code for articulation points), there `pre` is used, I'll add a reference to this too in a comment.

doc/attr.xml
```diff
@@ -1068,6 +1067,95 @@ gap> ArticulationPoints(D);
...
+    the sense of <Ref Prop="IsConnectedDigraph"/>) when any edge is
+    removed.  If the digraph <A>D</A> is not 2-edge-connected but is
+    connected, then any edge <C>[u, v]</C> of <A>D</A> whose removal
+    makes the resulting digraph disconnected is called a <E>bridge</E>.<P/>
+
+    <C>Bridges</C> returns a list of the bridges of
+    <A>D</A>, if any, and, in particular, returns the empty list if
+    <A>D</A> is not connected. <P/>
+
+    Multiple edges and loops are ignored by this method. <P/>
```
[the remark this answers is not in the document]
james-d-mitchell: Right!

doc/attr.xml
```diff
@@ -1068,6 +1067,95 @@ gap> ArticulationPoints(D);
...
+gap> Bridges(D);
+[  ]
+gap> D := Digraph([[2], [3], [4], [2]]);
+<immutable digraph with 4 vertices, 4 edges>
+gap> Bridges(D);
+[ [ 1, 2 ] ]
+gap> Bridges(ChainDigraph(2));
+[ [ 1, 2 ] ]
+gap> ArticulationPoints(ChainDigraph(2));
+[  ]
```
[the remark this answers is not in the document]
james-d-mitchell: Good point, I've written some blurb about the connection between articulation points and bridges.

doc/attr.xml
james-d-mitchell: So, there are other instances in this file where both the oper and attr version are documented, so doing this seems to be consistent (even if maybe not desirable)

doc/attr.xml
james-d-mitchell: Yup it is, made explicit

doc/attr.xml
wilfwilson: I'm fine with that where oper and attr versions exist (as with `StrongOrientation`) - i.e. for 'attributes' that return digraphs. But `Bridges` doesn't have an oper and an attr version...

gap/attr.gi
james-d-mitchell: Yes there is, I just didn't have time to write the code for it. The notion certainly makes sense, for every pair of edges u -> v and v -> u  in a digraph, then we'd be asking to choose one of them so that the resulting digraph had no such pairs and was strongly connected.

james-d-mitchell:
@wilfwilson I think I've made all the changes requested (or in two cases, not, but provided a rationale for not doing it in those cases). Please let me know if I messed anything up!

james-d-mitchell:
@MTWhyte I've made the changes you indicated too, thanks for the review!

doc/attr.xml
james-d-mitchell: Doh, right you are...

james-d-mitchell: Fixed

PR #327, opened 2020-06-24
Add DigraphNrLoops

marinaanagno:


marinaanagno:
@reiniscirpons  @wilfwilson  Thanks for the reviews! Sam and I made the changes and we thought that adding a DigraphByAdjacencyMatrix example would be good as well, so that we are sure that it works with this method as well. If this is not needed and needs to be I can remove it as soon as possible. Also, @wilfwilson reading your review I understood that it would be good to keep both methods in attr.gi for DigraphNrLoops, and thought that if we have the matrix then we only need to iterate over n values and just sum them up which sounds fast. Also I saw that this is done for many attributes. However I'm not sure if you'd like us to completely remove the adjacency matrix one or keep it.

gap/attr.gi
```diff
@@ -346,6 +346,43 @@ function(D)
...
+[IsDigraphByOutNeighboursRep],
+function(D)
+  local A, sum, n, i;
+  A := AdjacencyMatrix(D);
+  sum := 0;
+  n := DigraphNrVertices(D);
+  for i in [1 .. n] do
+     sum := sum + A[i][i];
+  od;
+  return sum;
```
wilfwilson, line 383: You could shorten this code as follows:
```suggestion
  return Sum(DigraphVertices(D), i -> A[i][i]);
```

PR #330, opened 2020-07-13
Dot digraph colours

marinaanagno:
Added coloured options for vertices and edges in DotDigraph and DotSymmetricDigraph. We also used a global variable so that DotDigraph, DotVertexLabelledDigraph, DotColoredDigraph, DotSymmetricDigraph and DotSymmetricColoredDigraph do not have repeated code.


doc/z-chap9.xml
```diff
@@ -3,7 +3,9 @@
   <Section><Heading>Visualising a digraph</Heading>
     <#Include Label="Splash">
     <#Include Label="DotDigraph">
+    <#Include Label="DotColoredDigraph">
```
james-d-mitchell, line 6: This line isn't necessary, and is probably what is causing the CI to fail, I suggest removing it.

doc/z-chap9.xml
```diff
@@ -3,7 +3,9 @@
...
     <#Include Label="DotSymmetricDigraph">
+    <#Include Label="DotSymmetricColoredDigraph">
```
james-d-mitchell, line 8: Same for this one.

doc/display.xml
```diff
@@ -98,9 +99,40 @@ gap> Splash(DotDigraph(RandomDigraph(4)));
   </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="DIGRAPHS_DotDigraph">
```
james-d-mitchell, line 102: We don't usually document functions starting `DIGRAPHS_` these are treated as (kind of) private internal functions/methods.

doc/attr.xml
```diff
@@ -133,6 +133,30 @@ gap> DigraphNrEdges(D);
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="DigraphNrLoops">
```
james-d-mitchell, line 136: This chunk shouldn't be here, looks like you've branched `DotDigraphColours` off `DigraphNrLoops` but you should have done it from the `master` branch.

james-d-mitchell:
Thanks for the PR @marinaanagno ! Obviously the CI is failing, I think you need to make a few changes, as mentioned in the comments before having another review.

doc/display.xml
```diff
@@ -139,13 +277,19 @@ gap> FileString("dot/k4.dot", DotDigraph(gr));
...
+    <Oper Name="DotSymmetricColoredDigraph" Arg="digraph, vert, edge"/>
     <Returns>A string.</Returns>
     <Description>
       This function produces a graphical representation of the symmetric
       digraph <A>digraph</A>. <C>DotSymmetricDigraph</C> will return an
       error if <A>digraph</A> is not a symmetric digraph. See
       <Ref Prop="IsSymmetricDigraph"/>.<P/>
 
+      The function <C>DotSymmetricColoredDigraph</C> differs from <C>DotDigraph</C>
+      only in
```
james-d-mitchell, line 289: Please remove the line break here.

doc/display.xml
```diff
@@ -110,6 +111,10 @@ gap> Splash(DotDigraph(RandomDigraph(4)));
       vertices, with the arrowhead of each line pointing towards the range
       of the edge.<P/>
 
+      <C>DotColoredDigraph</C> differs from <C>DotDigraph</C> only in
+      that the values in given in the two lists are used to color the vertices and
+      edges of the graph when displayed.  <P/>
```
james-d-mitchell, line 116: I think you should add more details about what values the arguments can have. I.e. what are the possible "colors" of the vertices and edges? what if I want to specify vertex colors but not edge colors, or vice versa? the list of edge colors has to have the same "shape" as the list of `Outneighbours` right?

gap/display.gi
```diff
@@ -7,59 +7,75 @@
...
-  return str;
+InstallMethod(DotColoredDigraph, "for a digraph by out-neighbours and two lists",
+[IsDigraphByOutNeighboursRep, IsList, IsList],
+function(D, vert, edge)
+  return(DIGRAPHS_DotDigraph(D,
+                             [i -> Concatenation("[color=",
+                             vert[i], ", style=filled]")],
+                             [{i, j} -> Concatenation("[color=",
+                             edge[i][j], "]")]));
+  # return DIGRAPHS_DotDigraph(D,
```
james-d-mitchell, line 60: Add a comment that the method above can be replaced with the commented out one when/if Digraphs requires GAP version ??.

gap/display.gi
```diff
@@ -7,59 +7,75 @@
...
+  #                             edge[i][j])]);
 end);
 
-InstallMethod(DotSymmetricDigraph, "for a digraph by out-neighbours",
+InstallMethod(DotVertexLabelledDigraph, "for a digraph by out-neighbours",
 [IsDigraphByOutNeighboursRep],
 function(D)
-  local out, str, i, j;
+  return DIGRAPHS_DotDigraph(D, [i -> Concatenation(" [label=\"",
+                                String(DigraphVertexLabel(D, i)), "\"]")], []);
```
james-d-mitchell, line 71: Same as above, please add a comment.

gap/display.gi
```diff
@@ -68,19 +84,53 @@ function(D)
...
+[IsDigraphByOutNeighboursRep],
+function(D)
+  return DIGRAPHS_DotSymmetricDigraph(D, [], []);
+end);
+
+InstallMethod(DotSymmetricColoredDigraph,
+"for a digraph by out-neighbours and two lists",
+[IsDigraphByOutNeighboursRep, IsList, IsList],
+function(D, vert, edge)
+  return DIGRAPHS_DotSymmetricDigraph(D,
```
james-d-mitchell, line 122: What happens if the list of colors (edge or vertex) is not valid? I.e. if the list is too short? Or has holes?

gap/display.gi
```diff
@@ -68,19 +84,53 @@ function(D)
...
+                                     [i -> Concatenation("[color=", vert[i],
+                                     ", style=filled]")],
+                                     [{i, j} -> Concatenation("[color=",
+                                     edge[i][j], "]")]);
+  # return DIGRAPHS_DotSymmetricDigraph(D,
```
james-d-mitchell, line 127: Add a comment as above.

marinaanagno:
We made the suggested changes and also we created some new functions `DotVertexColoredDigraph` `DotEdgeColoredDigraph`, `DotSymmetricVertexColoredDigraph` and `DotSymmetricEdgeColoredDigraph` that colour edges and vertices independently.

james-d-mitchell:
@marinaanagno I'm not sure that you've pushed your changes, can you please double check? If you've addressed my comments, then they should be showing as "outdated", and at present they aren't.

james-d-mitchell (changes requested):
We discussed this just now, and the only outstanding thing to do is to check that the "color" arguments to these functions are valid.

james-d-mitchell (approved):

PR #332, opened 2020-08-11
Remove two calls to RetypeBag from kernel extension

fingolfin:
Motivation: RetypeBag is a very low-level part of the GAP kernel, and kernel
extensions should ideally restrict use of it to what's really required. In
many cases, it is far better to avoid it, and instead use higher-level APIs
such as ASS_LIST to edit lists, instead of relying on the unsafe low-level API
SET_ELM_PLIST combined with RetypeBag.

However, in the two cases at hand, calling RetypeBag simply is unnecessary: In
the first one, the code actually uses ASS_LIST and as such any retyping is
automatically performed by the GAP kernel for us.

In the second case, an empty plist is marked as such explicitly by retyping it
to T_PLIST_EMPTY. But that's a purely optional optimization.

### DIFF
--- a/src/homos.c
+++ b/src/homos.c
@@ -260,10 +260,6 @@ homo_hook_collect(void* user_param, uint16_t const nr, uint16_t const* map) {
   Obj    t;
   UInt   i;
 
-  if (TNUM_OBJ((Obj) user_param) == T_PLIST_EMPTY) {
-    RetypeBag(user_param, T_PLIST);
-  }
-
   // copy map into new trans2
   t   = NEW_TRANS2(nr);
   ptr = ADDR_TRANS2(t);
--- a/src/planar.c
+++ b/src/planar.c
@@ -205,9 +205,6 @@ Obj boyers_planarity_check(Obj digraph, int flags, bool krtwsk) {
         }
         j = theGraph->E[j].link[1];
       }
-      if (nr == 0) {
-        RetypeBag(list, T_PLIST_EMPTY);
-      }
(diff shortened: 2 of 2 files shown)


james-d-mitchell:
Thanks @fingolfin, I'm happy to merge this provided the CI passes. Appveyor seems to have failed because of not being able to download something in the set up, not related to this PR. It'd be good if the appveyor tests could be restarted, but I don't seem to be able to restart them, perhaps you can @fingolfin?

fingolfin:
Huh, AppVeyor tests are reported as passing over here? Although there are "gzip: stdout: Broken pipe" messages in the test output, are those "normal" ?

There are two failures on Travis, though, which look as if `Digraphs` is not being loaded? I'll try to figure that one out.


BTW I noticed on https://travis-ci.org/github/gap-packages/Digraphs/branches that you have "cron builds" active for the master and stable-1.2 branches, but not stable-1.3. Intentional?

fingolfin:
OK, now I see what you mean with download failures. I restarted those two jobs. Weird.

As to you not being able to restart jobs on Travis: that's also strange. I thought that all members of the `gap-packages` org should be able (as in: have permissions) to do that. Just to double check, are you logged in on Travis? I've had similar issues a few times in the past, usually it helped me force-reload the page a few times, or log out and log back in. If the issue persists, please let me know -- you really should have control over CI builds for this repository, after all!

james-d-mitchell:
Thanks @fingolfin I couldn't restart the jobs on appveyor not travis (I didn't try on travis), so I think nothing to worry about.

PR #334, opened 2020-08-19
Adds MaximalCommonSubdigraph and MinimalCommonSupergraph

le27:
The function MaximalCommonSubdigraph takes as input 2 digraphs and returns as output a digraph of maximum size which embeds into both of them as an induced subdigraph, as well embeddings of it into the given digraphs. The embeddings are given as transformations.

The function MinimalCommonSuperdigraph takes as input 2 digraphs and returns as output a digraph of minimum size which embeds into both into which they both embed as induced subdigraphs, as well embeddings of the given digraphs into it. The embeddings are again given as transformations.


le27:
don't mearge, something weird is happening with the commits

gap/grahom.gi
```diff
@@ -872,3 +872,104 @@ function(D, t)
...
+
+  B := InducedSubdigraph(B, Filtered(DigraphVertices(B), good));
+  if DigraphNrVertices(B) = 0 then
+    return [Digraph([]), [], []];
+  fi;
+
+  Clqus := DigraphMaximalCliquesReps(B);
+
+  if Clqus = [] then
+    return [Digraph([]), [], []];
```
tomcontileslie, line 920: Would there potentially be an interest in getting the second and third list elements to be transformations in this case? I'm not sure what you think of returning `IdentityTransformation` rather than `[]` if the maximal common subdigraph has no vertices; but that way we can guarantee that this function will always output a list which has transformations as second and third elements. For example,
```gap
gap> comm := MaximalCommonSubdigraph(D1, D2);;
gap> 2 ^ comm[2];
```
will never return any errors.

tst/standard/grahom.tst
```diff
@@ -2492,6 +2492,46 @@ gap> IsDigraphFolding(Digraph([[1, 2, 3], [1, 2], [1, 3]]), Digraph([[1, 1]]),
...
+[ <immutable empty digraph with 1 vertex>, IdentityTransformation, 
+  IdentityTransformation ]
+gap> MinimalCommonSuperdigraph(NullDigraph(10), CompleteDigraph(10));
+[ <immutable digraph with 19 vertices, 90 edges>, IdentityTransformation, 
+  Transformation( [ 1, 11, 12, 13, 14, 15, 16, 17, 18, 19, 11, 12, 13, 14, 15,
+     16, 17, 18, 19 ] ) ]
+gap> MaximalCommonSubdigraph(CompleteDigraph(100), CompleteDigraph(100));
+[ <immutable digraph with 100 vertices, 9900 edges>, IdentityTransformation, 
+  IdentityTransformation ]
+gap> MinimalCommonSuperdigraph(CompleteDigraph(100), CompleteDigraph(100));
```
tomcontileslie, line 2518: To be even more certain that these functions continue working in the future, perhaps there should be some tests that check not only equality of display, but also equality of objects, here? For example:
```gap
gap> comm := MinimalCommonSuperdigraph(CompleteDigraph(100), CompleteDigraph(100));;
gap> comm[1] = CompleteDigraph(100);
true
```
I'm not sure to what extent this is implemented in other test files, it may be overkill :)

tomcontileslie (commented):
I'm afraid I got lost trying to figure out the mathematical details of the implementation, but from what I can see this looks great, and it works really well! I have asked two questions in my review, apologies if they are a bit naive, I may have overlooked the reasons you chose to do things otherwise.

Also, I don't know whether you wanted to maybe add some documentation for these two new functions; looks like they'd fit quite well in either section 3.3, or maybe 7.2 or 7.3.

gap/grahom.gi
le27: I agree, I'll fix this this thanks.

tst/standard/grahom.tst
le27: good idea

gap/grahom.gi
```diff
@@ -872,3 +872,104 @@ function(D, t)
...
+  iso := IsomorphismDigraphs(D1, D2);
+
+  if iso <> fail then
+    return [DigraphCopy(D1), IdentityTransformation, AsTransformation(iso)];
+  fi;
+
+  out1 := OutNeighbours(D1);
+  out2 := OutNeighbours(D2);
+
+  adj := function(pair)
```
james-d-mitchell, line 891: I'm pretty sure that we've discussed this already, but I think you could improve the performance here, by:

1. Not using `Cartesian` but instead doing:
~~~~
L := List([1 .. Size(out1) * Size(out2)], x -> []);
for i in [1 .. Size(out1)] do
  for j in [1 .. Size(out2)] do
    if isadj(i, j) then
       Add(L[i], j); 
    fi;
  od;
od;
~~~~

Also the function `isadj` could be improved, by to run through the out neighbours of `D1` and `D2` once by using the adjacency matrix of `D1` and `D2`

james-d-mitchell (changes requested):
Improve the efficiency!

james-d-mitchell (changes requested):

james-d-mitchell:
Superseded by #361

PR #336, opened 2020-10-07
Add Tarjan and Lengauer's (almost) linear time dominators algorithm

james-d-mitchell:
~~This work in progress~~


james-d-mitchell:
@marinaanagno @samanthaharper @finnbuck

james-d-mitchell:
The following things still need to be done (hoops to be jumped through):

- remove the quadratic time version of `Dominators`. This could be put into the test file as a stand alone function, for testing purposes. 
- rename `Dominators2` to `Dominators` and update the documentation etc. Also ensure that `Dominators2` is a proper GAP method (like `Dominators` currently is).
- make `DominatorTree` a proper GAP operation (like `Dominators` currently is).
- add documentation for `DominatorTree`
- add tests for `DominatorTree`

Also we might consider:

- reenabling the `lastlinked` optimisation
- making `buckets` flat again.

james-d-mitchell:
There are still a couple of things to do in this PR, so no need for review just yet.

marinaanagno:
Thanks @james-d-mitchell! Let me know if there's anything I could do!

james-d-mitchell:
@marinaanagno when you have a moment, can you please review this PR? Once you've done that I'll squash the commits into one again (I left them separate for the time being so that you can see what I did today), and add you and Sam as co-authors for the commit.

doc/oper.xml
```diff
@@ -1572,6 +1572,92 @@ gap> DigraphShortestDistance(D, [1, 3], [3, 5]);
...
+  <Returns>A list of lists </Returns>
+    <Description>
+    <C>Dominators</C> is a function that takes a <A>digraph</A> and a
+    specific root <A>root</A> and returns the dominators of each vertex with
+    respect to the root. The root must be a verex of the digraph. The algorithm
+    is an implementation of the almost linear time algorithm by
+    Thomas Lengauer and Robert Endre Tarjan <Cite Key="LT79"/>. If there is
+    no path from the root to a specific vertex, the output will contain a
+    hole in the corresponding position. The complexity of this algorithm is 
+    <M>O(mlog n)</M> where <M>m</M> is the number of edges and <M>n</M> is the
```
marinaanagno, line 1587: My knowledge in complexity is almost zero right now, but since we visit all the nodes in the digraph but only the edges of the subdigraph induced by the reachable vertices, shouldn't we get that "n is the number of nodes and m is the number of edges in the digraph induced by the vertices reachable from the root vertex"?

marinaanagno, line 1587: Lacking knowledge in complexity, but I was thinking that `DominatorTree` performs a dfs so it will have worst case complexity O(m + n), but `Dominators` only runs through the vertices that are reachable from the root vertex and so I cannot find a case, in which we would visit any of the edges not belonging in the subdigraph induced by the reachable nodes. Definitely not sure about this though!

marinaanagno (commented):
The code seems to agree with the theory stated in the paper and seems to be giving right results. Since this is a project I've been involved though and since I am a newcomer a quick look from another pair of eyes would be great! My only comment is on the complexity, but again it is very possible that I might be missing something.

doc/oper.xml
marinaanagno, line 1587:
```suggestion
    <M>O(mlog n)</M> where <M>m</M>  and <M>n</M> are the
```

doc/oper.xml
```diff
@@ -1572,6 +1572,92 @@ gap> DigraphShortestDistance(D, [1, 3], [3, 5]);
...
+    number of nodes in the subdigraph induced by the nodes in <A>digraph</A>
```
marinaanagno, line 1588:
```suggestion
    number of edges and nodes in the subdigraph induced by the nodes in <A>digraph</A> respectively.
```

marinaanagno, line 1588:
```suggestion
    number of edges and number of nodes in the subdigraph induced by the nodes in <A>digraph</A>
```

doc/oper.xml
```diff
@@ -1572,6 +1572,92 @@ gap> DigraphShortestDistance(D, [1, 3], [3, 5]);
...
+    reachable from <A>root</A>.
```
marinaanagno, line 1589:
```suggestion
    reachable from <A>root</A>, respectively.
```

marinaanagno (changes requested):
Everything looks good, I only suggested a change in the phrasing in the complexity part of the documentation

james-d-mitchell:
Thanks @marinaanagno much appreciated. I wonder if the wording about the complexity in the manual entry for `DominatorTree` should be adjusted in the same way you suggested for `Dominators`?

marinaanagno:
@james-d-mitchell you're absolutely right! Sorry I didn't annotate that!

gap/oper.gi
```diff
@@ -1725,6 +1725,146 @@ function(D, list)
...
+
+  # Step 2: find semidominators, and first pass of immediate dominators
+  semi := [1 .. M];
+  lastlinked := M + 1;  # never linked
+  label := [];
+  bucket := List([1 .. M], x -> []);
+  idom := [];
+  idom[root] := root;
+
+  compress := function(v)
```
marinaanagno, line 1780: This function is recursive and I remember that we said that functions in a package shouldn't be recursive, since there is a limit on how many recursions GAP can "handle".

marinaanagno (commented):
Sorry for the delay, it occurred to me after I had a look. `Compress` is a recursive function and I'm afraid we've said that we have to make it non-recursive before merging.

james-d-mitchell:
> Sorry for the delay, it occurred to me after I had a look. `Compress` is a recursive function and I'm afraid we've said that we have to make it non-recursive before merging.

Good point! I’ll have to fix this, maybe not today though 😎

marinaanagno:
> Sorry for the delay, it occurred to me after I had a look. `Compress` is a recursive function and I'm afraid we've said that we have to make it non-recursive before merging.

After discussing this issue with @james-d-mitchell we figured out that compress will only run log2 of the number of vertices and no digraph that can be calculated using GAP will hit the recursion limit for compress

wilfwilson:
Something that reviewers should have in mind while reviewing: what should the return value be?

gap/oper.gi
```diff
@@ -1725,6 +1725,146 @@ function(D, list)
...
+  od;
+  for v in bucket[root] do
+    idom[v] := root;
+  od;
+
+  # Step 3: finalize immediate dominators
+  for i in [2 .. N] do
+    w := preorder_num_to_node[i];
+    if idom[w] <> semi[w] then
+      idom[w] := idom[semi[w]];
```
wilfwilson, line 1842: I noticed that the pseudocode in the paper instead has
```suggestion
      idom[w] := idom[idom[w]];
```
I just tried putting it like that and it doesn't work! So this is a bug in the paper, right?

wilfwilson (changes requested):
Thanks @marinaanagno and @samanthaharper for your work on this! I've looked at it for quite a while, including reading parts of the paper (although I don't have the time to read the whole thing).

I see that you've followed the pseudocode of the paper quite closely, so I'm pretty much happy to believe that you've done that correctly (having checked a few bits myself) 🙂. The number of tests give me confidence that you have.

My one slight concern about the code is the following. At points, the paper defines things in terms of the vertices _less than_ another vertex (i.e. `5 < 7`). Seeing as you're most familiar with the code, I would just like you to convince yourselves that you are **not** relying on either the list of out-neighbours or the list of in-neighbours of a vertex being sorted. Because they aren't guaranteed to be sorted (although in-neighbours *usually* are sorted into ascending order).

Some quick testing indicates this isn't an issue, but I'd appreciate it if you could think about it.

I've 'requested changes' for my suggestions on the documentation.

As for the question 'what should the function return' which @james-d-mitchell brought up on the call today, I now understand that to mean 'should `DominatorsTree` exist as a user-facing function'? And I don't particularly mind about this either way. But if you keep it, I'm happy for it to return a record as it currently does (just please document what those things mean, mathematically 🙂 ).

gap/oper.gi
wilfwilson: See https://github.com/digraphs/Digraphs/pull/336#issuecomment-755413562

marinaanagno:
> Thanks @marinaanagno and @samanthaharper for your work on this! I've looked at it for quite a while, including reading parts of the paper (although I don't have the time to read the whole thing).
> 
> I see that you've followed the pseudocode of the paper quite closely, so I'm pretty much happy to believe that you've done that correctly (having checked a few bits myself) 🙂. The number of tests give me confidence that you have.
> 
> My one slight concern about the code is the following. At points, the paper defines things in terms of the vertices _less than_ another vertex (i.e. `5 < 7`). Seeing as you're most familiar with the code, I would just like you to convince yourselves that you are **not** relying on either the list of out-neighbours or the list of in-neighbours of a vertex being sorted. Because they aren't guaranteed to be sorted (although in-neighbours _usually_ are sorted into ascending order).
> 
> Some quick testing indicates this isn't an issue, but I'd appreciate it if you could think about it.
> 
> I've 'requested changed' for my suggestions on the documentation.
> 
> As for the question 'what should the function return' which @james-d-mitchell brought up on the call today, I now understand that to mean 'should `DominatorsTree` exist as a user-facing function'? And I don't particularly mind about this either way. But if you keep it, I'm happy for it to return a record as it currently does (just please document what those things mean, mathematically 🙂 ).

Thank you so much for the review @wilfwilson, I will look carefully at all the changes you suggested probably tomorrow :)

gap/oper.gi
marinaanagno: Yes, this is a bug in the paper that took us ages to fix! :)

gap/oper.gi
```diff
@@ -1725,6 +1725,146 @@ function(D, list)
...
+InstallMethod(Dominators, "for a digraph and a vertex",
+[IsDigraph, IsPosInt],
+function(D, root)
+  local tree, preorder, result, u, v;
+  tree := DominatorTree(D, root);
+  preorder := tree.preorder;
+  tree := tree.idom;
+  result := [];
+  for v in preorder do
+    result[v] := [v];
```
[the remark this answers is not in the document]
marinaanagno: I remember talking about this with @james-d-mitchell at some point and saying that we need to think about it, and it is probably something we need to reconsider! Thanks for pointing that out! :)

james-d-mitchell:
@wilfwilson @marinaanagno @samanthaharper I think this is good to go, except possibly: I'm not sure if we ever arrived at a decision about whether or not a vertex should dominate itself, any thoughts?

marinaanagno:
@james-d-mitchell thanks for this! According to the theory a vertex is not dominated by itself and if there's no reason to include it in the list of dominators, I would suggest we make the algorithm agree with the theory. :)

wilfwilson:
That was my feeling too.

james-d-mitchell:
Thanks @marinaanagno and @wilfwilson I've updated the PR to not include a vertex as a dominator of itself (I think). The only other thing that I think was pending was whether or not `DominatorTree` should be user facing. I think it should because it's really the fundamental thing computed by the algorithm, and `Dominators` is just a useful additional feature. I've updated `DominatorTree` to only return the preorder and the immediate dominators, both of which are used by the current `Dominators` method.

marinaanagno (approved):

marinaanagno:
@james-d-mitchell looks good to me! As far as `DominatorTree` is concerned, as a user working on this project, I almost always use `DominatorTree`, because it gives me much more information, such as preorder values, semidominators (which is now removed) etc. Unless there is any particular reason to prevent us from making it user facing, and since documentation for it already exists, I don't see any reason to not have it available for the user to use. :) Also, semidominators was a pretty important notion for this algorithm and it was very useful to have it as part of the output in `DominatorTree` and maybe keeping it in the record would be an option, however I'm not really sure whether it is commonly used as a concept, so maybe there's indeed no reason to keep it in the record. :)

wilfwilson (approved):
Nice! I've just updated this with the master branch; I'll merge if the CI passes again (which it should!).

marinaanagno:
Thanks so much @wilfwilson ! :)

Issue #260, opened 2019-09-25
Behaviour of `DigraphRemoveEdge`/`DigraphRemoveEdges` is inconsistent


flsmith:
Currently if `digraph` is immutable:
- `DigraphRemoveEdges(digraph, [ ])` returns `digraph`, but
- `DigraphRemoveEdges(digraph, [[src, ran]])` always returns an immutable copy of `digraph` even when `[src, ran]` is invalid, and
- `DigraphRemoveEdge(digraph, [src, ran])` always returns an immutable copy of `digraph` even when `[src, ran]` is invalid

which seems slightly odd, do we want this behaviour?

wilfwilson:
For these kinds of functions, I'm all for the operation returning the identical unchanged immutable digraph in the case that the argument is immutable and the operation doesn't want to change the digraph in any way.

I can't see why we should insist on creating on a new immutable copy in this case (which will take up new memory, and which will lose all of its stored attributes). If a user wants a new immutable copy, they can always ask for one explicitly.

james-d-mitchell:
Very late to the party, but I agree with @wilfwilson. If the digraph is not changed, then there's no need to copy it.

mtorpey:
@saffronmciver: Hopefully see you at the meeting tomorrow!  Did this issue go any further, or would it be a good one for you to get into as part of the new VIP?

saffronmciver:
I didn't get much further on this. As you suggest I think this would be a good issue for me to work on; I will start working on it again today.

saffronmciver:
As mentioned previously, an error might be appropriate when calling `DigraphRemoveEdge` where the edge does not exist in the graph. The case where the edge does not exist is the only case for `DigraphRemoveEdge` where no changes are made and a copy is returned unnecessarily. So if an error were to occur, there would be no unnecessary copies made. 

After looking at it again, I realise only `DigraphReverseEdge`, `DigraphReverseEdges`, `DigraphRemoveVertices` and `QuotientDigraph` create immutable copies when no changes were made for the cases below:

- `DigraphReverseEdge(D, src, ran)` when reversing an edge where `src = ran` 
- `DigraphReverseEdges` when the above is the case for all vertices
- `DigraphRemoveVertices` when all vertices are not present (this function as well as `DigraphRemoveVertex` does not throw an error when a vertex does not exist)
- `QuotientDigraph` for a graph with no vertices and an empty partition

wilfwilson:
[...]

PR #340, opened 2020-11-11
Make behaviour of `DigraphRemoveEdge{s}` consistent when removing no edges

Peter-Ing:
Added an additional method to preserve properties in the case of empty removals.

This addresses issue #260. _(Note from Wilf: currently only partially.)_

### DIFF
--- a/gap/oper.gi
+++ b/gap/oper.gi
@@ -279,6 +279,16 @@ InstallMethod(DigraphRemoveEdges, "for an immutable digraph and a list",
 [IsImmutableDigraph, IsList],
 {D, edges} -> MakeImmutable(DigraphRemoveEdges(DigraphMutableCopy(D), edges)));
 
+InstallMethod(DigraphRemoveEdges, "for an immutable digraph and an empty list",
+[IsImmutableDigraph, IsList and IsEmpty],
+{D, edges} -> D);
+
+InstallMethod(DigraphRemoveEdge, "for an immutable digraph and a list",
+[IsImmutableDigraph, IsList and IsEmpty],
+function(D, edges)
+  ErrorNoReturn("the 2nd argument must be non empty,");
+end);
+
 InstallMethod(DigraphReverseEdge,
 "for a mutable digraph by out-neighbours and two positive integers",
 [IsMutableDigraph and IsDigraphByOutNeighboursRep, IsPosInt, IsPosInt],
--- a/tst/standard/oper.tst
+++ b/tst/standard/oper.tst
@@ -39,6 +39,13 @@ gap> DigraphRemoveEdges(gr, [[2, 1]]);
 <immutable digraph with 2 vertices, 1 edge>
 gap> last = gr;
 true
+gap> DigraphRemoveEdges(gr, []);
+<immutable digraph with 2 vertices, 1 edge>
+gap> last = gr;
+true
+gap> DigraphRemoveEdges(gr, [[1, 3]]);
+Error, the 3rd argument <ran> must be a vertex of the digraph <D> that is the \
+1st argument,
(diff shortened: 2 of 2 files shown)


gap/oper.gi
```diff
@@ -270,6 +270,14 @@ InstallMethod(DigraphRemoveEdges, "for an immutable digraph and a list",
 [IsImmutableDigraph, IsList],
 {D, edges} -> MakeImmutable(DigraphRemoveEdges(DigraphMutableCopy(D), edges)));
 
+InstallMethod(DigraphRemoveEdge, "for an immutable digraph and an empty list",
+[IsImmutableDigraph, IsList and IsEmpty],
+{D, edge} -> D);
```
wilfwilson, line 275: I don't think you should add this method. It is undocumented and I think deliberately unsupported: `DigraphRemoveEdge` for a digraph and a list is documented for a _pair of vertices_. So I think the current behaviour is correct:
```
gap> DigraphRemoveEdge(ChainDigraph(2), []);
Error, the 2nd argument <edge> must be a list of length 2, at /Users/Wilf/gap/pkg/digraphs/gap/oper.gi:252 called from
DigraphRemoveEdge( DigraphMutableCopy( D ), edge
 ) at /Users/Wilf/gap/pkg/digraphs/gap/oper.gi:260 called from
<function "DigraphRemoveEdge for a immutable digraph and a list">( <arguments>\
 )
```

wilfwilson (changes requested):
Please can you also add a test for this behaviour, which verifies that the digraph return by `DigraphRemoveEdges` when given an immutable digraph `D` and an empty list is indeed the identical object to `D`.

wilfwilson:
Oh wait, what is the relationship between this PR and #338?

Peter-Ing:
> Oh wait, what is the relationship between this PR and #338?

this request should have been an update to #338 but as it hasn't updated it and comments have been left here I have closed #338.

gap/oper.gi
Peter-Ing: This pull request was opened to respond to #260 however my version of digraphs was outdated upon writing this, after updating the update renders this pull request redundant and I believe resolves #260 aswell.

james-d-mitchell:
@Peter-Ing I agree with @wilfwilson. I think the intended changes would be the following:

if `digraph` is immutable:
- `DigraphRemoveEdges(digraph, [ ])` returns `digraph` 
- `DigraphRemoveEdge(digraph, [ ])` gives the current error (the argument is not valid)
- `DigraphRemoveEdges(digraph, [[src, ran]])` returns an immutable copy of `digraph` only when `[src, ran]` is valid. If `[src, ran]` is not valid, then the function could either give an error, or just return `digraph` (i.e. not a copy)
- `DigraphRemoveEdge(digraph, [src, ran])` returns an immutable copy of `digraph` only when `[src, ran]` is valid. If `[src, ran]` is not valid, then the function could either give an error, or just return `digraph` (i.e. not a copy)

I'm not sure which is preferable: giving an error or just returning `digraph` and doing nothing. The latter would be more consistent with the current behaviour, but the former would probably be more helpful. Any thoughts?

james-d-mitchell (changes requested):
Thanks @Peter-Ing, I think this PR should be reworked to reflect the comments I made, feel free to leave it open, and just rework your new changes on top of this.

wilfwilson:
I tend to prefer the former kind of behaviour, I think it is what most people (or at least I) would expect. The latter kind of behaviour seems like it's done for optimisation, letting you reduce code (not having a separate existence check first, in some cases) and save time (by not twice checking whether an edge exists ). Unless users ask for that kind of behaviour, I don't think we should proactively offer it.

james-d-mitchell:
Resolved by #723

PR #341, opened 2020-11-12
ci: use azure pipelines

james-d-mitchell:
This is a first approximation of how to do the CI using azure pipelines, and docker. I don't have any preference for azure pipelines over github actions, but I couldn't figure out how to get docker containers to work using github actions (see https://github.com/libsemigroups/libsemigroups/issues/254), and I could figure that out using azure pipelines. The advantage of doing it with a docker container is: that it's possible to ensure that the CI is set up and will run locally (which is faster than waiting for some remote jobs to fail repeatedly); the building of GAP is completely separate (and nicely encapsulated).

There are some issues:

1. The gap-docker container only supports the "latest" tag, and so it isn't currently possible to specify which version of GAP is used inside the container. This would be trivial to fix for the maintainer of:

https://hub.docker.com/r/gapsystem/gap-docker/

I don't know if @alex-konovalov might be interested in supporting this feature, or allowing me access to the docker hub account so that I can enable it? For example, we are doing this with the `libsemigroups` docker container here:

https://hub.docker.com/r/libsemigroups/libsemigroups-docker/tags

2. There's currently only one job, it should be easy enough to scale this to the original jobs that we ran on travis once issue 1 is fixed. 

3. I think I probably broke the CI scripts support for appveyor.

PR #342, opened 2020-11-14
Implement initial GitHub actions support for CI

wilfwilson:
For now, this should suffice for getting a useful CI system back. This takes advantage of @ssiccha's work on https://github.com/gap-actions and @fingolfin's https://github.com/gap-system/pkg-ci-scripts. As you can see, it doesn't use much code.

Currently, all of the package tests are run with GAP `master` and `stable-4.11`. We can add `stable-4.10` and `stable-4.9` support if we manually compile NautyTracesInterface (which should be easy but that can come later).

This means we're missing the linting test, testing with different versions of packages, and also our custom code coverage test, for now. However, this system includeas Codecov integration - so we could look into turning that on. This also doesn't use Docker, but it is quite possible that someone using this kind of setup for a GAP package will do that, and then we can copy their solution.

So does it work? [Here is a hopefully successful build](https://github.com/wilfwilson/Digraphs/actions/runs/363257535) with the current contents of this PR, and [here is a build where I deliberately introduced an error](https://github.com/wilfwilson/Digraphs/actions/runs/363199153)  into `testinstall.tst`.

The list of all builds is at https://github.com/wilfwilson/Digraphs/actions

### DIFF
--- a/.github/workflows/GAP.yml
+++ b/.github/workflows/GAP.yml
@@ -1,0 +1,32 @@
+name: "GAP tests"
+
+# Trigger the workflow on push or pull request
+on:
+  - push
+  - pull_request
+
+jobs:
+  test:
+    name: "GAP ${{ matrix.gap-branch }} on ${{ matrix.os }}"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        digraphs-lib:
+          - digraphs-lib-0.6
+        gap-branch:
+          - master
+          - stable-4.11
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: gap-actions/setup-gap-for-packages@v1
+        with:
+          GAP_PKGS_TO_CLONE: "NautyTracesInterface"
+          GAP_PKGS_TO_BUILD: "io orb profiling grape NautyTracesInterface datastructures"
+          GAPBRANCH: ${{ matrix.gap-branch }}
+      - name: "Install digraphs-lib"
+        run: curl --retry 5 -L -O https://digraphs.github.io/Digraphs/${{ matrix.digraphs-lib }}.tar.gz
+             && tar xf ${{ matrix.digraphs-lib }}.tar.gz
+      - uses: gap-actions/run-test-for-packages@v1
--- a/tst/testall.g
+++ b/tst/testall.g
@@ -1,0 +1,6 @@
+LoadPackage("digraphs");
+if DigraphsTestInstall() and DigraphsTestAll() and DigraphsTestExtreme() then
+  FORCE_QUIT_GAP(0);
+fi;
+
+FORCE_QUIT_GAP(1);


james-d-mitchell:
@wilfwilson note the actions haven't actually run here.

fingolfin:
They only start running here once you have merged it; this is both for security, and also to prevent others from reducing your free minutes (which doesn't really matter for OSS repositories as they have unlimited minutes)

.travis.yml
```diff
@@ -114,7 +114,7 @@ install:
 
 before_script:
   - export GAPROOT="$HOME/gap"
-  - scripts/travis-build-dependencies.sh
+  - ci/travis-build-dependencies.sh
```
fingolfin, line 117: Our actions really shouldn't put anything inside the package dir, to prevent clashes like this. Some temp location outside should be used instead. Or if that's infeasible, at least a really unique dir name should be used. CC @ssiccha

james-d-mitchell:
Thanks @fingolfin, I see, I also see that @wilfwilson added a link to the successful build. 

I'm not really in favour of this approach. I think it'd be better to use a docker container containing GAP for the reasons I outlined in PR #341. At the same time, this is certainly significantly better than nothing.

.travis.yml
wilfwilson: Yeah I renamed this because `pkg-ci-scripts` was getting cloned into a `scripts` directory.

wilfwilson:
I think it makes sense to merge and use this for now, so that we can have some amount of useful CI until we have a solution using docker.

.travis.yml
wilfwilson: Although for now I'm using a forked version of one of the actions https://github.com/wilfwilson/setup-gap-for-packages/blob/master/action.yml - I'll submit a PR to the  `gap-actions` repo soon.

wilfwilson: https://github.com/gap-actions/setup-gap-for-packages/pull/6

wilfwilson:
I've removed some changes based on what @fingolfin said, so it's even smaller. The builds on this PR can all be seen at https://github.com/wilfwilson/Digraphs/actions.

wilfwilson:
I would be happy for this to be merged as a first step. I especially like its simplicity - the complication of our previous setup is mostly why I didn't enjoy working on it.

As a next step, we can investigate what the CAP/homalg people are doing with docker for GAP, see e.g. their workflow here https://github.com/homalg-project/CAP_project/blob/master/.github/workflows/Tests.yml

.github/workflows/CI.yml
```diff
@@ -0,0 +1,32 @@
+name: CI
+
+# Trigger the workflow on push or pull request
+on:
+  - push
+  - pull_request
+
+jobs:
+  test:
+    name: "Test: ${{ matrix.gap-branch }} on ${{ matrix.os }}"
```
fingolfin, line 10: Perhaps this would avoid potential confusion?
```suggestion
    name: "Test: GAP ${{ matrix.gap-branch }} on ${{ matrix.os }}"
```

wilfwilson: I thought I'd already done that! Thanks.

.github/workflows/GAP.yml
```diff
@@ -0,0 +1,32 @@
...
+          - digraphs-lib-0.6
+        gap-branch:
+          - master
+          - stable-4.11
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: gap-actions/setup-gap-for-packages@master
+        with:
+          GAP_PKGS_TO_CLONE: "wilfwilson/NautyTracesInterface"
```
james-d-mitchell, line 26: Why's this necessary? The "wilfwilson" part?

.github/workflows/GAP.yml
```diff
@@ -0,0 +1,32 @@
...
+          GAP_PKGS_TO_BUILD: "io orb profiling grape NautyTracesInterface datastructures"
+          GAPBRANCH: ${{ matrix.gap-branch }}
+      - name: "Install digraphs-lib"
+        run: curl --retry 5 -L -O https://digraphs.github.io/Digraphs/${{ matrix.digraphs-lib }}.tar.gz
+             && tar xf ${{ matrix.digraphs-lib }}.tar.gz
+      - uses: wilfwilson/run-test-for-packages@master
```
james-d-mitchell, line 32: Same, what does this do? Where's the source code?

tst/testall.g
```diff
@@ -0,0 +1,6 @@
+LoadPackage("digraphs");
+if DigraphsTestInstall() and DigraphsTestAll() and DigraphsTestExtreme() then
+  FORCE_QUIT_GAP(0);
```
james-d-mitchell, line 3: Doesn't `QuitGap(0);` do more or less the same thing in this context?

tst/testall.g
```diff
@@ -0,0 +1,6 @@
...
+fi;
+
+FORCE_QUIT_GAP(1);
```
james-d-mitchell, line 6: Same comment as above?

james-d-mitchell (commented):
I'm happy enough with this, I don't particularly like the opacity of the `gap-actions` steps, nor `wilfwilson/run-test-for-packages`, perhaps the required actions can be moved into the "digraphs" organisation? Not the `gap-actions` one, but the other one.

tst/testall.g
wilfwilson: Yes.

tst/testall.g
wilfwilson: Yes. Although if this line has been reached, then something went wrong, and that something could have gone badly wrong, and so I think it makes sense to be as drastic as possible when quitting GAP.

.github/workflows/GAP.yml
wilfwilson: The purpose of the action `gap-actions/setup-gap-for-packages` is to install GAP and its packages.

`NautyTracesInterface` doesn't have any releases, and so it is not installed by the action `gap-actions/setup-gap-for-packages` by default.

Therefore, the action needs to be told to install the `NautyTracesInterface` package explicitly. Writing
```
GAP_PKGS_TO_CLONE: "NautyTracesInterface"
```
would clone the master branch of `NautyTracesInterface` from the `gap-packages` repository. I intend for this to be what is written ultimately.

After cloning/installing the packages, `gap-actions/setup-gap-for-packages` then compiles the packages using GAP's `bin/BuildPackages.sh` script.

However, this currently fails for `NautyTracesInterface`, because `NautyTracesInterface` has an extra non-standard step for compiling it. (You have to first change into the nauty directory, and then compile nauty, before compiling the package.)

Since GAP 4.11, `BuildPackages.sh` looks for a `prerequisites.sh` script in each package to run before compiling the package; the `Semigroups` package has such a script, for example, to install `libsemigroups`.

`NautyTracesInterface` could use a `prerequisites.sh` script to do the step of compiling nauty, but it doesn't have one yet. I have made a pull request to `NautyTracesInterface` to do just this https://github.com/gap-packages/NautyTracesInterface/pull/40 however it has not yet been merged.

Therefore, for now, instead of pointing to `gap-packages/NautyTracesInterface`, I am pointing to my own fork, which is where the PR lives. Once the PR is merged, this line can be updated as follows:
```suggestion
          GAP_PKGS_TO_CLONE: "NautyTracesInterface"
```
I could also just compile `NautyTracesInterface` manually with an extra line or two in this workflow. However, I am trying to make this workflow as streamlined as possible.

.github/workflows/GAP.yml
wilfwilson: Where you see lines like this in a GitHub workflow, this is referring to a GitHub repository that hosts a GitHub workflow.

It is saying: at this point, run the current `master` branch of the workflow `run-test-for-packages` of the user/organisation `wilfwilson`. So the code that it is running is the workflow here: https://github.com/wilfwilson/run-test-for-packages

This is a fork of the workflow https://github.com/gap-actions/run-test-for-packages

I will use the `gap-actions` workflow directly, once the PR that I have made to it (https://github.com/gap-actions/run-test-for-packages/pull/1) has been merged.

The purpose of this step is to is to load GAP and run the Digraphs tests, and upload code coverage data. Ultimately, it runs the script https://github.com/gap-system/pkg-ci-scripts/blob/master/run_tests.sh which loads GAP and checks that the file `tst/testall.g` that I introduce below returns a 0 exit code.

wilfwilson:
The references involving `wilfwilson` are temporary (I imagine on the order of a few days) until pull requests are merged, as explained above.

I disagree that the lines involving `gap-actions` are opaque. They clearly tell you precisely what workflow is being used, e.g. the line
```
uses: gap-actions/setup-gap-for-packages@master
```
tells you that it is running the action from the `master` branch of the repository https://github.com/gap-actions/setup-gap-for-packages.

You can go there to see what it is doing: using the scripts at https://github.com/gap-system/pkg-ci-scripts (used by almost all other GAP packages) to build GAP, download and build packages, and build the current package. I think it is quite likely that someone will make this action (or an equivalent one) use Docker, which we could then benefit from, too.

Using these actions is about having as little custom code as possible to run our CI - the idea is that pretty much all GAP packages will use these workflows for their tests. We can always fork the various workflows to the `digraphs` organisation if we don't like the direction that they're taking, but for now I think it makes sense to use this shared code (and benefit from the shared effort) amongst GAP packages.

tst/testall.g
wilfwilson: Actually. `QuitGap` is only introduced in GAP `master` branch, so we can't use it.

tst/testall.g
wilfwilson: As mentioned above, `QuitGap` is a new function in the GAP `master` branch, not available in any released versions, so we can't use it until Digraphs requires the upcoming GAP 4.12 or later.

james-d-mitchell:
@wilfwilson you make some reasonable points. 

However, it will, in all likelihood, be me who is left to administer the CI for Digraphs when the other developers are having breaks or pursuing other interests/careers. As such I'd like to have as good an overview of how the CI works for Digraphs and Semigroups as possible, and the ability to spend the little time I have for this type of thing working on the actual code rather than on the related politics. 

There's a good reason that Semigroups and Digraphs don't use the GAP infrastructure for the tests: we started using CI before the GAP infrastructure existed. 

> I think it is quite likely that someone will make this action (or an equivalent one) use Docker, which we could then benefit from, too.

In my PR #348 this is already implemented and working now.

wilfwilson:
I appreciate that you would like to minimise the amount of time that you spend administering the CI system. I feel similarly.

I also appreciate that you want to understand exactly what this CI system is doing, and how it works. I continue to be happy to help explain the system that I am proposing in this PR.

Because of these shared desires, I believe it is important to aim for simplicity in the CI system. This workflow itself consists of a small amount of code, and the other scripts that it relies on are, I believe, quite clear, concise, and robust. It will perhaps require some tweaking in the short term (which I hint at with the word 'initial' in the PR title), which I am happy to do, but in the medium term I expect the required maintenance to be minimal.

I also feel that redundancy is a useful feature to have in our CI system. By having both GitHub actions and Azure Pipelines integrations, if either of them is down, or broken, then we can hopefully still see the results of the other.

Having GitHub Actions integration in particular has a further benefit: anyone who forks the Digraphs package (for example, [my own fork](https://github.com/wilfwilson/Digraphs), or [your fork](https://github.com/james-d-mitchell/Digraphs)) immediately benefits from having a working CI setup for their own fork with no further action required. This is because anyone forking Digraphs is already using GitHub, and they have no need to create an Appveyor/Travis/Azure Pipelines account to access this automated testing.

> it will, in all likelihood, be me who is left to administer the CI for Digraphs when the other developers are having breaks or pursuing other interests/careers

I am sorry to hear that you do not feel that you can rely on my continued support with helping to administer the Digraphs package. If you feel that I have let you down in this regard in the last past, then I sincerely apologise.

> There's a good reason that Semigroups and Digraphs don't use the GAP infrastructure for the tests: we started using CI before the GAP infrastructure existed.

I understand, indeed I know this from experience. However, I don't think having our pre-existing setup should stop us from _also_ plugging into the GAP infrastructure now: indeed, this PR takes only 32 lines on our end to implement a CI system from scratch (plus a 6-line test file). If we ever decide that it isn't working for us in the future, we can scrap it.

> In my PR #348 this is already implemented and working now.

That's great!

james-d-mitchell:
@wilfwilson I agree that having redundancy is a good idea, and that having both setups is the best way forwards for now. 

You don't have anything to apologise for, I don't, and haven't felt let down by you at any point on any issue.  I was very fortunate to have you doing basically all of the work on managing Digraphs (CI/releases/code review/etc) more or less from its inception until the start of 2020. Since then I did have to assume some of the responsibilities that you previously had (I pretty sure we talked about this at some point), and I was happy to do this, and didn't expect anything to be different.

Anyway, I'm pretty sure that we agree that the best way forwards is to merge both your PR and mine, no?

wilfwilson:
Ultimately, yes, I think both should be merged - but happy to take further suggestions into account.

.github/workflows/GAP.yml
wilfwilson: This now uses the `gap-actions` workflow directly, the reference to my `wilfwilson` fork is gone.

.github/workflows/GAP.yml
wilfwilson: This now uses the `gap-packages` repository directly, the reference to my `wilfwilson` fork is gone.

james-d-mitchell (approved):
I'm happy for this to be merged at this point, unless there's anything further you want to wait for @wilfwilson? Sooner rather than later seems preferable.

wilfwilson:
Thanks James!

PR #346, opened 2020-11-14
Don't copy an immutable digraph when adding zero vertices

wilfwilson:
This is similar to #338 and #340, for edges (although I made these changes over a year ago!).

### DIFF
--- a/doc/oper.xml
+++ b/doc/oper.xml
@@ -320,8 +320,10 @@ true
 
     If <A>digraph</A> belongs to <Ref Filt="IsMutableDigraph"/>, then the
     vertices are added directly to <A>digraph</A>, which is changed in-place.
-    If <A>digraph</A> belongs to <Ref Filt="IsImmutableDigraph"/>, then the
-    result is returned as an immutable digraph.
+    If <A>digraph</A> belongs to <Ref Filt="IsImmutableDigraph"/>, then
+    <A>digraph</A> itself is returned if no vertices are added
+    (i.e. <C><A>m</A>=0</C> or <A>labels</A> is empty), otherwise
+    the result is a new immutable digraph.
     <P/>
 
     <Example><![CDATA[
--- a/gap/oper.gi
+++ b/gap/oper.gi
@@ -63,8 +63,12 @@ end);
 
 InstallMethod(DigraphAddVertices, "for an immutable digraph and list",
 [IsImmutableDigraph, IsList],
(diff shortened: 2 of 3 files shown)


james-d-mitchell (changes requested):
I think the manual entry needs to be amended to note that the returned graph is not copied if 0 new nodes are added. Since I think this PR changes the behaviour, it's not inconceivable that there's code out there (or in here) that relies on the old behaviour.

wilfwilson:
I have updated the manual entry for `DigraphAddVertices` accordingly.

doc/oper.xml
```diff
@@ -320,8 +320,10 @@ true
 
     If <A>digraph</A> belongs to <Ref Filt="IsMutableDigraph"/>, then the
     vertices are added directly to <A>digraph</A>, which is changed in-place.
-    If <A>digraph</A> belongs to <Ref Filt="IsImmutableDigraph"/>, then the
-    result is returned as an immutable digraph.
+    If <A>digraph</A> belongs to <Ref Filt="IsImmutableDigraph"/>, then
+    <A>digraph</A> itself is returned if no vertices are added
+    (i.e. <C><A>m</A>=0</C> or <A>labels</A> is empty), otherwise
+    the result is returned as a new immutable digraph.
```
james-d-mitchell, line 326:
```suggestion
    the result is a new immutable digraph.
```
??

gap/oper.gi
```diff
@@ -63,8 +63,12 @@ end);
 
 InstallMethod(DigraphAddVertices, "for an immutable digraph and list",
 [IsImmutableDigraph, IsList],
-{D, labels} -> MakeImmutable(DigraphAddVertices(DigraphMutableCopy(D),
-                                                labels)));
+function(D, labels)
+  if IsEmpty(labels) then
```
james-d-mitchell, line 67: I know you didn't name this in your PR, but why is the argument named `labels`?

james-d-mitchell (dismissed):

gap/oper.gi
wilfwilson: Because this method adds `Length(labels)` new vertices, where the `i`th vertex to be added is given the label `labels[i]`.

doc/oper.xml
wilfwilson: Will do so. Was just editing the old language as little as possible.

gap/oper.gi
james-d-mitchell: Makes sense, thanks!

james-d-mitchell (approved):

PR #347, opened 2020-11-14
Small improvements to IsSymmetricDigraph

wilfwilson:
For the general method: first check whether there are an odd number of edges (ignoring loops), and return `false` if so. Computing `IsMultiDigraph`, `DigraphNrEdges`, and `DigraphLoops` should be a lot faster than computing the in-neighbours, which is what happens next.

And a method for when we have the adjacency matrix.

### DIFF
--- a/gap/prop.gi
+++ b/gap/prop.gi
@@ -199,6 +199,13 @@ InstallMethod(IsSymmetricDigraph, "for a digraph by out-neighbours",
 function(D)
   local out, inn, new, i;
 
+  if not IsMultiDigraph(D)
+      and (DigraphNrEdges(D) - Length(DigraphLoops(D))) mod 2 = 1 then
+    return false;
+  elif HasAdjacencyMatrix(D) then
+    TryNextMethod();
+  fi;
+
   out := OutNeighbours(D);
   inn := InNeighbours(D);
   if not ForAll(out, IsSortedList) then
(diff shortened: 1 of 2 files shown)


james-d-mitchell (approved):
I'm happy with this, it'd be nice to have some test cases where the performance improvement can be observed, but I can live without them too. Perhaps we can run the new CI on these when it's merged, before merging?

wilfwilson:
I just fully rebased this on `stable-1.3` just to make sure it definitely picked up all of the CI stuff that has been merged so far. Seems like it has 🙂 But it means your approving review has now been dismissed.

It's probably quite likely that the adjacency matrix method will make very little difference, since any digraph small enough to know its adjacency matrix will probably let you compute anything about it very quickly. Happy to remove the method if you prefer.

I think the other method has a much stronger case: it's surely (?!) much cheaper in almost all cases to compute whether something is a multidigraph, and count its edges, and count its loops (or already know these things at creation) than it is to construct in-neighbours.

wilfwilson:
(This PR contains #351 to make sure that AppVeyor tests; I'll remove that commit when #351 is merged, or this can be merged including the extra commit, and #351 can be closed).

gap/prop.gi
```diff
@@ -211,6 +218,22 @@ function(D)
   return inn = out;
 end);
 
+InstallMethod(IsSymmetricDigraph, "for a digraph with adjacency matrix",
+[IsDigraph and HasAdjacencyMatrix],
+function(D)
+  local mat, n, i, j;
+  mat := AdjacencyMatrix(D);
+  n := DigraphNrVertices(D);
+  for i in [1 .. n] do
```
james-d-mitchell, line 227:
```suggestion
  for i in [1 .. n - 1] do
```

james-d-mitchell (dismissed):
I'm happy with this one minor comment, which makes no difference.

wilfwilson:
@james-d-mitchell It looks like the Codecov integration is working reasonably well.

james-d-mitchell (approved):
J'approve

PR #350, opened 2020-11-16
Create ci.yml

james-d-mitchell:


james-d-mitchell:
This is ready to go, and doesn't depend on which approach we finally choose for the CI

wilfwilson (approved):
I think this should go into `stable-1.3` rather than `master`

james-d-mitchell:
Thanks @wilfwilson, it's into `stable-1.3` now.

wilfwilson (approved):

PR #352, opened 2020-11-17
Create a .codecov.yml file to configure Codecov

wilfwilson:
The GitHub actions workflows that I added include Codecov integration. This means that coverage reports get automatically uploaded to Codecov, which we can browse like on here: https://codecov.io/gh/digraphs/Digraphs/branch/stable-1.3

This PR as a configuration file to decide various options for this integration.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,15 @@
+coverage:
+  precision: 2      # Report coverage to 2 decimal places
+  round: down       # Round down
+  range: "90...100" # Anything below 90% is coloured red (considered bad)
+  status:           # Various customisations that it seems should be off
+    project: off
+    patch: off
+    changes: off
+
+comment:
+  layout: "header, diff, changes, tree"
+  behavior: default
+
+ignore:
+ - "extern"         # Do not check coverage on the extern directory


.codecov.yml
```diff
@@ -0,0 +1,13 @@
+coverage:
+  precision: 1      # Report coverage to 1 decimal place
+  round: down       # Round down
+  range: "90...100" # Anything below 90% is coloured red (considered bad)
+  status:           # Various customisations that it seems should be off
+    project: off
+    patch: off
+    changes: off
+
+comment: false      # Do not post a comment on PRs with a code coverage report
```
james-d-mitchell, line 10: I'm basically happy with this PR, I wonder why we don't want comments on PR's though? This is far superior to the rudimentary coverage checks we did previously.

james-d-mitchell (approved):

.codecov.yml
wilfwilson: Yeah, let's do it and see how we find it. Was just wanting to get something minimal started with this PR. We might want to tweak exactly what the comments look like, what information they contain, but we can tweak that later.

wilfwilson:
Coverage _should_ have increased to over 95%, since we're now excluding the `extern` directory. This is what Codecov says for this commit on my fork: https://codecov.io/gh/wilfwilson/Digraphs/tree/baf7bb6d528efd792a5ad91921cb341f70e16cf2

I guess that the Codecov comment in the PR here is reporting unchanged coverage because the base branch `stable-1.3` _is_ including `extern`, to make it fair? I guess we'll see when it's merged...

james-d-mitchell (approved):
Looks good!

PR #354, opened 2020-11-18
Add ModularProduct

james-d-mitchell:
This PR partially introduces the `ModularProduct` of two digraphs, as used in @le27's PR #334. When trying to understand the code in PR #334, I wrote the method in this PR. It seems to be about 10 times faster than the method in PR #334 for finding the modular product, but it doesn't do exactly the same thing. 

 @le27 can I suggest that you finish this PR off (add tests and documentation), and then use the `ModularProduct` in your other PR #334? 

It's possible that it would be beneficial to have a `ReducedModularProduct` operation too (which basically returns the same thing as `ReducedDigraph(ModularProduct(D));` but without the creation of the intermediate (possibly too large) `ModularProduct(D)`). Having said that I don't think the time consuming part of `ReducedDigraph(ModularProduct(D));` is the call to `ReducedDigraph` so this might also not be worth it.

PR #361, opened 2020-12-02
Maximal common subdigraph/Minimal common superdigraph

le27:
Implements the functions MaximalCommonSubdigraph and MinimalCommonSuperdigraph.

MaximalCommonSubdigraph takes 2 digraphs and returns a third digraph with maximum vertex set size which embedds in both of them (as well as the embeddings).

MinimalCommonSuperdigraph takes 2 digraphs and returns a third digraph with minimum vertex set size into which both of them embed (as well as the embeddings).

It also includes a function which builds the ModularProduct of 2 digraphs. This is analogous to the modular product of graphs as described here: https://en.wikipedia.org/wiki/Modular_product_of_graphs

### DIFF
--- a/doc/grahom.xml
+++ b/doc/grahom.xml
@@ -798,3 +798,66 @@ false
     </Description>
   </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="MaximalCommonSubdigraph">
+  <ManSection>
+    <Oper Name="MaximalCommonSubdigraph" Arg="D1, D2"/>
+    <Returns>A list containing a digraph and two transformations.</Returns>
+    <Description>
+  If <A>D1</A> and <A>D2</A> are digraphs without multiple edges, then
+  <C>MaximalCommonSubdigraph</C> returns a maximal common subgraph <C>M</C> of
+  <A>D1</A> and <A>D2</A> with the maximum number of vertices. So <C>M</C> is a
+  digraph which embeds into both <A>D1</A> and <A>D2</A> and has the largest
+  number of vertices amoung such digraphs.
+
+  It returns a list <C>[M, t1, t2]</C> where <C>M</C> is the maximal common
+  subdigraph and <C>t1, t2</C> are transformations embedding <C>M</C> into
+  <A>D1</A> and <A>D2</A> respectively.
+
+<Example><![CDATA[
+gap> MaximalCommonSubdigraph(PetersenGraph(), CompleteDigraph(10));
+[ <immutable digraph with 2 vertices, 2 edges>, 
+  IdentityTransformation, IdentityTransformation ]
+gap> MaximalCommonSubdigraph(PetersenGraph(),
+> DigraphSymmetricClosure(CycleDigraph(5)));
+[ <immutable digraph with 5 vertices, 10 edges>, 
+  IdentityTransformation, IdentityTransformation ]
+gap> MaximalCommonSubdigraph(NullDigraph(0), CompleteDigraph(10));
+[ <immutable empty digraph with 0 vertices>, IdentityTransformation, 
+  IdentityTransformation ]
+]]></Example>
+    </Description>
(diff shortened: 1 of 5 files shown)


doc/grahom.xml
```diff
@@ -798,3 +798,57 @@ false
...
+  IdentityTransformation ]
+]]></Example>
+    </Description>
+  </ManSection>
+<#/GAPDoc>
+
+<#GAPDoc Label="MinimalCommonSuperdigraph">
+  <ManSection>
+    <Oper Name="MinimalCommonSuperdigraph" Arg="D1, D2"/>
+    <Returns> A digraph and two transformations </Returns>
```
wilfwilson, line 831:
```suggestion
    <Returns>A list containing a digraph and two transformations.</Returns>
```

doc/grahom.xml
```diff
@@ -798,3 +798,57 @@ false
...
+  IdentityTransformation ]
+gap> MaximalCommonSubdigraph(NullDigraph(0), CompleteDigraph(10));
+[ <immutable empty digraph with 0 vertices>, IdentityTransformation,
+  IdentityTransformation ]
+]]></Example>
+    </Description>
+  </ManSection>
+<#/GAPDoc>
+
+<#GAPDoc Label="MinimalCommonSuperdigraph">
```
wilfwilson, line 828: Can you please include this piece of documentation at an appropriate point in some chapter? As it stands, this new documentation doesn't get linked when compiling it. That means you can't see it in the compiled manual, or access it via GAP's online help (eg `?MinimalCommonSubdigraph`)

doc/grahom.xml
```diff
@@ -798,3 +798,57 @@ false
...
+  <ManSection>
+    <Oper Name="MinimalCommonSuperdigraph" Arg="D1, D2"/>
+    <Returns> A digraph and two transformations </Returns>
+    <Description>
+  If <A>D1</A> and <A>D2</A> are digraphs without multiedges, then
```
wilfwilson, line 833: We haven't defined the term `multiedges` in the documentation or used it anywhere else so I'd stick with `multiple edges`.
```suggestion
  If <A>D1</A> and <A>D2</A> are digraphs without multiple edges, then
```

doc/grahom.xml
```diff
@@ -798,3 +798,57 @@ false
...
+  <C>MinimalCommonSuperdigraph</C> finds a minimal common subgraph <C>M</C> of
```
wilfwilson, line 834:
```suggestion
  <C>MinimalCommonSuperdigraph</C> returns a minimal common superdigraph <C>M</C> of
```

doc/grahom.xml
```diff
@@ -798,3 +798,57 @@ false
...
+  <A>D1</A>and <A>D2</A> with the minimum number of vertices. It then returns
+  <C>M</C> together with embeddings of <A>D1</A> and <A>D2</A> into <C>M</C>.
```
wilfwilson, line 836: I think it would be a good idea to spell out the mathematical details about what a minimal common superdigraph is (presumably a digraph with the minimal possible number of vertices such that `D1` and `D2` both embed into it).

It would also be a good idea to explicitly say that it returns the answer as a list `[M, t1, t2]` where `t1` and `t2`...

doc/oper.xml
```diff
@@ -1740,3 +1740,42 @@ gap> DigraphDijkstra(D, 1, 2);
 </Description>
 </ManSection>
 <#/GAPDoc>
+
+
+<#GAPDoc Label="ModularProduct">
```
wilfwilson, line 1745: Please remove the stuff about `ModularProduct` since it was already merged in #354.

wilfwilson (changes requested):
I've made some comments on the documentation for `MinimalCommonSuperdigraph` - they all apply to `MaximalCommonSubdigraph`, too 🙂

PR #363, opened 2020-12-13
Add DigraphShortestPathSpanningTree

jdebeule:
Dear James and Wilf, 

I have been using digraphs in a course on group theory. When using CayleyDigraph, I noticed that there was a minor mistake during the creation of a CayleyDigraph. The action was OnRight, it should be OnLeftInverse

Furthermore, when playing with Rubik's cube, in particular the small one, it turned out that having an operation that returns a Shortest Path Spanning Tree with relation to a given vertex would be very handy. I noticed that such a tree is almost there when the ConnectivityDataForVertex is computed. I added some lines of code to make sure the tree is stored. Also it would be useful if the Edges of a CayleyDigraph were labeled, which I also did. 

Finally I've created an operation DigraphShortestPathSpanningTree

Best regards,

Jan

### DIFF
--- a/doc/oper.xml
+++ b/doc/oper.xml
@@ -1400,13 +1400,61 @@ fail
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="DigraphShortestPathSpanningTree">
+<ManSection>
+  <Oper Name="DigraphShortestPathSpanningTree" Arg="digraph, v"/>
+  <Returns>A digraph, or <K>fail</K>.</Returns>
+  <Description>
+    If <A>v</A> is a vertex in <A>digraph</A> and every other
+    vertex of <A>digraph</A> is reachable from <A>v</A>,
+    then this operation returns the shortest path spanning tree of
+    <A>digraph</A> rooted at <A>v</A>.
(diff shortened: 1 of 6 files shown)


wilfwilson:
Thanks for this @jdebeule, hopefully we've review this in the new year. For now, we've taken the fix for `CayleyDigraph` and put it in PR #378 so that we can get that corrected as soon as possible.

wilfwilson:
Hi @jdebeule, thanks again for this pull request. I've modified your original PR somewhat and pushed a new version to your branch, so that is what exists now in this PR.

The stuff about Cayley graphs was put into separate PRs #378 and #385.

For the rest of it, I've added documentation and tests, fixed a few bugs, and added a little functionality (so that the spanning tree respects mutability of the arguments, and so that it retains the appropriate vertex and edge labels, if present).

I think this is ready to be merged now.

wilfwilson (approved):

james-d-mitchell (approved):
Looks good to me! @jdebeule I suggest we merge this pR unless you have any further comments?

jdebeule:
Looks great! I have no further comments! Thanks, Jan

PR #373, opened 2020-12-18
Merge stable-1.3 into master

wilfwilson:
I think this is how I did it before.


wilfwilson:
When I tried to simply do a PR on this repo directly from `stable-1.3` to `master` (see #372) it didn't work, because with our current settings, in order to submit a PR of a branch to `master`, that branch has to be up to date with `master`. In other words, `master` needs to be merged into `stable-1.3`. But we don't want to merge `master` into our stable branches, that's the point of them!

james-d-mitchell (approved):

Issue #367, opened 2020-12-16
Segfault when feeding an multi digraph to GeneratorsOfEndomorphismMonoid


wilfwilson:
```
gap> D := Digraph([[2,2],[]]);
<immutable multidigraph with 2 vertices, 2 edges>
gap> GeneratorsOfEndomorphismMonoid(D);
/usr/local/bin/gap: line 3: 58880 Segmentation fault: 11
```
Presumably `GeneratorsOfEndomorphismMonoid` should not accept multidigraphs.

james-d-mitchell:
This is bad, I'm going to assign myself, and try to fix this soon!

james-d-mitchell:
Resolved in v1.4.0

PR #376, opened 2020-12-23
Resolve Issue #367

james-d-mitchell:
This PR fixes Issue #367, turns out it pointed out two issues in the homomorphism finding code, and both are fixed in this PR.

### DIFF
--- a/src/homos.c
+++ b/src/homos.c
@@ -133,6 +133,7 @@ Obj FuncOutNeighbours(Obj, Obj);
 
 // GAP level things, imported in digraphs.c
 extern Obj IsDigraph;
+extern Obj IsMultiDigraph;
 extern Obj DIGRAPHS_ValidateVertexColouring;
 extern Obj Infinity;
 extern Obj IsSymmetricDigraph;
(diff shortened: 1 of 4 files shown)


wilfwilson (approved):

src/homos.c
```diff
@@ -292,23 +293,28 @@ homo_hook_collect(void* user_param, uint16_t const nr, uint16_t const* map) {
...
   out->degree = PERM_DEGREE;
   DIGRAPHS_ASSERT(out->capacity >= LEN_LIST(o));
+  Int k = 0;  // perm coll index
   for (Int i = 1; i <= LEN_LIST(o); ++i) {
     DIGRAPHS_ASSERT(ISB_LIST(o, i));
     DIGRAPHS_ASSERT(IS_PERM2(ELM_LIST(o, i)) || IS_PERM4(ELM_LIST(o, i)));
     Obj p = ELM_LIST(o, i);
     DIGRAPHS_ASSERT(LargestMovedPointPerm(p) <= PERM_DEGREE);
     if (LargestMovedPointPerm(p) == 0) {
+      // the index k must be separate from the index i because of this continue
```
wilfwilson, line 312: Is `k` supposed to be modified by this loop? Because as far as I can tell it isn't, and so I don't see the point of `k` because below, you're always doing:
```
Perm q = out->perms[0];
out->size++;
DIGRAPHS_ASSERT(out->size == 1);
```

src/homos.c
```diff
@@ -292,23 +293,28 @@ homo_hook_collect(void* user_param, uint16_t const nr, uint16_t const* map) {
...
       continue;
     }
-    Perm q = out->perms[i - 1];
+    Perm q = out->perms[k++];
```
james-d-mitchell, line 315: `k` is modified right here!

src/homos.c
james-d-mitchell: Thanks @wilfwilson `k` is modified though, in line 315.

wilfwilson: Duh, silly me. My internal C parser is broken...

I was thinking that the `}` brace on line 314 was ending the `for` loop, which meant that line 315 was outside the loop! I must be tired...

Issue #374, opened 2020-12-18
Close files when erroring


wilfwilson:
Inspired by https://github.com/semigroups/Semigroups/pull/633 we should make sure that we properly close open files when we exit a method earlier than usual because of an error.
```
gap> IO.OpenFiles;
[  ]
gap> DigraphsTestStandard();
gap> IO.OpenFiles;
[ <file fd=25 rbufsize=65536 rpos=1 rdata=0>, <file fd=24 rbufsize=65536 rpos=
    6 rdata=44>, <file fd=18 rbufsize=65536 rpos=1 rdata=0>,
  <file fd=23 rbufsize=65536 rpos=9 rdata=41>,
  <file fd=22 rbufsize=65536 rpos=1 rdata=0>, <file fd=17 rbufsize=65536 rpos=
    6 rdata=44>, <file fd=21 wbufsize=65536 wdata=0>,
  <file fd=16 rbufsize=65536 rpos=9 rdata=41>,
  <file fd=20 rbufsize=65536 rpos=1 rdata=0>, <file fd=15 rbufsize=65536 rpos=
    1 rdata=0>, <file fd=19 wbufsize=65536 wdata=0>,
  <file fd=14 wbufsize=65536 wdata=0>, <file fd=13 rbufsize=65536 rpos=
    1 rdata=0>, <file fd=12 wbufsize=65536 wdata=0>,
  <file fd=11 rbufsize=65536 rpos=1 rdata=0>, <file fd=10 rbufsize=65536 rpos=
    6 rdata=44>, <file fd=9 rbufsize=65536 rpos=9 rdata=41>,
  <file fd=8 rbufsize=65536 rpos=1 rdata=0>, <file fd=7 wbufsize=65536 wdata=
    0>, <file fd=6 rbufsize=65536 rpos=1 rdata=0>,
  <file fd=4 wbufsize=65536 wdata=0> ]
```
Running the tests should not leave any newly-opened files.

james-d-mitchell:
Resolved in v1.4.0

PR #377, opened 2020-12-24
io: always close files (resolves Issue #374)

james-d-mitchell:
This resolves issue #374, there was one place where we failed to close a file in `io.gi` the other unclosed files arose from carelessness in the tests.


wilfwilson (approved):
Brilliant, thanks.

Issue #366, opened 2020-12-16
CayleyDigraph can return wrong answer


wilfwilson:
This was discovered by @jdebeule (see #363). Thanks for spotting this. For example:
```
gap> G := DihedralGroup(8);;
gap> gens := GeneratorsOfGroup(G);;
gap> elts := AsList(G);;
gap> D := CayleyDigraph(G, gens);;
gap> Filtered(DigraphEdges(D), edge ->
> not ForAny(gens, x -> elts[edge[1]] * x = elts[edge[2]]));
[ [ 2, 8 ], [ 3, 5 ], [ 5, 3 ], [ 5, 2 ], [ 6, 5 ], [ 7, 8 ], [ 8, 7 ], [ 8, 6 ] ]
```
This is incorrect: the pair `[i,j]` is an edge of the Cayley digraph if and only if the `i`th element multiplied by some generator is the `j`th element. So the final command should return an empty list.

james-d-mitchell:
Thanks for the report @wilfwilson, am I correct in thinking that this can be resolved by replacing the use of `OnRight` in line 110 of `grape.gi` with `OnLeftInverse`?

jdebeule:
Dear James, dear Wilf, yes, it is sufficient to replace the action function OnRight by OnLeftInverse. I am not sure whether I did the pull request correctly, but exactly this correction is in the pull request, together with some other things.

wilfwilson:
Yeah @jdebeule's PR makes this change, I created the issue to help us keep track of the bug. And I'll happily look at PR #378 and merge the fix as appropriate (even though Jan's PR has it too) because it's good to have the bug fixed, and I've not got around to reviewing Jan's slightly larger PR yet.

james-d-mitchell:
Resolved in v1.4.0

PR #378, opened 2020-12-24
Resolve Issue #366

james-d-mitchell:
This fixes issue #366 by replacing `OnRight` by `OnLeftInverse` and updating the test output and manual examples accordingly.


wilfwilson:
Hi James, I think you've done `commit --amend` when checked out at the commit 'Resolve Issue #367'. So the commit in this PR contains both the fix for issue #367 (which has already been merged already into `stable-1.3`) and the fix for #366. Would you mind sorting that out and using the right commit message? Thanks.

wilfwilson:
Rebasing on `stable-1.3` and renaming the commit should suffice!

wilfwilson:
Actually I can just do it directly thanks to the magic of GitHub 🙂

wilfwilson (approved):

PR #379, opened 2020-12-28
Add VerticesReachableFrom

marinaanagno:


gap/oper.gi
```diff
@@ -1725,6 +1725,36 @@ function(D, list)
   return DigraphShortestDistance(D, list[1], list[2]);
 end);
 
+InstallMethod(IsReachableFrom, "for a digraph and a vertex",
```
wilfwilson, line 1728: I'd suggest a different name. In GAP, I expect `Is...` functions to be things that return either `true` or `false`. Something more appropriate could be:
* `ReachableFrom`
* `VerticesReachableFrom`

marinaanagno: Thanks @wilfwilson! Did that now! :)

doc/oper.xml
```diff
@@ -1300,6 +1300,40 @@ false
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="VerticesReachableFrom">
+<ManSection>
+  <Oper Name="VerticesReachableFrom" Arg="digraph, root"/>
+  <Returns>A list.</Returns>
+  <Description>
+    This operation returns a list of the vertices  <A>v</A> , for which there exists
+    a non-trivial directed walk from vertex <A>root</A> to vertex <A>v</A> in the digraph
```
wilfwilson, line 1309: That's not quite true - the function currently always returns the root to be reachable from the root, even if it is only reachable by a *trivial* directed walk, e.g.:
```
gap> D := Digraph([[2], []]);;
gap> VerticesReachableFrom(D, 1);
[ 1, 2 ]
```
I think the currently-documented behaviour is the right behaviour, so please update the code to match it 🙂

doc/oper.xml
```diff
@@ -1300,6 +1300,40 @@ false
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="VerticesReachableFrom">
+<ManSection>
+  <Oper Name="VerticesReachableFrom" Arg="digraph, root"/>
+  <Returns>A list.</Returns>
+  <Description>
+    This operation returns a list of the vertices  <A>v</A> , for which there exists
```
wilfwilson, line 1308:
```suggestion
    This operation returns a list of the vertices  <A>v</A>, for which there exists
```

doc/oper.xml
```diff
@@ -1300,6 +1300,40 @@ false
...
+    a non-trivial directed walk from vertex <A>root</A> to vertex <A>v</A> in the digraph
+    <A>digraph</A>. See section <Ref Subsect="Definitions" Style="Number" />
```
wilfwilson, line 1310:
```suggestion
    <A>digraph</A>. See Section <Ref Subsect="Definitions" Style="Number" />
```

doc/oper.xml
```diff
@@ -1300,6 +1300,40 @@ false
...
+    for the definition of anon-trivial directed walk.
```
wilfwilson, line 1311:
```suggestion
    for the definition of a non-trivial directed walk.
```

wilfwilson (commented):
I've added a few suggestions - I haven't done a full review, I've only looked at the documentation, but not the code.

doc/oper.xml
marinaanagno: Thank you @wilfwilson ! I've fixed this, but for some reason if I try to just remove the root vertex from the list of visited vertices and start off with an empty list instead of a list containing  the root the function does not return output (probably runs forever for some reason). I had this problem in the beginning and that is basically why I included the root. Still haven't found what I'm missing. I removed the root at the end when we have performed the DFS, but I guess this is not the best way to do it :)

wilfwilson: Yeah it might be fiddly! The problem with removing the root at the end is that that's not always a valid thing to do. Indeed the root is quite often reachable from itself via a *non-trivial* walk. For instance, with `D := CompleteDigraph(2)`, then `VerticesReachableFrom(D, 1)` should be `[ 1, 2 ]` because `1 → 2 → 1` and `1 → 2` are non-trivial walks from 1 to 1 and 1 to 2 respectively.

marinaanagno: oh yes, right! That doesn't solve the problem! :(
I'll try to find some other way of doing this, but for the time being I can't really see what causes this.

marinaanagno: @wilfwilson my thinking might be completely wrong, but since we execute a depth first search to find which vertices are reachable from a root vertex, isn't the only case we meet the root vertex the start of the process, despite other possible non-trivial paths? Also I did a google search and reachability is mentioned to be reflexive, due to the trivial path of zero length, so maybe I need to change documentation instead, or if we're also interested in non-trivial cases make a second function? :)

wilfwilson: I understand what you're saying. 🙂

I'm sure there are some sources (perhaps even most?) that consider each vertex to be reachable from itself. Mathematically, that is probably the cleanest definition. But when writing a software package, as well as the purity of the mathematics, we have to think about:

1. what is the most *useful* thing to do? and
2. does our package behave consistently?

To the first point, it's surely the most useful to have the definition with a non-trivial walk, since that gives you *extra* information. If a user prefers this definition, then our function works perfectly for them; if a user prefers the definition with an arbitrary walk, then our function can work fine for them - they just have to remember to add the root itself to the result if it is not already in there. But you can't go the other way: if `VerticesReachableFrom(D, v)` *always* includes `v`, then there is no way to recover the information about whether or not `v` is reachable from itself by a non-trivial walk! So the function is not very useful to someone who only wants to know about non-trivial walks.

For the second point, we already have a function `IsReachable(D, u, v)` which (according to its documentation) only returns `true` if there's a non-trivial walk in `D` from `u` to `v:
```
gap> D := ChainDigraph(5);; IsReachable(D, 1, 1);
false
gap> D := CompleteDigraph(5);; IsReachable(D, 1, 1);
true
```
So we should remain consistent with the behaviour of this function, which already exists in previously published versions of the package.

> since we execute a depth first search to find which vertices are reachable from a root vertex, isn't the only case we meet the root vertex the start of the process, despite other possible non-trivial paths

No. In a depth-first search from a root `v`, we follow every edge that leaves any vertex that's reachable from `v` (including `v` itself). (By definition, these edges all point towards vertices that are also reachable from `v`). When we follow an edge, either it leads to a new vertex, which is great, otherwise it leads to a vertex that we've seen before - in which case we immediately backtrack. So we typically meet each vertex (including the root vertex, if it is reachable by a non-trivial walk) several times during the course of the search. It's just that we backtrack immediately for all but the first meeting.

I will comment on the code about how you can fix the code.

gap/oper.gi
```diff
@@ -1725,6 +1725,37 @@ function(D, list)
...
+  succ := OutNeighbours(D);
+  visited := [root];
+  parent := [];
+  parent[root] := fail;
+  repeat
+    prev := current;
+    for i in [index[current] .. Length(succ[current])] do
+      n := succ[current][i];
+      if not (n in visited) then
+        Add(visited, n);
```
wilfwilson, line 1744: The test in the `if` statement should be a lot faster if you do this:
```suggestion
      if not (n in visited) then
        AddSet(visited, n);
```
That's because a set is a *sorted* list in GAP, and GAP can use binary search with complexity `log n` to search in a sorted list of length `n`, otherwise it uses a linear search with complexity `n`.

gap/oper.gi
```diff
@@ -1725,6 +1725,37 @@ function(D, list)
...
 
+InstallMethod(VerticesReachableFrom, "for a digraph and a vertex",
+[IsDigraph, IsPosInt],
+function(D, root)
+  local N, index, current, succ, visited, prev, n, i, parent;
+  N := DigraphNrVertices(D);
+  index := ListWithIdenticalEntries(N, 1);
+  current := root;
+  succ := OutNeighbours(D);
+  visited := [root];
```
wilfwilson, line 1736:
```suggestion
  visited := [];
```

gap/oper.gi
```diff
@@ -1725,6 +1725,37 @@ function(D, list)
...
+        index[current] := i + 1;
+        current := n;
+        break;
+      fi;
+    od;
+    if prev = current then
+      current := parent[current];
+    fi;
+  until current = fail;
+  Remove(visited, 1);
```
wilfwilson, line 1755:
```suggestion
```

gap/oper.gi
```diff
@@ -1725,6 +1725,37 @@ function(D, list)
...
+  repeat
+    prev := current;
+    for i in [index[current] .. Length(succ[current])] do
+      n := succ[current][i];
+      if not (n in visited) then
+        Add(visited, n);
+        parent[n] := current;
+        index[current] := i + 1;
+        current := n;
+        break;
```
wilfwilson, line 1748:
```suggestion
        if n <> root then
          parent[n] := current;
          index[current] := i + 1;
          current := n;
          break;
        fi;
```

wilfwilson (commented):
I think these tweaks to the code will make it work properly 🙂

doc/oper.xml
marinaanagno: Thank you so much @wilfwilson that makes absolute sense! :)

wilfwilson (dismissed):
Looks good to me 🙂

marinaanagno:
Thanks for all this help @wilfwilson! :)

gap/oper.gi
```diff
@@ -1725,6 +1725,38 @@ function(D, list)
...
 
+InstallMethod(VerticesReachableFrom, "for a digraph and a vertex",
+[IsDigraph, IsPosInt],
+function(D, root)
+  local N, index, current, succ, visited, prev, n, i, parent;
+  N := DigraphNrVertices(D);
+  index := ListWithIdenticalEntries(N, 1);
+  current := root;
+  succ := OutNeighbours(D);
+  visited := [];
```
james-d-mitchell, line 1736: I think `visited` is unnecessary. If `index` is initialised to all `0`, then we can check `if index[n] <> 0` in line 1743. This will use a bit less space, and will also be quicker because the check `n in visited` has complexity O(log2(n)) where n is the length of `visited` but the suggested check has constant time.

james-d-mitchell (changes requested):
Just one change and I'm happy for this to be merged. Thanks so much @marinaanagno, great work!

gap/oper.gi
james-d-mitchell: What I wrote above isn't really right, of course, `visited` is required because it's the return value of the function. So, my suggested changes are as indicated.

gap/oper.gi
```diff
@@ -1725,6 +1725,38 @@ function(D, list)
   return DigraphShortestDistance(D, list[1], list[2]);
 end);
 
+InstallMethod(VerticesReachableFrom, "for a digraph and a vertex",
+[IsDigraph, IsPosInt],
+function(D, root)
+  local N, index, current, succ, visited, prev, n, i, parent;
+  N := DigraphNrVertices(D);
+  index := ListWithIdenticalEntries(N, 1);
```
james-d-mitchell, line 1733:
```suggestion
  index := ListWithIdenticalEntries(N, 0);
```

gap/oper.gi
```diff
@@ -1725,6 +1725,38 @@ function(D, list)
...
+  current := root;
+  succ := OutNeighbours(D);
+  visited := [];
+  parent := [];
+  parent[root] := fail;
+  repeat
+    prev := current;
+    for i in [index[current] .. Length(succ[current])] do
+      n := succ[current][i];
+      if not (n in visited) then
```
james-d-mitchell, line 1743:
```suggestion
      if index[n] = 0 then
```

gap/oper.gi
```diff
@@ -1725,6 +1725,38 @@ function(D, list)
...
+        AddSet(visited, n);
```
james-d-mitchell, line 1744:
```suggestion
        Add(visited, n);
```

gap/oper.gi
```diff
@@ -1725,6 +1725,38 @@ function(D, list)
...
+        if n <> root then
```
james-d-mitchell, line 1745:
```suggestion
```

gap/oper.gi
```diff
@@ -1725,6 +1725,38 @@ function(D, list)
...
+          parent[n] := current;
+          index[current] := i + 1;
+          current := n;
+          break;
+        fi;
```
james-d-mitchell, line 1750:
```suggestion
  
```

gap/oper.gi
```diff
@@ -1725,6 +1725,38 @@ function(D, list)
...
+  repeat
+    prev := current;
+    for i in [index[current] .. Length(succ[current])] do
+      n := succ[current][i];
+      if not (n in visited) then
+        AddSet(visited, n);
+        if n <> root then
+          parent[n] := current;
+          index[current] := i + 1;
+          current := n;
```
james-d-mitchell, line 1748:
```suggestion
          current := n;
          index[current] := 1;
```

gap/oper.gi
james-d-mitchell, line 1736:
```suggestion
  visited := [root];
```

james-d-mitchell: I'm not 100% if you want the root in the list or not, so please take or leave this suggestion as appropriate

james-d-mitchell (commented):
I've suggested explicitly the changes I thought of, my comment of yesterday was not completely accurate, sorry!

gap/oper.gi
wilfwilson: I've argued above that the root should not be in the returned if and only if (as documented) it is reachable from itself by a **non-trivial** walk. That explained partly why the code was structured as it was.

wilfwilson: This was my reasoning to @marinaanagno https://github.com/digraphs/Digraphs/pull/379/files#r553562550

james-d-mitchell: Yup, makes sense to me too, looks to me like @marinaanagno used this definition

gap/oper.gi
```diff
@@ -1725,6 +1725,42 @@ function(D, list)
...
+  current := root;
+  succ := OutNeighbours(D);
+  visited := [];
+  parent := [];
+  parent[root] := fail;
+  repeat
+    prev := current;
+    for i in [index[current] .. Length(succ[current])] do
+      n := succ[current][i];
+      if n = root and have_visited_root = false then
```
james-d-mitchell, line 1745:
```suggestion
      if n = root and not have_visited_root then
```

james-d-mitchell (approved):

PR #380, opened 2021-01-07
Correct some capitalisation in the manual

wilfwilson:
We should consistently refer to sections as `Section <nr>` rather than `section <nr>`.


james-d-mitchell (approved):
Good

PR #382, opened 2021-01-12
Add Lawler and Byskov's dynamic programming algorithms for chromatic number

EwanGilligan:
This leaves the current attribute for chromatic number in place, but adds an alternative operation that takes an object which determines which algorithm to use. These algorithms are currently slower than the current implementation, but have space for later optimisation (being able to specify a bound for the maximum size of independent sets would be useful in this regard), but serve as a good baseline for later algorithms that will be added.

As part of the method selection, the files `algobj.gi` and `algobj.gd` were added. The purpose of these files is to contain objects that relate to method selection for algorithms, in this case it houses the objects used to select the colouring algorithm. Alongside this are a number of categories and filters, such as `IsDigraphAlgorithm` which serves as a general category for objects related to operation on Digraphs.

### DIFF
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -1417,7 +1417,17 @@ gap> DigraphLoops(D);
     least one vertex, <C>ChromaticNumber</C> returns the least number <C>n</C>
     such that <C>DigraphColouring(<A>digraph</A>, n)</C> does not return
     <K>fail</K>. See <Ref Oper="DigraphColouring"
-      Label="for a digraph and a number of colours"/>.
+      Label="for a digraph and a number of colours"/>. <P/>
+
+    It is possible to select the algorithm to compute the chromatic number via
+    the use of value options. The permitted algorithms and values to
+    pass as options are:
+    <List>
+      <Item><C>lawler</C> - Lawler's Algorithm 
+      <Cite Key="Law1976"/></Item>
+      <Item><C>byskov</C> - Byskov's Algorithm 
+      <Cite Key="Bys2002"/></Item>
+    </List>
 
     <Example><![CDATA[
 gap> ChromaticNumber(NullDigraph(10));
@@ -1434,6 +1444,10 @@ gap> D := PetersenGraph(IsMutableDigraph);
 <mutable digraph with 10 vertices, 30 edges>
 gap> ChromaticNumber(D);
 3
+gap> ChromaticNumber(CompleteDigraph(10) : lawler);
+10
+gap> ChromaticNumber(CompleteDigraph(10) : byskov);
+10
 ]]></Example>
   </Description>
 </ManSection>
--- a/gap/attr.gi
+++ b/gap/attr.gi
@@ -174,6 +174,201 @@ function(D)
   return DIGRAPHS_ArticulationPointsBridgesStrongOrientation(D)[4];
 end);
 
+# Utility function to calculate the maximal independent sets (as BLists) of a
+# subgraph induced by removing a set of vertices.
+BindGlobal("DIGRAPHS_MaximalIndependentSetsSubtractedSet",
+function(I, subtracted_set, size_bound)
+  local induced_mis, temp, i;
+  # First remove all vertices in the subtracted set from each MIS
+  temp := List(I, i -> DifferenceBlist(i, subtracted_set));
+  induced_mis := [];
+  # Then remove any sets which are no longer maximal
+  # Sort in decreasing size
+  Sort(temp, {x, y} -> SizeBlist(x) > SizeBlist(y));
+  # Then check elements from back to front for if they are a subset
+  for i in temp do
+    if SizeBlist(i) <= size_bound and
+        ForAll(induced_mis, x -> not IsSubsetBlist(x, i)) then
+      Add(induced_mis, i);
+    fi;
+  od;
+  return induced_mis;
+end
+);
+
+BindGlobal("DIGRAPHS_ChromaticNumberLawler",
+function(D)
+  local n, vertices, subset_colours, s, S, i, I, subset_iter, x,
+  mis, subset_mis;
+  n := DigraphNrVertices(D);
+  vertices := List(DigraphVertices(D));
+  # Store all the Maximal Independent Sets, which can later be used for
+  # calculating the maximal independent sets of induced subgraphs.
+  mis := DigraphMaximalIndependentSets(D);
+  # Convert each MIS to a Blist
+  mis := List(mis, x -> BlistList(vertices, x));
+  # Store current best colouring for each subset
+  subset_colours := ListWithIdenticalEntries(2 ^ n, infinity);
+  # Empty set can be colouring with only one colour.
+  subset_colours[1] := 0;
+  # Iterator for blist subsets.
+  subset_iter := ListWithIdenticalEntries(n, [false, true]);
+  subset_iter := IteratorOfCartesianProduct2(subset_iter);
+  # Skip the first one, which should be the empty set.
+  S := NextIterator(subset_iter);
+  # Iterate over all vertex subsets.
+  for S in subset_iter do
+    # Cartesian iterator is ascending lexicographically, but we want reverse
+    # lexicographic ordering. We treat this as iterating over the complement.
+    # Index the current subset that is being iterated over.
+    s := 1;
+    for x in [1 .. n] do
+      # Need to negate, as we are iterating over the complement.
+      if not S[x] then
+        s := s + 2 ^ (x - 1);
+      fi;
+    od;
+    # Iterate over the maximal independent sets of V[S]
+    subset_mis := DIGRAPHS_MaximalIndependentSetsSubtractedSet(mis, S, infinity);
+    # Flip the list, as we now need the actual set.
+    FlipBlist(S);
+    for I in subset_mis do
+        # Calculate S \ I. This is destructive, but is undone.
+        SubtractBlist(S, I);
+        # Index S \ I
+        i := 1;
+        for x in [1 .. n] do
+          if S[x] then
+            i := i + 2 ^ (x - 1);
+          fi;
+        od;
+        # The chromatic number of this subset is the minimum value of all
+        # the maximal independent subsets of D[S].
+        subset_colours[s] := Minimum(subset_colours[s], subset_colours[i] + 1);
+        # Undo the changes to the subset.
+        UniteBlist(S, I);
+    od;
+  od;
+  return subset_colours[2 ^ n];
+end
+);
+
+BindGlobal("DIGRAPHS_UnderThreeColourable",
+function(D)
+  local nr;
+  nr := DigraphNrVertices(D);
+  if DigraphHasLoops(D) then
+    ErrorNoReturn("the argument <D> must be a digraph with no loops,");
+  elif nr = 0 then
+    return 0;  # chromatic number = 0 iff <D> has 0 verts
+  elif IsNullDigraph(D) then
+    return 1;  # chromatic number = 1 iff <D> has >= 1 verts & no edges
+  elif IsBipartiteDigraph(D) then
+    return 2;  # chromatic number = 2 iff <D> has >= 2 verts & is bipartite
+               # <D> has at least 2 vertices at this stage
+  elif DigraphColouring(D, 3) <> fail then
+               # Check if there is a 3 colouring
+    return 3;
+  fi;
+  return infinity;
+end
+);
+
+BindGlobal("DIGRAPHS_ChromaticNumberByskov",
+function(D)
+  local n, a, vertices, subset_colours, S, i, j, I, subset_iter,
+  index_subsets, vertex_blist, k, MIS;
+
+  n := DigraphNrVertices(D);
+  vertices := DigraphVertices(D);
+  vertex_blist := BlistList(vertices, vertices);
+  # Store all the Maximal Independent Sets, which can later be used for
+  # calculating the maximal independent sets of induced subgraphs.
+  MIS := DigraphMaximalIndependentSets(D);
+  # Convert each MIS to a Blist
+  MIS := List(MIS, x -> BlistList(vertices, x));
+  # Store current best colouring for each subset
+  subset_colours := ListWithIdenticalEntries(2 ^ n, infinity);
+  # Empty set is 0 colourable
+  subset_colours[1] := 0;
+  # Function to index the subsets of the vertices of D
+  index_subsets := function(subset)
+    local x, index;
+    index := 1;
+    for x in [1 .. n] do
+      if subset[x] then
+        index := index + 2 ^ (x - 1);
+      fi;
+    od;
+    return index;
+  end;
+  # Iterate over vetex subsets
+  subset_iter := IteratorOfCombinations(vertices);
+  # Skip the first one, which should be the empty set
+  S := NextIterator(subset_iter);
+  Assert(1, IsEmpty(S), "First set from iterator should be the empty set");
+  # First find the 3 colourable subgraphs of D
+  for S in subset_iter do
+    a := DIGRAPHS_UnderThreeColourable(InducedSubdigraph(D, S));
+    S := BlistList(vertices, S);
+    i := index_subsets(S);
+    # Mark this as three or less colourable if it is.
+    subset_colours[i] := Minimum(a, subset_colours[i]);
+  od;
+  # Process 4 colourable subgraphs
+  for I in MIS do
+    SubtractBlist(vertex_blist, I);
+    # Iterate over all subsets of V(D) \ I as blists
+    # This is done by taking the cartesian product of n copies of [true, false]
+    # or [true] if the vertex is in I. The [true] is used as each element will
+    # be flipped to get reverse lexicographic ordering.
+    subset_iter := EmptyPlist(n);
+    for i in [1 .. n] do
+      if I[i] then
+        subset_iter[i] := [true];
+      else
+        subset_iter[i] := [true, false];
+      fi;
+    od;
+    subset_iter := IteratorOfCartesianProduct2(subset_iter);
+    # Skip the empty set.
+    NextIterator(subset_iter);
+    for S in subset_iter do
+      FlipBlist(S);
+      i := index_subsets(S);
+      if subset_colours[i] = 3 then
+        # Index union of S and I
+        j := index_subsets(UnionBlist(S, I));
+        subset_colours[j] := Minimum(subset_colours[j], 4);
+      fi;
+    od;
+    # Undo the changes made.
+    UniteBlist(vertex_blist, I);
+  od;
+  # Iterate over vetex subset blists.
+  subset_iter := ListWithIdenticalEntries(n, [true, false]);
+  subset_iter := IteratorOfCartesianProduct2(subset_iter);
+  # Skip the first one, which should be the empty set
+  S := NextIterator(subset_iter);
+  for S in subset_iter do
+    # Cartesian iteratator goes in lexicographic order, but we want reverse.
+    FlipBlist(S);
+    # Index the current subset that is being iterated over
+    i := index_subsets(S);
+    if 4 <= subset_colours[i] and subset_colours[i] < infinity then
+      k := SizeBlist(S) / subset_colours[i];
+      # Iterate over the maximal independent sets of D[V \ S]
+      for I in DIGRAPHS_MaximalIndependentSetsSubtractedSet(MIS, S, k) do
+        # Index S union I
+        j := index_subsets(UnionBlist(S, I));
+        subset_colours[j] := Minimum(subset_colours[j], subset_colours[i] + 1);
+      od;
+    fi;
+  od;
+  return subset_colours[2 ^ n];
+end
+);
+
 InstallMethod(ChromaticNumber, "for a digraph by out-neighbours",
 [IsDigraphByOutNeighboursRep],
 function(D)
@@ -190,6 +385,13 @@ function(D)
   elif IsBipartiteDigraph(D) then
     return 2;  # chromatic number = 2 iff <D> has >= 2 verts & is bipartite
                # <D> has at least 2 vertices at this stage
+  fi;
+
+  # Value option for alternative dispatch
+  if ValueOption("lawler") <> fail then
+    return DIGRAPHS_ChromaticNumberLawler(D);
+  elif ValueOption("byskov") <> fail then
+    return DIGRAPHS_ChromaticNumberByskov(D);
   fi;
 
   # The chromatic number of <D> is at least 3 and at most nr
--- a/init.g
+++ b/init.g
@@ -37,6 +37,13 @@ fi;
 if not IsBound(Adjacency) then
   Adjacency := IdFunc;
 fi;
+# Avoid warnings with GAP 4.10 for use of FlipBlist
+# Can be removed once Digraphs requires GAP>=4.11
+if not IsBound(FlipBlist) then
+  FlipBlist := function(BList)
+    Apply(BList, x -> not x);
+  end;
+fi;
 
 ReadPackage("digraphs", "gap/digraph.gd");
 ReadPackage("digraphs", "gap/digraphs.g");
--- a/tst/standard/attr.tst
+++ b/tst/standard/attr.tst
@@ -1548,6 +1548,112 @@ gap> D := Digraph(OutNeighbours(CycleDigraph(13)));;
 gap> ChromaticNumber(D);
 3
 
+#  Test ChromaticNumber Lawler
+gap> ChromaticNumber(NullDigraph(10) : lawler);
+1
+gap> ChromaticNumber(CompleteDigraph(10) : lawler);
+10
+gap> ChromaticNumber(CompleteBipartiteDigraph(5, 5) : lawler);
+2
+gap> ChromaticNumber(DigraphRemoveEdge(CompleteDigraph(10), [1, 2]) : lawler);
+10
+gap> ChromaticNumber(Digraph([[4, 8], [6, 10], [9], [2, 3, 9], [],
+> [3], [4], [6], [], [5, 7]]) : lawler);
+3
+gap> ChromaticNumber(DigraphDisjointUnion(CompleteDigraph(1),
+> Digraph([[2], [4], [1, 2], [3]])) : lawler);
+3
+gap> ChromaticNumber(DigraphDisjointUnion(CompleteDigraph(1),
+> Digraph([[2], [4], [1, 2], [3], [1, 2, 3]])) : lawler);
+4
+gap> gr := Digraph([[2, 3, 4], [3], [], []]);
+<immutable digraph with 4 vertices, 4 edges>
+gap> ChromaticNumber(gr : lawler);
+3
+gap> ChromaticNumber(EmptyDigraph(0) : lawler);
+0
+gap> gr := CompleteDigraph(4);;
+gap> gr := DigraphAddVertex(gr);;
+gap> ChromaticNumber(gr : lawler);
+4
+gap> gr := Digraph([[2, 4, 7, 3], [3, 5, 8, 1], [1, 6, 9, 2],
+> [5, 7, 1, 6], [6, 8, 2, 4], [4, 9, 3, 5], [8, 1, 4, 9], [9, 2, 5, 7],
+> [7, 3, 6, 8]]);;
+gap> ChromaticNumber(gr : lawler);
+3
+gap> gr := DigraphSymmetricClosure(ChainDigraph(5));
+<immutable symmetric digraph with 5 vertices, 8 edges>
+gap> ChromaticNumber(gr : lawler);
+2
+gap> gr := DigraphFromGraph6String("KmKk~K??G@_@");
+<immutable symmetric digraph with 12 vertices, 42 edges>
+gap> ChromaticNumber(gr : lawler);
+4
+gap> gr := CycleDigraph(7);
+<immutable cycle digraph with 7 vertices>
+gap> ChromaticNumber(gr : lawler);
+3
+gap> ChromaticNumber(gr : lawler);
+3
+gap> ChromaticNumber(gr : lawler);
+3
+
+#  Test ChromaticNumber Byskov 
+gap> ChromaticNumber(NullDigraph(10) : byskov);
+1
+gap> ChromaticNumber(CompleteDigraph(10) : byskov);
+10
+gap> ChromaticNumber(CompleteBipartiteDigraph(5, 5) : byskov);
+2
+gap> ChromaticNumber(DigraphRemoveEdge(CompleteDigraph(10), [1, 2]) : byskov);
+10
+gap> ChromaticNumber(Digraph([[4, 8], [6, 10], [9], [2, 3, 9], [],
+> [3], [4], [6], [], [5, 7]]) : byskov);
+3
+gap> ChromaticNumber(DigraphDisjointUnion(CompleteDigraph(1),
+> Digraph([[2], [4], [1, 2], [3]])) : byskov);
+3
+gap> ChromaticNumber(DigraphDisjointUnion(CompleteDigraph(1),
+> Digraph([[2], [4], [1, 2], [3], [1, 2, 3]])) : byskov);
+4
+gap> gr := Digraph([[2, 3, 4], [3], [], []]);
+<immutable digraph with 4 vertices, 4 edges>
+gap> ChromaticNumber(gr : byskov);
+3
+gap> ChromaticNumber(EmptyDigraph(0) : byskov);
+0
+gap> gr := CompleteDigraph(4);;
+gap> gr := DigraphAddVertex(gr);;
+gap> ChromaticNumber(gr : byskov);
+4
+gap> gr := Digraph([[2, 4, 7, 3], [3, 5, 8, 1], [1, 6, 9, 2],
+> [5, 7, 1, 6], [6, 8, 2, 4], [4, 9, 3, 5], [8, 1, 4, 9], [9, 2, 5, 7],
+> [7, 3, 6, 8]]);;
+gap> ChromaticNumber(gr : byskov);
+3
+gap> gr := DigraphSymmetricClosure(ChainDigraph(5));
+<immutable symmetric digraph with 5 vertices, 8 edges>
+gap> ChromaticNumber(gr : byskov);
+2
+gap> gr := DigraphFromGraph6String("KmKk~K??G@_@");
+<immutable symmetric digraph with 12 vertices, 42 edges>
+gap> ChromaticNumber(gr : byskov);
+4
+gap> gr := CycleDigraph(7);
+<immutable cycle digraph with 7 vertices>
+gap> ChromaticNumber(gr : byskov);
+3
+gap> ChromaticNumber(gr : byskov);
+3
+gap> ChromaticNumber(gr : byskov);
+3
+
+# Extra tests for under three colourable check
+gap> DIGRAPHS_UnderThreeColourable(Digraph([[1]]));
+Error, the argument <D> must be a digraph with no loops,
+gap> DIGRAPHS_UnderThreeColourable(EmptyDigraph(0));
+0
+
 #  DegreeMatrix
 gap> gr := Digraph([[2, 3, 4], [2, 5], [1, 5, 4], [1], [1, 1, 2, 4]]);;
 gap> DegreeMatrix(gr);
--- a/doc/digraphs.bib
+++ b/doc/digraphs.bib
@@ -163,3 +163,23 @@
   Year = {1967},
   Bdsk-Url-1 = {https://dx.doi.org/10.1093/comjnl/10.1.85}
 }
+
+@article{Law1976,
+    Title={A Note on the Complexity of the Chromatic Number Problem},
+    Author={E. Lawler},
+    Journal={Inf. Process. Lett.},
+    Year={1976},
+    Volume={5},
+    Pages={66-67}
+}
+
+@article{Bys2002,
+    Author = {Byskov, Jesper},
+    Year = {2002},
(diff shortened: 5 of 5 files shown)


gap/attr.gi
```diff
@@ -290,6 +290,177 @@ function(D)
...
+  NextIterator(subset_iter);
+  # Iterate over all vertex subsets.
+  for s in subset_iter do
+    # Index the current subset that is being iterated over.
+    S := Sum(s, x -> 2 ^ (x - 1)) + 1;
+    x[S] := infinity;
+    # Get the subgraph induced by the vertex subset.
+    induced_subgraph := InducedSubdigraph(D, s);
+    # Iterate over the maximal independent sets of D[S]
+    for I in DigraphMaximalIndependentSets(induced_subgraph) do
```
mtorpey, line 326: Hopefully we don't need to create `induced_subgraph` here, because we can do `DigraphMaximalIndependentSets(D, [], complement_of_s);` or something like that.

gap/attr.gi
```diff
@@ -290,6 +290,177 @@ function(D)
...
+  elif IsNullDigraph(D) then
+    return 1;  # chromatic number = 1 iff <D> has >= 1 verts & no edges
+  elif IsBipartiteDigraph(D) then
+    return 2;  # chromatic number = 2 iff <D> has >= 2 verts & is bipartite
+               # <D> has at least 2 vertices at this stage
+  fi;
+
+  vertices := DigraphVertices(D);
+  # Store current best colouring for each subset
+  x := [1 .. 2 ^ n];
```
mtorpey, line 312: `ListWithIdenticalEntries(infinity, 2^n)` might be quicker

mtorpey: Better name than `x` while we're at it?

gap/attr.gi
```diff
@@ -290,6 +290,177 @@ function(D)
...
+  # Empty set can be colouring with only one colour.
+  x[1] := 0;
+  subset_iter := IteratorOfCombinations(vertices);
+  # Skip the first one, which should be the empty set.
+  NextIterator(subset_iter);
+  # Iterate over all vertex subsets.
+  for s in subset_iter do
+    # Index the current subset that is being iterated over.
+    S := Sum(s, x -> 2 ^ (x - 1)) + 1;
```
mtorpey, line 321: Could we use a loop counter instead of this sum?  Yes, if `NextIterator` gives us a consistent ordering.  Look this up?

If not, then this isn't too bad.

gap/attr.gi
```diff
@@ -290,6 +290,177 @@ function(D)
...
+    # Get the subgraph induced by the vertex subset.
+    induced_subgraph := InducedSubdigraph(D, s);
+    # Iterate over the maximal independent sets of D[S]
+    for I in DigraphMaximalIndependentSets(induced_subgraph) do
+        # Create a clone of the subset that we can remove the current
+        # independent set from.
+        s_without_I := ShallowCopy(s);
+        SubtractSet(s_without_I, SetX(I,
+        # Need to relable the independent set back to the original labels.
+        x -> DigraphVertexLabel(induced_subgraph, x)));
```
mtorpey, line 332: Get `DigraphVertexLabels(induced_subgraph)` and store them in a variable, then use that variable.

mtorpey: Also indent this line more, since it's not a new statement.

gap/attr.gi
```diff
@@ -290,6 +290,177 @@ function(D)
...
+        # Index S \ I
+        i := Sum(s_without_I, x -> 2 ^ (x - 1)) + 1;
```
mtorpey, line 334: Again, see if you could avoid this by using some kind of loop variable?  Might be hard though.

gap/attr.gi
```diff
@@ -290,6 +290,177 @@ function(D)
...
+    return 1;  # chromatic number = 1 iff <D> has >= 1 verts & no edges
+  elif IsBipartiteDigraph(D) then
+    return 2;  # chromatic number = 2 iff <D> has >= 2 verts & is bipartite
+               # <D> has at least 2 vertices at this stage
+  fi;
+  # Now check if the graph is three colourable. This is done by searching
+  # for a maximal independent set where the subgraph induced by removing the
+  # vertices in the set is bipartite.
+  # Need to make a copy in case we are given a mutable digraph
+  D := DigraphImmutableCopy(D);
```
mtorpey, line 364: Change to `DigraphImmutableCopyIfMutable` to avoid extra work.

gap/attr.gi
```diff
@@ -290,6 +290,177 @@ function(D)
...
+    return 1;  # chromatic number = 1 iff <D> has >= 1 verts & no edges
+  elif IsBipartiteDigraph(D) then
+    return 2;  # chromatic number = 2 iff <D> has >= 2 verts & is bipartite
+               # <D> has at least 2 vertices at this stage
+  fi;
+  vertices := DigraphVertices(D);
+  # Store current best colouring for each subset
+  x := [1 .. 2 ^ n];
+  # Empty set is 0 colourable
+  x[1] := 0;
```
mtorpey, line 397: Maybe double-check this is the empty set?

gap/attr.gi
```diff
@@ -290,6 +290,177 @@ function(D)
...
+  od;
+  # Process 4 colourable subgraphs
+  for I in DigraphMaximalIndependentSets(D) do
+    vertex_copy := ShallowCopy(vertices);
+    SubtractSet(vertex_copy, I);
+    # Iterate over all subsets of V(D) \ I
+    for s in IteratorOfCombinations(vertex_copy) do
+      i := index_subsets(s);
+      if x[i] = 3 then
+        s_copy := ShallowCopy(s);
```
mtorpey, line 421: Can we get away without copying `s` here?  Or do we use `s` later?

gap/attr.gi
```diff
@@ -290,6 +290,177 @@ function(D)
...
+        if x[j] > 4 then
+          x[j] := 4;
+        fi;
+      fi;
+    od;
+  od;
+  # Iterate over all vertex subsets.
+  subset_iter := IteratorOfCombinations(vertices);
+  # Skip the first one, which should be the empty set.
+  NextIterator(subset_iter);
```
mtorpey, line 433: Again, an `IsEmpty` check would be good here.

mtorpey: ...unless there's anything in the doc that promises that this is the empty set.

gap/attr.gi
```diff
@@ -290,6 +290,177 @@ function(D)
...
+      # Iterate over the maximal independent sets of D[S]
+      for I in DigraphMaximalIndependentSets(induced_subgraph) do
+        # Bound the size of sets we need to consider.
+        # TODO Filter maximal independent set sizes during calculation.
+        if Length(I) <= Length(s) / x[i] then
+          s_copy := ShallowCopy(s);
+          # Union with I, but need to relabel the induced subgraph
+          # labels back to their original labels
+          UniteSet(s_copy,
+            SetX(I, x -> DigraphVertexLabel(induced_subgraph, x)));
```
mtorpey, line 451: Better indent `SetX` to be in line with `s_copy`.  C'est la mode!

mtorpey (dismissed):
This is looking good!  @james-d-mitchell and @wilfwilson might have other comments later, but I've highlighted a few things to look at here (mostly low-hanging improvements).

Also, as James suggested, be sure to run both algorithms on (say) all graphs up to size 10 and check it matches the current implementation.  Let us know here if it passes!

gap/attr.gi
EwanGilligan: Changed now

gap/attr.gi
EwanGilligan: From reading the documentation, there is no guaranteed order that it will give, so a loop counter can't be used.

gap/attr.gi
EwanGilligan: I don't think using `DigraphVertexLabels(induced_subgraph)` will work, as it only returns a list of the labels and not a mapping from the vertices of `induced_subgraph' to the old labels. This mapping is needed so that I can know which vertices in the original set correspond to the verticies in the maximal indepent sets of the induced subdigraph.

gap/attr.gi
EwanGilligan: Done!

gap/attr.gi
EwanGilligan: Done!

gap/attr.gi
EwanGilligan: I've changed it to undoing the changes at the end, so the copy is no longer required.

gap/attr.gi
EwanGilligan: Done!

gap/attr.gi
EwanGilligan: Fixed!

gap/attr.gi
EwanGilligan: See previous comment

gap/attr.gi
EwanGilligan: Found a corner case with that, which is the cycle digraph inducing with a single vertex. In the induced subgraph there is only one vertex, which is a maximal independent set. The issue is in the overall graph the single vertex isn't a maximal independent set, as it is contained in another maximal independent set (but this set is excluded due to the vertices that we exclude). See below for the example:
```
gap> D := CycleDigraph(5);;
gap> D1 := InducedSubdigraph(D, [5]);
<immutable empty digraph with 1 vertex>
gap> DigraphMaximalIndependentSets(D, [1, 2, 3, 4]);
[  ]
gap> DigraphMaximalIndependentSets(D1);
[ [ 1 ] ]
```
I'll do some more investigation to see if this is a specific corner case that could be explicitly checked (as removing the induced subdigraph call does speed up things enough that an extra check will still make this faster), but for now I've reverted it to the previous working version.

EwanGilligan:
> Also, as James suggested, be sure to run both algorithms on (say) all graphs up to size 10 and check it matches the current implementation. Let us know here if it passes!

I've ran tests for both algorithms on all simple connected graphs up to size 8, and it passes that. The graphs were generated using nauty.

gap/attr.gi
mtorpey: Good thinking!  We've now spoken about this, and I expect there would be other non-trivial cases -- the problem is the difference between "maximal in the whole graph" and "maximal in the restriction".

Have a think about it – but I expect it'll be hard to do anything better than using `InducedSubdigraph` as you have.

gap/attr.gi
mtorpey: Fair enough!  This'll do, then.

gap/attr.gi
mtorpey: I think the list of labels can be used in the same way, because the vertices are just integers.  So,
```gap
DigraphVertexLabel(induced_subgraph, x)
```
is the same as
```gap
labels := DigraphVertexLabels(induced_subgraph);
labels[x]
```
the only difference being the second one cuts down on method calls when used repeatedly.

EwanGilligan: Ah I see, forgot that the vertices would index into that. I'll make that change now

gap/attr.gi
```diff
@@ -290,6 +290,226 @@ function(D)
...
+  # Iterator for blist subsets.
+  subset_iter := ListWithIdenticalEntries(n, [false, true]);
+  subset_iter := IteratorOfCartesianProduct2(subset_iter);
+  # Skip the first one, which should be the empty set.
+  S := NextIterator(subset_iter);
+  # Iterate over all vertex subsets.
+  for S in subset_iter do
+    # Cartesian iterator is ascending lexicographically, but we want reverse
+    # lexicographic ordering.
+    FlipBlist(S);
```
mtorpey, line 351: `FlipBlist` was added in GAP 4.11, and seems to be quite useful here.  @james-d-mitchell and @wilfwilson, would we be happy with advancing the required GAP from `>=4.9.0` to `>=4.11.0` now that it's been out for a year?  Otherwise, Ewan should provide some kind of fallback using `if IsBound(FlipBlist)`.

tst/standard/attr.tst
```diff
@@ -1509,6 +1509,110 @@ gap> ChromaticNumber(a);
...
+4
+gap> gr := CycleDigraph(7);
+<immutable cycle digraph with 7 vertices>
+gap> ChromaticNumber(gr, DigraphColouringAlgorithmByskov);
+3
+gap> ChromaticNumber(gr, DigraphColouringAlgorithmByskov);
+3
+gap> ChromaticNumber(gr, DigraphColouringAlgorithmByskov);
+3
+
```
mtorpey, line 1615: Good tests here!  Did you run your functions on all graphs up to size `n` again, like you did before?  It'd be good to have that extra bit of confidence now you've changed a few things.

mtorpey (commented):
Some thoughts here.  Looking good overall!

tst/standard/attr.tst
EwanGilligan: I kept running the same tests on all graphs up to size 8 as I performed the optimisations, and it still passes them.

gap/attr.gi
```diff
@@ -290,6 +290,226 @@ function(D)
...
+  elif IsNullDigraph(D) then
+    return 1;  # chromatic number = 1 iff <D> has >= 1 verts & no edges
+  elif IsBipartiteDigraph(D) then
+    return 2;  # chromatic number = 2 iff <D> has >= 2 verts & is bipartite
+               # <D> has at least 2 vertices at this stage
+  fi;
+  vertices := List(DigraphVertices(D));
+  # Store all the Maximal Independent Sets, which can later be used for
+  # calculating the maximal independent sets of induced subgraphs.
+  MIS := DigraphMaximalIndependentSets(D);
```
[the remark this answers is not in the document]
EwanGilligan: I've changed that to `mis` now; I'll be sure to keep the style comments in mind for future pull requests

gap/attr.gi
```diff
@@ -290,6 +290,226 @@ function(D)
...
+    if SizeBlist(i) <= size_bound and
+        ForAll(induced_mis, x -> not IsSubsetBlist(x, i)) then
+      Add(induced_mis, i);
+    fi;
+  od;
+  return induced_mis;
+end
+);
+
+InstallMethod(ChromaticNumber, "for a digraph and colouring algorithm",
```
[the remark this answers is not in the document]
EwanGilligan: Changed!

alex-konovalov:
This needs rebasing.

EwanGilligan:
> This needs rebasing.

Thanks I'd forgotten to do that with the review changes, so that's been done now. Thanks for pointing that out!

mtorpey:
@wilfwilson

gap/attr.gi
wilfwilson: Sorry I missed this last week. In short, I don't care that much, but my slight inclination is to maintain backwards compatibility with GAP 4.10 if it's not too much effort/too painful to do so, for now. However I'm happy to defer to @james-d-mitchell on this question.

Just looking briefly at the code, however, my first question is to ask whether maybe you don't need to be flipping the blist (at least not so much) - you're flipping the blist three times; is there some easy way to do your operations on the original blist directly, without having to to flip it so much?

doc/attr.xml
```diff
@@ -1439,6 +1439,42 @@ gap> ChromaticNumber(D);
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="ChromaticNumberFunc">
+<ManSection>
+    <Func Name="ChromaticNumber" Arg="digraph, colouring algorithm"
```
alex-konovalov, line 1444: The argument name with a space is misleading - perhaps use an underscore?

doc/attr.xml
```diff
@@ -1439,6 +1439,42 @@ gap> ChromaticNumber(D);
...
+        Label="for a digraph and colouring algorithm"/>
+    <Returns> A non-negative integer.</Returns>
+    <Description>
+      Alterative implementation of the Chromatic Number attribute
+      (See <Ref Attr="ChromaticNumber"/>) that allows the specific algorithm 
+      used to be selected. 
+
+      If <A>digraph</A> is a digraph without loops (see <Ref 
+      Prop="DigraphHasLoops"/>), then <C>ChromaticNumber</C> uses 
+      <A>colouring algorithm</A> to find the least non-negative integer <C>n</C>
```
alex-konovalov, line 1454: Would it be worth to specify that the algorithm is not given as a string, but as a function?

gap/attr.gi
EwanGilligan: Having a look I've managed to eliminate a few of the flip calls, but that can't be done everywhere. For now I've just added a fallback that checks if FlipBlist can be used.

wilfwilson: Nice. Thanks for doing that, I hope you don't mind too much. How much slower roughly does your fallback method make your `ChromaticNumber` methods, have you done any experiments?

alex-konovalov:
@EwanGilligan this PR needs rebasing.

doc/attr.xml
EwanGilligan: Thanks, I've changed that

doc/attr.xml
EwanGilligan: Good point, will make it more clear

gap/attr.gi
EwanGilligan: I've tested it out on a few graphs, and it is worse but only marginally and it is unlikely to affect the overall performance of any code that would use this. Here is an example, just with quickly timing them (I made sure to run it a few times to check it was consistent)
```
D := DigraphRemoveLoops(RandomDigraph(15));
<immutable digraph with 15 vertices, 24 edges>
ChromaticNumber(D, DigraphColouringAlgorithmLawler);time;
3
940
MakeReadWriteGlobal("FlipBlist");
FlipBlist := function(B) Apply(B, x -> not x); end;
ChromaticNumber(D, DigraphColouringAlgorithmLawler);time;
3
962
```
For some larger graphs one went from taking 30 seconds to run to 31 seconds, so the fallback isn't too bad.

alex-konovalov (commented):
@EwanGilligan I can show how to rebase a feature branch instead of merging the master branch into it, if you'd like.

EwanGilligan:
> @EwanGilligan I can show how to rebase a feature branch instead of merging the master branch into it, if you'd like.

Thanks @alex-konovalov, I've rebased this instead now. I'd clicked the update branch github provides without really thinking, but I forgot a rebase would be better

wilfwilson:
It doesn't matter so much whether you rebase or use the GitHub 'update branch' feature, because we ultimately perform a 'rebase and merge' when we accept pull requests which should (in theory) get rid of any of those 'update branch' commits.

doc/attr.xml
```diff
@@ -1439,6 +1439,42 @@ gap> ChromaticNumber(D);
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="ChromaticNumberFunc">
+<ManSection>
+    <Func Name="ChromaticNumber" Arg="digraph, colouring_algorithm" 
+        Label="for a digraph and colouring algorithm"/>
+    <Returns> A non-negative integer.</Returns>
```
wilfwilson, line 1446:
```suggestion
    <Returns>A non-negative integer.</Returns>
```

doc/attr.xml
```diff
@@ -1439,6 +1439,42 @@ gap> ChromaticNumber(D);
...
+    <Description>
+      Alterative implementation of the Chromatic Number attribute
+      (See <Ref Attr="ChromaticNumber"/>) that allows the specific algorithm 
+      used to be selected.
```
wilfwilson, line 1450:
```suggestion
      Alternative implementation of chromatic number
      (See <Ref Attr="ChromaticNumber"/>) that allows the specific algorithm 
      to be selected. 
```

doc/attr.xml
```diff
@@ -1439,6 +1439,42 @@ gap> ChromaticNumber(D);
...
+        Label="for a digraph and colouring algorithm"/>
+    <Returns> A non-negative integer.</Returns>
+    <Description>
+      Alterative implementation of the Chromatic Number attribute
+      (See <Ref Attr="ChromaticNumber"/>) that allows the specific algorithm 
+      used to be selected. 
+
+      If <A>digraph</A> is a digraph without loops (see <Ref
+      Prop="DigraphHasLoops"/>), then <C>ChromaticNumber</C> uses
+      <A>colouring algorithm</A> (which is given as a globally bound object)
```
wilfwilson, line 1454: I think this is clearer
```suggestion
      <A>colouring_algorithm</A> (permitted values for this are given below)
```

doc/attr.xml
```diff
@@ -1439,6 +1439,42 @@ gap> ChromaticNumber(D);
...
+      to find the least non-negative integer <C>n</C> such that there is a
+      proper colouring of <A>digraph</A> with <C>n</C> colours. See
+      <Ref Attr="ChromaticNumber"/> for the definition of a proper colouring.
+
+      The current algorithm implementations are:
```
wilfwilson, line 1459:
```suggestion
      The permitted implementations, and values for <A>colouring_algorithm</A>, are:
```

doc/attr.xml
```diff
@@ -1439,6 +1439,42 @@ gap> ChromaticNumber(D);
...
+      <List>
+          <Item><A>DigraphColouringAlgorithmLawler</A> - Lawler's Algorithm 
+          <Cite Key="Law1976"/></Item>
+          <Item><A>DigraphColouringAlgorithmByskov</A> - Byskov's Algorithm
```
wilfwilson, line 1463: `<A>` tags are reserved for the specific names of arguments. So `<C>` tags are more appropriate here.
```suggestion
          <Item><C>DigraphColouringAlgorithmLawler</C> - Lawler's Algorithm 
          <Cite Key="Law1976"/></Item>
          <Item><C>DigraphColouringAlgorithmByskov</C> - Byskov's Algorithm 
```

gap/attr.gi
```diff
@@ -289,6 +289,223 @@ function(D)
...
+        ForAll(induced_mis, x -> not IsSubsetBlist(x, i)) then
+      Add(induced_mis, i);
+    fi;
+  od;
+  return induced_mis;
+end
+);
+
+InstallMethod(ChromaticNumber, "for a digraph and Lawler's colouring algorithm",
+[IsDigraph, IsDigraphColouringAlgorithm and IsDigraphColouringAlgorithmLawler],
```
wilfwilson, line 315: Surely the `IsDigraphColouringAlgorithm` is redundant?
```suggestion
[IsDigraph, IsDigraphColouringAlgorithmLawler],
```
(and if not, maybe something isn't set up correctly?)

And similarly for the other one?

gap/algobj.gi
```diff
@@ -0,0 +1,18 @@
...
+                DigraphColouringAlgorithmFamilyObj,
+                IsDigraphColouringAlgorithmLawler),
+              rec()));
+BindGlobal("DigraphColouringAlgorithmByskov",
+            Objectify(
+              NewType(
+                DigraphColouringAlgorithmFamilyObj,
+                IsDigraphColouringAlgorithmByskov),
+              rec()));
+Unbind(DigraphColouringAlgorithmFamilyObj);
```
wilfwilson, line 18: Why have you chosen to go for this setup? This way, `DigraphColouringAlgorithmLawler` (for instance) is just an object that basically only knows its name, and that it's a name representing an algorithm. Were there some discussions on this topic this that I missed out on?

I've not thought this through... and I'm not sure this is even possible... but... another option could be that `IsDigraphAlgorithm` objects (amongst other things) are actual functions (or at least they contain a function) that implement the algorithm in question. So `DigraphColouringAlgorithmLawler` could be an object satisfying `IsDigraphAlgorithm`, but at the same time it contains its own implementation, so you could do `DigraphColouringAlgorithmLawler(D)` for a digraph `D` if you really wanted and get the result of applying the algorithm to the graph.

And then you'd just have the one method:
```
InstallMethod(ChromaticNumber, "for a digraph and colouring algorithm",
[IsDigraph, IsDigraphColouringAlgorithm],
{D, f} -> f(D));
```
What do you think of that? I'm just curious if you had thought of that, or if you or anyone else (@mtorpey?) have opinions on the best setup.

wilfwilson (commented):
A few comments - I've not actually looked at the code yet 🙂  I got distracted by the documentation and a more conceptual question.

gap/algobj.gi
mtorpey: We discussed what the best setup would be with @james-d-mitchell a few months ago, and we settled on a setup which is what Ewan's implemented here: a filter for each algorithm, which basically just knows its own name, to allow a user to specify the right one.  However, I don't think any of us considered actually having the algorithm **be** the function like @wilfwilson suggests.

This suggestion seems to make a lot of sense!  It'd be less surprising, more "object-oriented" (as far as anything in GAP is!) and would still let us set up filters in interesting ways later on.

@EwanGilligan and @james-d-mitchell, does Wilf's suggestion sound good to you?

wilfwilson: Thanks for that context @mtorpey! (And just to be clear: my suggestion isn't a request - I don't really mind what the setup is.)

EwanGilligan: Like @mtorpey has said, I don't think we thought about the suggestion of the object passed in be the function was considered. To me this would make more sense, but I'll wait to hear what @james-d-mitchell has to say before I make the change.

EwanGilligan: @wilfwilson This has been discussed a bit further now, with the outcome mostly being that passing a function would be nicer, but the current method works better due to the way GAP filters are set up.

So the main issue with the approach of passing the function directly to `ChromaticNumber` is that you would not be able to use filters to restrict the function that could be passed. As functions are internal representations, it is not possible to set filters on them after creation. It would be possible to wrap this function in an object that contains the function and has the right filters, but then you would lose the ability to call the function directly. This would be a nice solution if GAP allowed you to override the call method, but unfortunately this is not possible.

So then the option to store an object that only knows it's own name or storing an object that stores the function would be identically used by the user. The current mechanism is then a bit simpler, as the function wrapping way would require an additional method to be installed that just called the wrapped function. I've found out that the name of the current mechanism is something known as tagged dispatch, so is a common pattern for these types of problems.

Hope this ramble has helped clear up why we've chose to perform the method selection this way.

wilfwilson: That's fine @EwanGilligan 🙂. Nevertheless, this might be interesting...
> This would be a nice solution if GAP allowed you to override the call method, but unfortunately this is not possible.

I learned this morning that (I think?!)  you _can_ override the call method - i.e. look at this:
```
gap> InstallMethod(CallFuncList, [IsDigraph, IsList], {f, args} -> Sum(args));
gap> D := RandomDigraph(10);;
gap> D(4, 6, 2);
12
```
So you can 'call' a digraph on some arguments! (Obviously nonsense.) But if your wrapper objects were called `IsEwanWrapper`, and they were component objects with the component `func` being the underlying function, then you could do:
```
gap> InstallMethod(CallFuncList, [IsEwanWrapper, IsList], {f, args} -> CallFuncList(f!.func, args));
```
So if `f` was one of these wrapper objects, you could do `f(x)` and this call would be passed to `f!.func`, i.e. the result of `f(x)` would be `f!.func(x)`.

wilfwilson: But again, I'm not requesting that you do this 🙂 Thanks for thinking about this and replying thoughtfully. I'm happy with whatever.

EwanGilligan: @wilfwilson Very interesting about overriding `CallFuncList`, I'd asked around but nobody seemed to know that it was possible.

I think overall the current "tagged dispatch" approach is probably simpler, as it doesn't require the additional method to override CallFuncList for the wrapper object. Thank you for taking the time to let me know about the other option

wilfwilson (changes requested):
Hi @EwanGilligan. Thanks again for the technical discussion we had about the way that you implemented that algorithm objects.

I've finally had time to think about this a bit more properly and (unfortunately, sorry!) I don't think your current solution is the appropriate one. However, I don't think it'll take much time or effort to change it to my preferred option, and if you don't have time to do it then I can do it soon-ish, just let me know.

(I'm happy with the actual implementation by the way.)

In short, I think the problem that you were trying to solve is exactly what GAP's `ValueOption`s are for. See https://www.gap-system.org/Manuals/doc/ref/chap8.html#X7F9373AD7DB88D1F for more information about this feature, and https://www.gap-system.org/Manuals/doc/ref/chap4.html#X867D54987EF86D1D for information about how you call a function with options.

In particular, I think your new functionality should be available, and documented as such, through a function call like `ChromaticNumber(D : lawler)`, or `ChromaticNumber(D : Byskov)`, or `ChromaticNumber(D : algorithm := "Lawler")`  - whichever you prefer.

The main chromatic number method can then take this form:

https://github.com/digraphs/Digraphs/blob/ccae5d35fbfa50349b5524993c28fe97127e0fbb/gap/attr.gi#L177-L199

(in particular, these pre-checks don't need to be duplicated in each of the new methods) and then immediately following this, you could parse the value options and dispatch to one of your new algorithms as appropriate, in some way like this:
```
if ValueOption("lawler") <> fail then
  # dispatch to `DIGRAPHS_ColouringAlgorithmLawler`
elif ValueOption("byskov") <> fail then
  # dispatch to `DIGRAPHS_ColouringAlgorithmByskov`
fi;
# else continue as before
```
Your new `DigraphColouring` methods would be renamed to functions/operations/whatever called `DIGRAPHS_ColouringAlgorithmLawler` and `DIGRAPHS_ColouringAlgorithmByskov`, as appropriate.

Your new documentation can then be merged with the original documentation for `ChromaticNumber`, and the new `gap/algobj.g*` files can just be deleted.

I think this is a lighter-weight, easier to maintain solution, but (unless I'm missing something - let me know if so!) just as useful.

gap/attr.gi
```diff
@@ -300,6 +300,223 @@ function(D)
...
+  if DigraphHasLoops(D) then
+    ErrorNoReturn("the argument <D> must be a digraph with no loops,");
+  elif n = 0 then
+    return 0;  # chromatic number = 0 iff <D> has 0 verts
+  elif IsNullDigraph(D) then
+    return 1;  # chromatic number = 1 iff <D> has >= 1 verts & no edges
+  elif IsBipartiteDigraph(D) then
+    return 2;  # chromatic number = 2 iff <D> has >= 2 verts & is bipartite
+               # <D> has at least 2 vertices at this stage
+  fi;
```
wilfwilson, line 430: These lines could also go.
```suggestion
```

doc/attr.xml
```diff
@@ -1439,6 +1439,43 @@ gap> ChromaticNumber(D);
...
+      <Ref Attr="ChromaticNumber"/> for the definition of a proper colouring.
+
+      The permitted implementations, and values for <A>colouring_algorithm</A>,
+      are:
+      <List>
+          <Item><C>DigraphColouringAlgorithmLawler</C> - Lawler's Algorithm 
+          <Cite Key="Law1976"/></Item>
+          <Item><C>DigraphColouringAlgorithmByskov</C> - Byskov's Algorithm 
+          <Cite Key="Bys2002"/></Item>
+      </List>
```
wilfwilson, line 1466: This documentation could then be added to the main bit of documentation, perhaps like:
> it is possible to select the algorithm used to compute the chromatic number of a digraph via ...
> The permitted algorithms, and values for BLAH, are...

gap/algobj.gi
wilfwilson, line 18: This file would go!

init.g
```diff
@@ -37,12 +37,20 @@ fi;
...
+    Apply(BList, x -> not x);
+  end;
+fi;
 
 ReadPackage("digraphs", "gap/digraph.gd");
 ReadPackage("digraphs", "gap/digraphs.g");
 ReadPackage("digraphs", "gap/constructors.gd");
 ReadPackage("digraphs", "gap/grape.gd");
 ReadPackage("digraphs", "gap/labels.gd");
+ReadPackage("digraphs", "gap/algobj.gd");
```
wilfwilson, line 53:
```suggestion
```

read.g
```diff
@@ -29,6 +29,7 @@ ReadPackage("digraphs", "gap/digraph.gi");
 ReadPackage("digraphs", "gap/constructors.gi");
 ReadPackage("digraphs", "gap/grape.gi");
 ReadPackage("digraphs", "gap/labels.gi");
+ReadPackage("digraphs", "gap/algobj.gi");
```
wilfwilson, line 32:
```suggestion
```

wilfwilson (commented):
Another benefit of doing it my suggested way is that then attributes are stored correctly. Mathematically, all three algorithms (should be) return the same answer for the chromatic number, and therefore it is sensible to store the value of the attribute `ChromaticNumber` for any immutable digraph, no matter which algorithm is used. Currently, this is not implemented (although you could manually add it), but you would get it automatically with the use of `ValueOption`s:
```
gap> D := Digraph([ [ 2 ], [ 1, 3, 5 ], [  ], [ 2 ], [ 1 ] ]);;
gap> ChromaticNumber(D); HasChromaticNumber(D);
3
true
gap> D := Digraph([ [ 2 ], [ 1, 3, 5 ], [  ], [ 2 ], [ 1 ] ]);;
gap> ChromaticNumber(D, DigraphColouringAlgorithmLawler); HasChromaticNumber(D);
3
false
gap> D := Digraph([ [ 2 ], [ 1, 3, 5 ], [  ], [ 2 ], [ 1 ] ]);;
gap> ChromaticNumber(D : something); HasChromaticNumber(D);
3
true
```

EwanGilligan:
@wilfwilson The value option way does seem to pretty much be the "correct" way of doing what we're already doing, so thanks for letting me know about those. Shouldn't be too bad to make the changes, so I should be able to get them sorted this afternoon.

gap/attr.gi
```diff
@@ -174,6 +174,201 @@ function(D)
...
+  elif DigraphColouring(D, 3) <> fail then
+               # Check if there is a 3 colouring
+    return 3;
+  fi;
+  return infinity;
+end
+);
+
+BindGlobal("DIGRAPHS_ChromaticNumberByskov",
+function(D, Byskov)
```
wilfwilson, line 278: Thanks a lot for making these changes @EwanGilligan. I've spotted another thing that needs to be changed.
```suggestion
function(D)
```
I'm not sure why the CI is crashing so much... Perhaps making that change will magically fix it?!

wilfwilson: Nope.

EwanGilligan: Ah that seems to be the issue, was searching for why everything was crashing. Thanks @wilfwilson !

wilfwilson: Unfortunately the situation persists on my computer, and probably will do on Azure/GitHub Actions (when those jobs finally run).

However, I can make the problem disappear by commenting out the lines:
```
  if ValueOption("lawler") <> fail then
    return DIGRAPHS_ChromaticNumberLawler(D);
  elif ValueOption("byskov") <> fail then
    return DIGRAPHS_ChromaticNumberByskov(D);
  fi;
```
(i.e. if your code is bypassed).

Now, this doesn't prove that your code is buggy, but it at least strongly suggests that your code is at least running another big of code which is buggy. And given the kind of crash that's happening, this suggests that it's whichever code your calling that is written in C.

For me, that points to `DigraphMaximalIndependentSets`...

wilfwilson: I'm poking around a bit more.

EwanGilligan: I don't believe anything should have changed, such that it would now run out of memory. Checking back it passed all CI previously, before the new method selection was used. I'll investigate to see if I can reproduce it, but it all works fine on my end. Interestingly (of the ones that have currently run) only the github action for 4.11 seems to fail, with the 4.10 one passing

tst/standard/attr.tst
```diff
@@ -1548,6 +1548,116 @@ gap> D := Digraph(OutNeighbours(CycleDigraph(13)));;
 gap> ChromaticNumber(D);
 3
 
+#  Test ChromaticNumber Lawler
+gap> ChromaticNumber(Digraph([[1]]) : lawler);
+Error, the argument <D> must be a digraph with no loops,
```
wilfwilson, line 1553: Okay @EwanGilligan this is bonkers but I think (even though this should obviously be completely harmless!) that if you just delete this test, as well as the corresponding one for `Byskov`), then this should pass. (And those tests aren't even necessary for this PR, since that code is already tested).
```suggestion
```
I need to investigate why, but I think that's obviously independent of this PR...

EwanGilligan: That's incredibly strange, but I'll give that a go

wilfwilson (approved):

wilfwilson:
Thanks again Ewan. Just to be safe, I'm adding the "do not merge" tag, while I pinpoint the cause of those crashes that we had, in the very slight chance that it is somehow related to this PR.

gap/attr.gi
```diff
@@ -174,6 +174,201 @@ function(D)
...
+  mis, subset_mis;
+  n := DigraphNrVertices(D);
+  vertices := List(DigraphVertices(D));
+  # Store all the Maximal Independent Sets, which can later be used for
+  # calculating the maximal independent sets of induced subgraphs.
+  mis := DigraphMaximalIndependentSets(D);
+  # Convert each MIS to a Blist
+  mis := List(mis, x -> BlistList(vertices, x));
+  # Store current best colouring for each subset
+  subset_colours := ListWithIdenticalEntries(2 ^ n, infinity);
```
wilfwilson, line 211: The ultimate reason for the crashes is that on the second run-through of the test file, your new algorithms are (incorrectly) being run when they shouldn't be run (see https://github.com/gap-system/gap/issues/4510 - the option value was being (in my opinion, wrongly) retained between different calls to `ChromaticNumber` after an error).

The reason that this is a problem is that a list with `2 ^ n` entries quickly becomes HUGE and crashes GAP when `n` is big enough (and it doesn't have to be that big).

Would it be easy to avoid having such a large list? If not, perhaps these functions could give an error - or at least a warning - if they are called with a digraph with 'too many' vertices (whatever that should be).

PR #386, opened 2021-01-27
Add `FolkmanGraph`

finnbuck:
Added functions for constructing a [Folkman Graph](https://en.wikipedia.org/wiki/Folkman_graph). @james-d-mitchell


finnbuck:
We have decided to go a different direction with named graphs. However, this version of the Folkman Graph implementation could still be useful as a template.

james-d-mitchell:
Thanks @finnbuck we did decide to go in a different direction for sporadic (i.e. non-infinite families of) digraphs, but I think we should leave this open for the purpose that you mentioned @finnbuck, as an exemplar of what to do to add an infinite family. Thanks for this!

alex-konovalov (approved):
Looks good to me - thanks!

wilfwilson:
If this is not intended to be merged, I think it's probably an appropriate point to close this PR now, right? (Seeing as there are already several examples of this kind that have now been merged).

james-d-mitchell:
Closing this as it was an exemplar only!

PR #388, opened 2021-01-28
Merge stable-1.4 into master

james-d-mitchell:
Merge stable-1.4 into master and update the year in the README.


wilfwilson:
Merging stable branches into master is my least favourite part of the current branch protection rules.

One of these commits (304493d5cf1047a28ebd348a310707145c604f3d) already exists in `origin/stable-1.4`, while the other two commits only exist in your `master` branch, and this PR asks to merge them all into `origin/master`. If we do that via a rebase-and-merge, will `origin/master` contain commit 304493d5cf1047a28ebd348a310707145c604f3d, or will this commit be duplicated and we'll have two copies of the same commit in the repository?

Moreover, the two commits that are only in your `master` branch should probably be in `origin/stable-1.4` also. So once we've merged this request, we'll also need to merge the new `origin/master` into `origin/stable-1.4`. And again, at that point, will that duplicate commits again, or will we get nice history? This screenshot below is how we surely _want_ everything to look:

Screen Shot 2021-01-28 at 10 23 04

I prefer the previous system, where we do these merges locally, and then push to our forks, which will run the CI on our forks (i.e. I would see the result at https://github.com/digraphs/Digraphs/actions), and only once that's passed, should we push to this `origin` repository.

I think that any _new_ commits (i.e. new features, bug fixes, etc etc) should only enter the repository via pull requests, but for merging stable and master branches into each other when required, I would like to go back to doing it manually.

wilfwilson (approved):

james-d-mitchell:
Ah sorry @wilfwilson I merged before I saw your comment. I agree with completely, I didn't intentionally change the setup, and would be happy to revert it to the old setup, I'm not 100% certain how to do that though. I'm happy for you to do it!

wilfwilson:
No worries. I'll look into how I need to set the branch protection rules 🙂 

And to answer my earlier question - yes, the commit `Update version numbers etc for v1.4.0` is now duplicated 🙁 :

Screen Shot 2021-01-28 at 10 50 49

PR #398, opened 2021-02-03
Prototype for Group 1's effort for the named digraphs stuff

james-d-mitchell:


james-d-mitchell:
One thing to bear in mind @tomcontileslie and @MTWhyte when you're looking at this: the CI is failing because GAP doesn't seem to distinguish between an empty list `[]` and an empty string `""`, and so when we do `Digraph([])` GAP tries to use the new method in this PR.

tomcontileslie:
Ah thanks James, that makes sense. A `TryNextMethod()` at the beginning of the implementation might be a sensible fix for this? Or perhaps we should give the whole thing a lower priority score than the other digraph functions?

james-d-mitchell:
I'm going close this in favour of a revised PR from @MTWhyte @mariatsalakou and @tomcontileslie

PR #405, opened 2021-02-10
Shorten lines in standard test files

wilfwilson:
I have been slightly annoyed on occasion by long lines when looking in test files, so I just shortened them all to get it over and done with. It would be nice if the linter enforced line lengths on test files so they don't creep back in.

~~In the process of doing this, I discovered some commented-out tests in `tst/standard/display.tst` that I guess predate the Semigroups package requiring Digraphs, and Semigroups had its own copy of `Splash`?.~~

~~Everything seems fine with Semigroups loaded on my end. So reinstating these might improve our code coverage!~~

PR #406, opened 2021-02-10
Reinstate Splash-related tests in display.tst

wilfwilson:
I've broken this off from #405. The following lines in `tst/standard/display.tst` are commented out, and I'm trying to work out a way of uncommenting and adjusting them, so that we get better code coverage, and don't have commented-out tests lying around.


wilfwilson:
I'm not going to deal with this for a while, possibly a long while, so I'll close it.

PR #408, opened 2021-02-13
Add HaarGraph

finnbuck:
Add global function for generating a [Haar Graph](https://mathworld.wolfram.com/HaarGraph.html) based on a single parameter. @flsmith

### DIFF
--- a/doc/examples.xml
+++ b/doc/examples.xml
@@ -389,3 +389,50 @@ gap> KnightsGraph(IsMutable, 3, 9);
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="HaarGraph">
+<ManSection>
+  <Oper Name="HaarGraph" Arg="[filt, ]n"/>
+  <Returns>A digraph.</Returns>
+  <Description>
+    If <A>n</A> is a positive integer then this operation returns
+    the <E>Haar graph</E> <M>H(<A>n</A>)</M>. <P/>
+
+    From
+    <URL>https://mathworld.wolfram.com/HaarGraph.html</URL>:
+    <P/>
+
+    <Q>A Haar graph H(<A>n</A>) is a bipartite regular vertex-transitive graph
+    indexed by a positive integer and obtained by a simple binary encoding of
+    cyclically adjacent vertices. Haar graphs may be connected or
+    disconnected.</Q><P/>
+
+    The number of vertices in the Haar graph <M>H(<A>n</A>)</M> is
+    equal to twice <M>m</M>, where <M>m</M> is the number of digits
+    required to represent <A>n</A> in binary.
+    These vertices are arranged into bicomponents
+    <C>[1..m]</C> and <C>[m+1..2*m]</C>.
+    Vertices <M>i</M> and <M>j</M> in different bicomponents are
+    adjacent by a symmetric pair of edges if and only if
+    the binary representation of <A>n</A> has a 1 in position
+    (<M>j-i</M> modulo <M>m</M>) + 1 from the left. <P/>
+
+    If the optional first argument <A>filt</A> is present, then this should
+    specify the category or representation the digraph being created will
+    belong to. For example, if <A>filt</A> is <Ref Filt="IsMutableDigraph"/>,
+    then the digraph being created will be mutable, if <A>filt</A> is <Ref
+    Filt="IsImmutableDigraph"/>, then the digraph will be immutable.
+    If the optional first argument <A>filt</A> is not present, then <Ref
+      Filt="IsImmutableDigraph"/> is used by default.<P/>
+
+<Example><![CDATA[
+gap> HaarGraph(3);
+<immutable bipartite vertex-transitive symmetric digraph with bicompon\
+ent sizes 2 and 2>
+gap> D := HaarGraph(16);
+<immutable bipartite vertex-transitive symmetric digraph with bicompon\
+ent sizes 5 and 5>
+]]></Example>
+  </Description>
+</ManSection>
+<#/GAPDoc>
--- a/gap/examples.gi
+++ b/gap/examples.gi
@@ -446,3 +446,44 @@ KnightsGraphCons);
 
 InstallMethod(KnightsGraph, "for two positive integers", [IsPosInt, IsPosInt],
 {m, n} -> KnightsGraphCons(IsImmutableDigraph, m, n));
+
+InstallMethod(HaarGraphCons,
+"for IsMutableDigraph and a positive integer",
+[IsMutableDigraph, IsPosInt],
+function(filt, n)
+  local m, binaryList, D, i, j;
+  m := Log(n, 2) + 1;
+  binaryList := DIGRAPHS_BlistNumber(n + 1, m);
+  D := EmptyDigraph(IsMutableDigraph, 2 * m);
+
+  for i in [1 .. m] do
+    for j in [1 .. m] do
+      if binaryList[((j - i) mod m) + 1] then
+          DigraphAddEdge(D, [i, m + j]);
+      fi;
+    od;
+  od;
+
+  return DigraphSymmetricClosure(D);
+end);
+
+InstallMethod(HaarGraphCons,
+"for IsImmutableDigraph and a positive integer",
+[IsImmutableDigraph, IsPosInt],
+function(filt, n)
+  local D;
+  D := MakeImmutable(HaarGraphCons(IsMutableDigraph, n));
+  SetIsMultiDigraph(D, false);
+  SetIsSymmetricDigraph(D, true);
+  SetIsRegularDigraph(D, true);
+  SetIsVertexTransitive(D, true);
+  SetIsBipartiteDigraph(D, true);
+  return D;
+end);
+
+InstallMethod(HaarGraph, "for a function and a positive integer",
+[IsFunction, IsPosInt],
+HaarGraphCons);
+
+InstallMethod(HaarGraph, "for a positive integer", [IsPosInt],
+{n} -> HaarGraphCons(IsImmutableDigraph, n));
--- a/tst/standard/examples.tst
+++ b/tst/standard/examples.tst
@@ -252,6 +252,28 @@ false
 gap> KnightsGraph(IsMutable, 3, 9);
 <mutable digraph with 27 vertices, 88 edges>
 
+# DIGRAPHS_HaarGraph
+gap> HaarGraph(1);
+<immutable bipartite vertex-transitive symmetric digraph with bicomponent size\
+s 1 and 1>
+gap> OutNeighbours(last);
+[ [ 2 ], [ 1 ] ]
+gap> HaarGraph(2);
+<immutable bipartite vertex-transitive symmetric digraph with bicomponent size\
+s 2 and 2>
+gap> OutNeighbours(last);
+[ [ 3 ], [ 4 ], [ 1 ], [ 2 ] ]
+gap> HaarGraph(3);
+<immutable bipartite vertex-transitive symmetric digraph with bicomponent size\
+s 2 and 2>
+gap> OutNeighbours(last);
+[ [ 3, 4 ], [ 3, 4 ], [ 1, 2 ], [ 1, 2 ] ]
+gap> D := HaarGraph(16);
+<immutable bipartite vertex-transitive symmetric digraph with bicomponent size\
+s 5 and 5>
+gap> IsBipartiteDigraph(D);
+true
+
 #
 gap> DIGRAPHS_StopTest();
 gap> STOP_TEST("Digraphs package: standard/examples.tst", 0);
--- a/doc/z-chap2.xml
+++ b/doc/z-chap2.xml
@@ -87,6 +87,7 @@
     <#Include Label="JohnsonDigraph">
     <#Include Label="PetersenGraph">
     <#Include Label="GeneralisedPetersenGraph">
+    <#Include Label="HaarGraph">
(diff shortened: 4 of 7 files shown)


gap/examples.gi
```diff
@@ -370,3 +370,46 @@ GeneralisedPetersenGraphCons);
 
 InstallMethod(GeneralisedPetersenGraph, "for integer, integer", [IsInt, IsInt],
 {n, k} -> GeneralisedPetersenGraphCons(IsImmutableDigraph, n, k));
+
+InstallGlobalFunction(DIGRAPHS_HaarGraph,
+function(n)
+  local m, binaryList, D, i, j;
+  if n < 1 and IsInt(n) then
```
Peter-Ing, line 377: This will attempt to evaluate n < 1 before IsInt(n) which can still return an error if n is incomparable with 1, if the order is changed to check IsInt(n) first this would stop any such error.  I believe you would also need to negate the check for IsInt and likely change the and to an or as you with to return the error for when n is incomparable with 1 or when n < 1.

Peter-Ing (changes requested):

gap/examples.gi
alex-konovalov: You may want to use `IsPosInt` here - see https://www.gap-system.org/Manuals/doc/ref/chap14.html#X82A854757DFA9C76

finnbuck: Right, that makes sense. IsPosInt() seems like a good way to go.

gap/examples.gi
```diff
@@ -370,3 +370,46 @@ GeneralisedPetersenGraphCons);
 
 InstallMethod(GeneralisedPetersenGraph, "for integer, integer", [IsInt, IsInt],
 {n, k} -> GeneralisedPetersenGraphCons(IsImmutableDigraph, n, k));
+
+InstallGlobalFunction(DIGRAPHS_HaarGraph,
+function(n)
+  local m, binaryList, D, i, j;
+  if (not IsPosInt(n)) then
```
alex-konovalov, line 377: A minor nitpick - no need in extra brackets, `if not IsPosInt(n)` works.

samanthaharper:
I think documentation would be really helpful to reviewing this graph, since I'm not sure what you're using to construct the Haar graph.

marinaanagno (commented):
I have consulted Wolfram Alpha to see what this graph is about, and it seems to agree with the definition. However, if we proceed according to what was said in the meeting today, documentation should be added and also this function should be declared as a constructor and operations that return the final mutable and immutable version of the digraph should be added :)

finnbuck:
Ok sounds good! I'll have a look at some examples and try to fix the documentation and function declaration...

james-d-mitchell:
@finnbuck looks ok to me, I think you need to write some doc (as per the other comments) and also to follow the same "pattern" as in the Folkman PR #386. Other than that: good stuff!

finnbuck:
DIGRAPHS_BlistNumber([IsInt], [IsInt]) is taken from a GAP Semigroups function of the same name and is not meant to be user facing here, so I'm assuming it's meant to stay a global function.

wilfwilson:
Closing and reopening to (hopefully) re-trigger the CI.

doc/examples.xml
```diff
@@ -309,3 +309,49 @@ gap> GeneralisedPetersenGraph(IsMutableDigraph, 9, 4);
...
+    <URL>https://mathworld.wolfram.com/HaarGraph.html</URL>:
+    <P/>
+
+    <Q>A Haar graph is a bipartite regular graph indexed by a positive integer.
+    The number of vertices in the resulting graph is equal to the number of digits 
+    required to represent the indexing integer in binary. The vertices are 
+    then split into two groups and connected based on the binary representation
+    of the indexing integer. The two groups can be thought of as two vertical 
+    columns opposite each other. To check which vertices are connected, each vertex in
+    the first collumn is checked in turn. For each of these vertices the binary
```
flsmith, line 331: collumn -> column

doc/examples.xml
```diff
@@ -309,3 +309,49 @@ gap> GeneralisedPetersenGraph(IsMutableDigraph, 9, 4);
...
+  <Returns>A digraph.</Returns>
+  <Description>
+    If <A>n</A> is a positive integer then this operation returns
+    the <E>Haar graph</E> <M>H(<A>n</A>)</M>. <P/>
+
+    From
+    <URL>https://mathworld.wolfram.com/HaarGraph.html</URL>:
+    <P/>
+
+    <Q>A Haar graph is a bipartite regular graph indexed by a positive integer.
```
flsmith, line 325: This paragraph isn't a quote from this URL!

doc/examples.xml
```diff
@@ -309,3 +309,49 @@ gap> GeneralisedPetersenGraph(IsMutableDigraph, 9, 4);
...
+    The number of vertices in the resulting graph is equal to the number of digits 
+    required to represent the indexing integer in binary. The vertices are 
+    then split into two groups and connected based on the binary representation
+    of the indexing integer. The two groups can be thought of as two vertical 
+    columns opposite each other. To check which vertices are connected, each vertex in
+    the first collumn is checked in turn. For each of these vertices the binary 
+    representation is mapped onto the second collumn, the 1s place coinciding with 
+    the vertex directly opposite the original vertex. An edge is added from the vertex
+    being checked to each vertex that is mapped to a 1 in the binary representation.</Q>
```
flsmith, line 334: I think this isn't totally clear, and I would probably replace it with something like 
```
The <M>i</M>th vertex in the first column is connected to the <M>j</M>th vertex in the second column if the binary representation of <A>n</A> has a 1 in position (<M>j-i</M> modulo <M>m</M>) + 1 from the left, where <M>m</M is the number of digits in the binary representation.
```

gap/examples.gi
```diff
@@ -370,3 +370,66 @@ GeneralisedPetersenGraphCons);
...
+end);
+
+InstallMethod(HaarGraphCons,
+"for IsImmutableDigraph, integer, int",
+[IsImmutableDigraph, IsInt],
+function(filt, n)
+  local D;
+  D := MakeImmutable(HaarGraphCons(IsMutableDigraph, n));
+  SetIsMultiDigraph(D, false);
+  SetIsSymmetricDigraph(D, true);
```
flsmith, line 406: There are some other properties you might as well set here, in particular `IsBipartiteDigraph`

gap/examples.gi
```diff
@@ -370,3 +370,66 @@ GeneralisedPetersenGraphCons);
...
+  return D;
+end);
+
+InstallMethod(HaarGraph, "for a function, integer",
```
flsmith, line 410: "for a function and an int"
or
"for a function and a pos int"

gap/examples.gi
```diff
@@ -370,3 +370,66 @@ GeneralisedPetersenGraphCons);
 
 InstallMethod(GeneralisedPetersenGraph, "for integer, integer", [IsInt, IsInt],
 {n, k} -> GeneralisedPetersenGraphCons(IsImmutableDigraph, n, k));
+
+InstallMethod(HaarGraphCons,
+"for IsMutableDigraph and one integer",
+[IsMutableDigraph, IsInt],
```
flsmith, line 376: I would probably just change `IsInt` -> `IsPosInt` here and not check in the function

gap/examples.gi
```diff
@@ -370,3 +370,66 @@ GeneralisedPetersenGraphCons);
...
+function(filt, n)
+  local D;
+  D := MakeImmutable(HaarGraphCons(IsMutableDigraph, n));
+  SetIsMultiDigraph(D, false);
+  SetIsSymmetricDigraph(D, true);
+  return D;
+end);
+
+InstallMethod(HaarGraph, "for a function, integer",
+[IsFunction, IsInt],
```
flsmith, line 411: Again, I would probably replace this `IsInt`

gap/examples.gi
```diff
@@ -370,3 +370,66 @@ GeneralisedPetersenGraphCons);
...
+HaarGraphCons);
+
+InstallMethod(HaarGraph, "for integer", [IsInt],
+{n} -> HaarGraphCons(IsImmutableDigraph, n));
+
+# TODO: Taken from semigroups...
```
flsmith, line 417: Shouldn't leave TODOs in the code, though I'm not entirely sure what the best way of resolving this is. Either we should change `SEMIGROUPS_BlistNumber` to `DIGRAPHS_BlistNumber` everywhere in Semigroups or just make it an alias of `DIGRAPHS_BlistNumber`

tst/standard/examples.tst
```diff
@@ -226,6 +226,16 @@ Error, the arguments <n> and <k> must be non-negative integers,
...
 <mutable digraph with 6 vertices, 24 edges>
 
+# DIGRAPHS_HaarGraph
+gap> HaarGraph(3);
+<immutable symmetric digraph with 4 vertices, 8 edges>
+gap> D := HaarGraph(16);
+<immutable symmetric digraph with 10 vertices, 10 edges>
+gap> IsBipartiteDigraph(D);
+true
+gap> HaarGraph(-2);
```
flsmith, line 236: If you make the filter `IsPosInt` above this test case disappears.

flsmith (changes requested):

gap/examples.gi
```diff
@@ -370,3 +370,66 @@ GeneralisedPetersenGraphCons);
...
+          DigraphAddEdge(D, [i, m + j]);
+      fi;
+    od;
+  od;
+
+  return DigraphSymmetricClosure(D);
+end);
+
+InstallMethod(HaarGraphCons,
+"for IsImmutableDigraph, integer, int",
```
flsmith, line 400: "for IsImmutableDigraph an an int"
or 
"for IsImmutableDigraph and a pos int"

gap/examples.gi
flsmith: And also change the string above if you do!

doc/examples.xml
finnbuck: Ah okay that makes more sense then, I hadn't registered that it was meant to be a citation!

gap/examples.gi
finnbuck: Ok, I will ask about it today...

gap/examples.gi
```diff
@@ -397,21 +392,22 @@ function(filt, n)
 end);
 
 InstallMethod(HaarGraphCons,
-"for IsImmutableDigraph, integer, int",
-[IsImmutableDigraph, IsInt],
+"for IsImmutableDigraph and one positive integer",
```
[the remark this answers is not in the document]
finnbuck: Ok, will do!

doc/examples.xml
```diff
@@ -309,3 +309,46 @@ gap> GeneralisedPetersenGraph(IsMutableDigraph, 9, 4);
...
+
+    From
+    <URL>https://mathworld.wolfram.com/HaarGraph.html</URL>:
+    <P/>
+
+    <Q>A Haar graph H(<A>n</A>) is a bipartite regular vertex-transitive graph indexed by a 
+    positive integer and obtained by a simple binary encoding of cyclically adjacent 
+    vertices. Haar graphs may be connected or disconnected. </Q><P/>
+
+    The number of vertices in a given haar graph is equal to the number of digits
```
wilfwilson, line 329:
```suggestion
    The number of vertices in a given Haar graph is equal to the number of digits 
```

wilfwilson (commented):
I think this is looking nice! Thanks. I'll commit these changes, and then squash, rebase and onto master, and push... and then merge if all seems well 🙂

wilfwilson:
@finnbuck I'll note that the documentation you wrote wasn't actually linked into the manual - see for reference: https://github.com/digraphs/Digraphs/pull/425/files#r592297697

But I'll do that for you now.

doc/examples.xml
```diff
@@ -309,3 +309,46 @@ gap> GeneralisedPetersenGraph(IsMutableDigraph, 9, 4);
...
+    <Q>A Haar graph H(<A>n</A>) is a bipartite regular vertex-transitive graph indexed by a 
+    positive integer and obtained by a simple binary encoding of cyclically adjacent 
+    vertices. Haar graphs may be connected or disconnected. </Q><P/>
+
+    The number of vertices in a given haar graph is equal to the number of digits 
+    required to represent the indexing integer <A>n</A> in binary. The <M>i</M>th vertex in the 
+    first column is connected to the <M>j</M>th vertex in the second column if the 
+    binary representation of <A>n</A> has a 1 in position (<M>j-i</M> modulo 
+    <M>m</M>) + 1 from the left, where <M>m</M> is the number of digits in the binary 
+    representation.<P/>
```
wilfwilson, line 334:
```suggestion
    The number of vertices in the Haar graph <M>H(<A>n</A>)</M> is
    equal to twice <M>m</M>, where <M>m</M> is the number of digits
    required to represent <A>n</A> in binary.
    These vertices are arranged into bicomponents
    <C>[1..m]</C> and <C>[m+1..2*m]</C>.
    Vertices <M>i</M> and <M>j</M> in different bicomponents are
    adjacent by a symmetric pair of edges if and only if
    the binary representation of <A>n</A> has a 1 in position
    (<M>j-i</M> modulo <M>m</M>) + 1 from the left. <P/>
```

wilfwilson (approved):
Hopefully it's good to go!

doc/examples.xml
finnbuck: Ah yeah, sorry, I accidentally wrote the number of vertices is equal to m, when it's actually twice that... The new explanation looks good to me, thanks.

PR #409, opened 2021-02-13
Add KingsGraph

finnbuck:
Create a global function for generating a [King's Graph](https://en.wikipedia.org/wiki/King%27s_graph) based on two parameters (board height and width).


gap/examples.gi
```diff
@@ -370,3 +370,33 @@ GeneralisedPetersenGraphCons);
 
 InstallMethod(GeneralisedPetersenGraph, "for integer, integer", [IsInt, IsInt],
 {n, k} -> GeneralisedPetersenGraphCons(IsImmutableDigraph, n, k));
+
+InstallGlobalFunction(DIGRAPHS_KingsGraph,
+function(n, k)
+  local D, i, j;
+  if (n < 1) or (k < 1) then
```
Peter-Ing, line 377: This could return an error if n or k is incomparable with 1, ie a function or digraph so a check should be put in place for n and k being integers before this check.

Peter-Ing (changes requested):

Peter-Ing (dismissed):
Looks good to me.

gap/examples.gd
```diff
@@ -48,3 +48,5 @@ DeclareOperation("PetersenGraph", [IsFunction]);
 DeclareConstructor("GeneralisedPetersenGraphCons", [IsDigraph, IsInt, IsInt]);
 DeclareOperation("GeneralisedPetersenGraph", [IsInt, IsInt]);
 DeclareOperation("GeneralisedPetersenGraph", [IsFunction, IsInt, IsInt]);
+
+DeclareGlobalFunction("DIGRAPHS_KingsGraph");
```
samanthaharper, line 52: Based on the conversation today, this should probably not be a global function, bur rather an operation (like Peterson above).

gap/examples.gi
```diff
@@ -370,3 +370,33 @@ GeneralisedPetersenGraphCons);
...
+    for j in [1 .. (n - 1)] do
+      DigraphAddEdge(D, [((i - 1) * n) + j, ((i - 1) * n) + j + (n + 1)]);
+    od;
+  od;
+  for i in [1 .. (k - 1)] do
+    for j in [1 .. (n - 1)] do
+      DigraphAddEdge(D, [((i - 1) * n) + j + 1, ((i - 1) * n) + j + n]);
+    od;
+  od;
+  return DigraphSymmetricClosure(D);
```
samanthaharper, line 401: Again, based on our conversation today, the standard is to return an immutable digraph, and have MutableKingsGraph return a mutable version. (James does this for Pancake as an example).

samanthaharper (commented):
Now that we are using individual functions, you will also need documentation (you can just copy the structure from another documentation entry).

gap/examples.gi
james-d-mitchell: The main function where the work happens should construct a mutable digraph, the default function just calls that function and makes the returned mutable digraph immutable, then returns that. Again I think you need to follow the same pattern as in the `Folkman` PR

doc/examples.xml
```diff
@@ -309,3 +309,46 @@ gap> GeneralisedPetersenGraph(IsMutableDigraph, 9, 4);
...
 <#/GAPDoc>
+
+<#GAPDoc Label="KingsGraph">
+<ManSection>
+  <Oper Name="KingsGraph" Arg="[filt, ]n, k"/>
+  <Returns>A digraph.</Returns>
+  <Description>
+    If <A>n</A> is a positive integer and <A>k</A> is a positive
+    integer, then this operation returns
+    the <E>King's graph</E> for an <A>n</A> by <A>k</A> board. <P/>
```
alex-konovalov, line 320: Is it "King" or "king"? Both versions are used in this ManSection.

doc/examples.xml
```diff
@@ -309,3 +309,46 @@ gap> GeneralisedPetersenGraph(IsMutableDigraph, 9, 4);
...
+
+    From
+    <URL>https://en.wikipedia.org/wiki/King%27s_graph</URL>:
+    <P/>
+
+    <Q>The king's graph represents all legal moves of the king piece in chess. 
+    Each vertex of the graph represents a square, while each edge represents 
+    a legal move. The input values determine the size of the chess board for 
+    which the king's graph is generated, with n = k = 8 representing a normal
```
alex-konovalov, line 329: Arguments here and below should be enclosed in `<A>`

gap/examples.gi
```diff
@@ -370,3 +370,53 @@ GeneralisedPetersenGraphCons);
 
 InstallMethod(GeneralisedPetersenGraph, "for integer, integer", [IsInt, IsInt],
 {n, k} -> GeneralisedPetersenGraphCons(IsImmutableDigraph, n, k));
+
+InstallMethod(KingsGraphCons,
+"for IsMutableDigraph and two integers",
+[IsMutableDigraph, IsInt, IsInt],
+function(filt, n, k)
+  local D, i, j;
+  if not IsPosInt(n) or not IsPosInt(k)  then
```
flsmith, line 379: Like in #408, I would just make the filters on all these functions `IsPosInt`

gap/examples.gi
```diff
@@ -370,3 +370,53 @@ GeneralisedPetersenGraphCons);
...
+  for i in [1 .. (k - 1)] do
+    for j in [1 .. (n - 1)] do
+      DigraphAddEdge(D, [((i - 1) * n) + j, ((i - 1) * n) + j + (n + 1)]);
+    od;
+  od;
+  for i in [1 .. (k - 1)] do
+    for j in [1 .. (n - 1)] do
+      DigraphAddEdge(D, [((i - 1) * n) + j + 1, ((i - 1) * n) + j + n]);
+    od;
+  od;
```
flsmith, line 402: these last two loops might as well be merged since they're over the same indices

tst/standard/examples.tst
```diff
@@ -226,6 +226,20 @@ Error, the arguments <n> and <k> must be non-negative integers,
...
+<immutable symmetric digraph with 64 vertices, 420 edges>
+gap> D := KingsGraph(4, 7);
+<immutable symmetric digraph with 28 vertices, 162 edges>
+gap> IsConnectedDigraph(D);
+true
+gap> D := KingsGraph(2, 2);
+<immutable symmetric digraph with 4 vertices, 12 edges>
+gap> OutNeighbors(D);     
+[ [ 2, 3, 4 ], [ 1, 3, 4 ], [ 1, 2, 4 ], [ 1, 2, 3 ] ]
+gap> KingsGraph(-4, 2);
```
flsmith, line 240: This case disappears if you make the filter `IsPosInt` above

gap/examples.gi
```diff
@@ -412,5 +412,6 @@ InstallMethod(KingsGraph, "for a function, positive integer, positive integer",
 [IsFunction, IsPosInt, IsPosInt],
 KingsGraphCons);
 
-InstallMethod(KingsGraph, "for positive integer, positive integer", [IsPosInt, IsPosInt],
+InstallMethod(KingsGraph, "for positive integer, positive integer",
```
alex-konovalov: There is no exact requirement for such strings, only some conventions,  but GAP library would rather use "for two positive integers". Unless your use is consistent for Digraphs, I'd suggest to do like GAP.

finnbuck: Oh ok, thanks for the tip! I will change it to "for two positive integers"

gap/examples.gi
```diff
@@ -412,6 +412,6 @@ InstallMethod(KingsGraph, "for a function, positive integer, positive integer",
 [IsFunction, IsPosInt, IsPosInt],
 KingsGraphCons);
 
-InstallMethod(KingsGraph, "for positive integer, positive integer", 
+InstallMethod(KingsGraph, "for positive integer, positive integer",
```
alex-konovalov: "for two positive integers" would be a better formulation.

gap/examples.gi
```diff
@@ -370,3 +370,48 @@ GeneralisedPetersenGraphCons);
...
+      DigraphAddEdge(D, [((i - 1) * n) + j, ((i - 1) * n) + j + n]);
+    od;
+  od;
+  for i in [1 .. (k - 1)] do
+    for j in [1 .. (n - 1)] do
+      DigraphAddEdge(D, [((i - 1) * n) + j, ((i - 1) * n) + j + (n + 1)]);
+      DigraphAddEdge(D, [((i - 1) * n) + j + 1, ((i - 1) * n) + j + n]);
+    od;
+  od;
+  return DigraphSymmetricClosure(D);
```
wilfwilson, line 396: I think, once `TriangularGridGraph` is merged, this function should proceed as follows:
```suggestion
  D := TriangularGridGraph(IsMutableDigraph, n, k);
  for i in [1 .. (k - 1)] do
    for j in [1 .. (n - 1)] do
      a := ((i - 1) * n) + j;
      b := ((i - 1) * n) + j + (n + 1);
      DigraphAddEdge(D, a, b);
      DigraphAddEdge(D, b, a);       
    od;
  od;
  return D;
```
This will also mean that the digraph vertex labels are properly set, to give the correspondence between vertices and grid positions.

doc/examples.xml
```diff
@@ -309,3 +309,50 @@ gap> GeneralisedPetersenGraph(IsMutableDigraph, 9, 4);
...
+
+    <Q>In graph theory, a king's graph is a graph that represents all legal 
+    moves of the king chess piece on a chessboard where each vertex represents 
+    a square on a chessboard and each edge is a legal move. More specifically, 
+    an <A>n</A> by <A>k</A> king's graph is a king's graph of an <A>n</A> by
+    <A>k</A> chessboard. It is the map graph formed from the squares of a chessboard 
+    by making a vertex for each square and an edge for each two squares that 
+    share an edge or a corner. It can also be constructed as the strong product 
+    of two path graphs.</Q>
+    <P/>
```
wilfwilson, line 334:
```suggestion
    <P/>
    
    In particular, the <C><A>n</A> * <A>k</A></C>  vertices can be arranged
    into an <A>n</A> by <A>k</A> grid such that two vertices are adjacent in
    the digraph if and only if they are orthogonally or diagonally adjacent in the grid.
    The correspondence between vertices and grid positions is given by
    <Ref Oper="DigraphVertexLabels"/>.  <P/>
    
    See also <Ref Oper="SquareGridGraph"/> and
    <Ref Oper="TriangularGridGraph"/>.
```

gap/examples.gi
```diff
@@ -370,3 +370,48 @@ GeneralisedPetersenGraphCons);
...
+
+InstallMethod(KingsGraphCons,
+"for IsImmutableDigraph and two positive integers",
+[IsImmutableDigraph, IsPosInt, IsPosInt],
+function(filt, n, k)
+  local D;
+  D := MakeImmutable(KingsGraphCons(IsMutableDigraph, n, k));
+  SetIsMultiDigraph(D, false);
+  SetIsSymmetricDigraph(D, true);
+  SetIsConnectedDigraph(D, true);
```
wilfwilson, line 407: I'm pretty sure the same properties should apply as apply to the triangular grid graph. So please add all of those missing ones.

wilfwilson (changes requested):
This looks good, thanks. I made some suggestions along the lines of what I suggested in #416 for the square and triangular grid graphs. However, my changes will depend on that PR being merged first (which should be done soon, probably tomorrow).

gap/examples.gi
finnbuck: Right, that makes sense! Another consideration is that I'll be working on graph products next, and there is an even simpler way to construct a king's graph using the strong product of two path graphs. So I'll make a note to update the Kingsgraph function once the strong product function exists.

gap/examples.gi
finnbuck: Ok, will do.

gap/examples.gi
wilfwilson: Oh that's brilliant!

gap/examples.gi
```diff
@@ -371,6 +371,55 @@ GeneralisedPetersenGraphCons);
...
+  SetIsBipartiteDigraph(D, n * k in Difference([n, k], [1]));
+  SetIsPlanarDigraph(D, true);
```
wilfwilson, line 409: Oh actually, sorry, I'm wrong about planarity! It'll be planar if n=1 or k=1 or n=k=2... what about any other combinations? If you can't work it out in a nice way, then feel free to delete this line.

wilfwilson: Perhaps it's planar if and only if `n <= 2 or k <= 2 or n = k = 3`? That's my current best guess. If you can prove this, (or prove it to be something else), then please put in the correct condition. But don't spend too long thinking about it - we can easily add it later if we subsequently work it out.

finnbuck: Right, yeah. I think the rule is probably just that it's planar when n <= 2 or k <= 2, so I'll add that condition.

finnbuck: ...and on further consideration n=k=3 is also planar so I'll add that too

wilfwilson: `n = k = 3` is also definitely planar, but I think that's the only other exception.

wilfwilson:
Just to note that since #416 has been merged into `master`, you are now free to use `SquareGridGraph` and `TriangularGridGraph`, once this branch is up to date with `master` again (which will require resolving merge conflicts - happy to help if you need it).

wilfwilson (approved):
I just (re?)linked the documentation into the manual and pushed that, along with my two documentation suggestions. So if the CI passes for one final time, I think it's time to merge!

wilfwilson (approved):
Thanks for that extra change. In my opinion this is good enough to be merged now, so I'll do that.

Of course, if now or at a later point there are further changes that you'd like to make, you're always welcome to make a further pull request 🙂 

Thanks for this contribution.

finnbuck:
Awesome, thanks for the merge!

PR #412, opened 2021-02-17
init: only define GRAPE variables when undefined

wilfwilson:
This should remove one of the two remaining incompatibilities between Digraphs and YAGS; see https://github.com/yags/yags/issues/1#issuecomment-780504933.


james-d-mitchell (approved):
Good!

Issue #413, opened 2021-02-17
Improve performance of DigraphSymmetricClosure and OutNeighbours on mutable digraphs


reiniscirpons:
Consider the following timing comparison for taking the symmetric closure of an immutable and mutable version of a digraph:
```
gap> DigraphSymmetricClosure(CycleDigraph(IsImmutableDigraph, 10000));; time;
33
gap> DigraphSymmetricClosure(CycleDigraph(IsMutableDigraph, 10000));; time;
6174
```
Creating the mutable and immutable version of the cycle graph takes about the same time, and the cycle digraph is not symmetric, so the time discrepancy is not due to the immutable version returning a stored attribute. This is an example of a mutable digraph taking more time on a simple modification than its immutable counterpart, which goes against the main purpose of mutable digraphs - to provide efficient modification operations.

According to  #318, the issue is most likely due to  `OutNeighbours` being copied every time it is accessed, as it is an attribute not a function, and mutable digraphs do not store attributes. I would argue that, since `OutNeighbours` (or some similar variable storing neighbour information) is dynamically updated as the mutable digraph changes, allowing more efficient access to it is essential, since some nontrivial modifications we may want to make require knowledge of a vertices local neighbourhood (for example, swapping edges, or taking symmetric closure).

This issue comes in relation to the current named graphs project, part of which requires implementing infinite families of graphs - in some cases the graphs can be created by taking products of existing structures and making small modifications to them (for example, swapping some edges in the product etc), which is a prime use case for the mutable graphs, and would benefit from the speedup associated to making `OutNeighbours` more efficient, and the graphs we are implementing also generally tend to be symmetric, so having `DigraphSymmetricClosure` be faster also helps.

wilfwilson:
For me, oon the `stable-1.4` and `master` branches now:
```
gap> DigraphSymmetricClosure(CycleDigraph(IsImmutableDigraph, 10000));; time;
14
gap> DigraphSymmetricClosure(CycleDigraph(IsMutableDigraph, 10000));; time;
17
```
(Roughly). Aka resolved!

wilfwilson:
Resolved in v1.4.1.

Issue #318, opened 2020-05-04
IsDigraphHomomorphism is significantly slower for mutable graphs


jjonusas:
```
gap> g := CompleteDigraph(100);;  
gap> g2 := DigraphMutableCopy(g);;
gap> IsDigraphHomomorphism(g, g, (1, 12));; time;
19
gap> IsDigraphHomomorphism(g2, g2, (1, 12));; time;
753
```

james-d-mitchell:
Thanks @jjonusas, it seems that the issue here is with OutNeighbours the output of which, maybe because OutNeighbours is an attribute,  is copied every time it is called (9900 times in the code above). This can be fixed by making OutNeighbours a function instead of an attribute, must think a bit more about whether or not that’s a good idea.

james-d-mitchell:
If we do:

~~~~
gap> for i in [1 .. 9900] do
> OutNeighbours(g);
> od;
gap> time;
2
~~~~

but if we do:

~~~~
gap> g := DigraphMutableCopy(g);
<mutable digraph with 100 vertices, 9900 edges>
gap> for i in [1 .. 9900] do
> OutNeighbours(g);
> od;
gap> time;
607
~~~~

So, the only reason I can I think of **not** to make `OutNeighbours` a function, is that at present with `OutNeighbours` an attribute, the return value is immutable, and so too are each of the sublists it contains. If `OutNeighbours` becomes a function, then the return value is mutable, and so too are each of its sublists. Any thoughts @wilfwilson?

james-d-mitchell:
I'm fairly sure that if the return value of `OutNeighbours` is mutable, and it is changed outside of the relevant Digraphs functions (`DigraphAddVertex` etc), then the digraph will be corrupted, but obviously what @jjonusas reports is also bad

wilfwilson:
It’s bad, but imo it’s also not really how mutable digraphs are intended to be used. The point of mutability is to make it cheaper to modify/build digraphs. But once you want to do really any kind of computation, you should make it immutable. That’s my initial gut feeling - but perhaps I’m missing something.

jjonusas:
The reason I've encountered this is that ChromaticNumber makes a mutable copy of the graph, which would appear to go against the intended use case.

wilfwilson:
Ah! Good to know.

james-d-mitchell:
[...]

PR #419, opened 2021-02-28
Resolve Issue #413 (OutNeighbours performance)

james-d-mitchell:
This PR makes `OutNeighbours` a function rather than an attribute, to avoid the performance penalty of the return value being copied for mutable digraphs. This might not be desirable, but I can't think why it isn't at the moment, and the fact that the test all run after this change at least partially indicates that it might be desirable. 

This resolves both #413 and #318.


wilfwilson (approved):
I've not checked that this is actually faster but I'm happy to accept that it is! I also can't think of a downside.

PR #422, opened 2021-03-01
Update edge-addition-planarity-suite to v3.0.1.0

james-d-mitchell:


extern/edge-addition-planarity-suite-Version_3.0.1.0/c/.deps/planarity.Po
```diff
@@ -0,0 +1,292 @@
+c/planarity.o: c/planarity.c c/planarity.h \
+  /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/string.h \
```
wilfwilson, line 2: Is this file supposed to be included? I don't know much about this sort of thing, but since this is Mac-specific stuff, I wonder if it's something specific to your computer in particular?

james-d-mitchell: Right you are, I've checked in all of the files (including the executables), which I should not have done...

james-d-mitchell:
I think this is ready to go @wilfwilson, any further comments?

wilfwilson (approved):

PR #423, opened 2021-03-01
Add TadpoleDigraph

LRacine:
This creates a tadpole graph, which consists of a cycle graph and a path graph connected by a bridge (as described by https://en.wikipedia.org/wiki/Tadpole_graph).


doc/examples.xml
```diff
@@ -309,3 +309,36 @@ gap> GeneralisedPetersenGraph(IsMutableDigraph, 9, 4);
...
+
+    If the optional first argument <A>filt</A> is present, then this should
+    specify the category or representation the digraph being created will
+    belong to. For example, if <A>filt</A> is <Ref Filt="IsMutableDigraph"/>,
+    then the digraph being created will be mutable, if <A>filt</A> is <Ref
+    Filt="IsImmutableDigraph"/>, then the digraph will be immutable. 
+    If the optional first argument <A>filt</A> is not present, then <Ref
+      Filt="IsImmutableDigraph"/> is used by default.<P/>
+
+    The <E>Tadpole digraph</E> consists of a cycle graph on m vertices and a path graph on
```
alex-konovalov, line 329: In this paragraph, m and n are arguments, so should be enclosed in `<A>` tags.

doc/examples.xml
```diff
@@ -309,3 +309,36 @@ gap> GeneralisedPetersenGraph(IsMutableDigraph, 9, 4);
...
+    If the optional first argument <A>filt</A> is present, then this should
+    specify the category or representation the digraph being created will
+    belong to. For example, if <A>filt</A> is <Ref Filt="IsMutableDigraph"/>,
+    then the digraph being created will be mutable, if <A>filt</A> is <Ref
+    Filt="IsImmutableDigraph"/>, then the digraph will be immutable. 
+    If the optional first argument <A>filt</A> is not present, then <Ref
+      Filt="IsImmutableDigraph"/> is used by default.<P/>
+
+    The <E>Tadpole digraph</E> consists of a cycle graph on <A>m</A> vertices and a path graph on
+    <A>n</A> vertices. These two graphs are connected by a bridge.<P/>
```
wilfwilson, line 330: Please put the description of the tadpole digraph before the previous paragraph (the one all about mutability).

Please can you also link to the Wikipedia entry https://en.wikipedia.org/wiki/Tadpole_graph (with `<URL>` tags).

You should also refer to the terms `cycle digraph` and `chain digraph` rather than `cycle graph` and `path graph`, since these are the standard terms we use in the Digraphs package. The term `bridge` is also undefined in this package, so perhaps you could avoid this terminology and just explicitly say what the bridge is.

Indeed, I think overall you should also be more explicit about which vertices and edges correspond to which part of the tadpole. So something along the lines of:

```
The <E>Tadpole digraph</E> is the symmetric closure of the disjoint union
of the cycle digraph on <C>[1..<A>m</A>]</C> (the 'head' of the tadpole)
and the chain digraph on <C>[<A>m</A>+1..<A>m</A>+<A>n</A>]</C>
(the 'tail' of the tadpole), along with the additional
edges <C>[1, <A>m</A>+1]</C> and <C>[1, <A>m</A>+1]</C>
which connect the 'head' and the 'tail'. <P/>

See <Ref Oper="DigraphSymmetricClosure"/>, <Ref Oper="DigraphDisjointUnion"/>,
<Ref Oper="CycleDigraph"/>, and <Ref Oper="ChainDigraph"/>.
```
So I suppose I am also suggesting that you move the bridge from connecting the vertices `m` and `n + m` to connecting the vertices `1` and `m + 1` - but this is unimportant (as long as whatever you choose is documented correctly 🙂).

tst/standard/examples.tst
```diff
@@ -226,6 +226,20 @@ Error, the arguments <n> and <k> must be non-negative integers,
 gap> JohnsonDigraph(IsMutableDigraph, 4, 2);
 <mutable digraph with 6 vertices, 24 edges>
 
-#
+# TadPoleDigraph
+gap> TadpoleDigraph(2, 2);
+Error, m needs to greater than 2
+gap> TadpoleDigraph(3, -1);
+Error, n needs to be positive (or 0)
```
wilfwilson, line 233: This error doesn't immediately tell the user which argument needs to be greater than 2. Perhaps the error messages you could be something more like:
* `the first numeric argument <m> must be an integer greater than 2`
and
* `the final argument <n> must be a non-negative integer`

gap/examples.gi
```diff
@@ -370,3 +370,47 @@ GeneralisedPetersenGraphCons);
 
 InstallMethod(GeneralisedPetersenGraph, "for integer, integer", [IsInt, IsInt],
 {n, k} -> GeneralisedPetersenGraphCons(IsImmutableDigraph, n, k));
+
+InstallMethod(TadpoleDigraphCons, "for IsMutableDigraph and two integers",
+[IsMutableDigraph, IsInt, IsInt],
+function(filt, m, n)
+  local i, j, tail, graph;
+  if (m < 3) then
```
wilfwilson, line 378: Brackets aren't required in this situation in GAP, and we typically avoid them unless we really need them for clarity.
```suggestion
  if m < 3 then
```

gap/examples.gi
```diff
@@ -370,3 +370,47 @@ GeneralisedPetersenGraphCons);
...
+    ErrorNoReturn("m needs to greater than 2");
+  elif (n < 0) then
+    ErrorNoReturn("n needs to be positive (or 0)");
+  fi;
+  graph := CycleDigraph(IsMutable, m);
+  DigraphAddEdge(graph, [1, m]);
+  for j in [2 .. m] do
+    DigraphAddEdge(graph, [j, j - 1]);
+  od;
```
wilfwilson, line 387: This is a shorter way of doing this
```suggestion
  graph := DigraphSymmetricClosure(CycleDigraph(IsMutableDigraph, m));
```

gap/examples.gi
```diff
@@ -370,3 +370,47 @@ GeneralisedPetersenGraphCons);
...
+  if (n = 0) then
+    return graph;
+  fi;
+  tail := ChainDigraph(IsMutable, n);
+  for i in [2 .. n] do
+    DigraphAddEdge(tail, [i, i - 1]);
+  od;
```
wilfwilson, line 394: And here:
```suggestion
  tail := DigraphSymmetricClosure(ChainDigraph(IsMutable, n));
```

doc/examples.xml
```diff
@@ -309,3 +309,36 @@ gap> GeneralisedPetersenGraph(IsMutableDigraph, 9, 4);
...
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="TadpoleDigraph">
+<ManSection>
+  <Oper Name="TadpoleDigraph" Arg="[filt, ]m, n"/>
+  <Returns>A digraph.</Returns>
+  <Description>
+    If <A>m</A> and <A>n</A> are non-negative integers (where <A>m</A>> 2), then this operation
+    returns a Tadpole digraph.<P/>
```
wilfwilson, line 319: In fact I don't think you should allow `n` to be `0` - I don't think it makes sense. In any case you would need to define what the tadpole digraph would look like with n = 0 (is there a bridge or not?) but that would be a bit arbitrary.

gap/examples.gd
```diff
@@ -48,3 +48,7 @@ DeclareOperation("PetersenGraph", [IsFunction]);
 DeclareConstructor("GeneralisedPetersenGraphCons", [IsDigraph, IsInt, IsInt]);
 DeclareOperation("GeneralisedPetersenGraph", [IsInt, IsInt]);
 DeclareOperation("GeneralisedPetersenGraph", [IsFunction, IsInt, IsInt]);
+
+DeclareConstructor("TadpoleDigraphCons", [IsDigraph, IsInt, IsInt]);
+DeclareOperation("TadpoleDigraph", [IsInt, IsInt]);
+DeclareOperation("TadpoleDigraph", [IsFunction, IsInt, IsInt]);
```
wilfwilson, line 54: If you're happy to disallow the second argument to be `0`, then please make the following change, and similarly please update the method installations to be `IsPosInt` rather than `IsInt` (and correspondingly, the error message for `n < 0` and any code relating to `n = 0` can go).
```suggestion
DeclareConstructor("TadpoleDigraphCons", [IsDigraph, IsPosInt, IsPosInt]);
DeclareOperation("TadpoleDigraph", [IsPosInt, IsPosInt]);
DeclareOperation("TadpoleDigraph", [IsFunction, IsPosInt, IsPosInt]);
```

wilfwilson (changes requested):
Thanks for this @LRacine.

As I mentioned in https://github.com/digraphs/Digraphs/pull/424/files#r592302744 you'll need to link your manual entry into the manual.

wilfwilson:
There are also merge conflicts now with the master branch. If you're confident enough with git to resolve them yourself, then please go ahead, otherwise let me know and I'll be happy to sort it out for you.

LRacine:
> There are also merge conflicts now with the master branch. If you're confident enough with git to resolve them yourself, then please go ahead, otherwise let me know and I'll be happy to sort it out for you.

On it

wilfwilson (approved):
Thanks for this!

LRacine:
> Thanks for this! Nice work.
Thank you for all the help @wilfwilson !

PR #424, opened 2021-03-01
Star graph

LRacine:
This creates a star graph as defined by https://en.wikipedia.org/wiki/Star_(graph_theory).

PR #425, opened 2021-03-03
Knightsgraph

finnbuck:
This adds a constructor function for the [Knight's Graph](https://en.wikipedia.org/wiki/Knight%27s_graph).

### DIFF
--- a/doc/examples.xml
+++ b/doc/examples.xml
@@ -345,3 +345,47 @@ true
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="KnightsGraph">
+<ManSection>
+  <Oper Name="KnightsGraph" Arg="[filt, ]m, n"/>
+  <Returns>A digraph.</Returns>
+  <Description>
+    If <A>m</A> and <A>n</A> are both positive integers, then this operation
+    returns the <E>Knight's Graph</E> for a <A>m</A> by <A>n</A> board. <P/>
+
+    From
+    <URL>https://en.wikipedia.org/wiki/Knight%27s_graph</URL>:
+    <P/>
+
+    <Q>In graph theory, a knight's graph, or a knight's tour graph, is a
+    graph that represents all legal moves of the knight chess piece on a
+    chessboard.  Each vertex of this graph represents a square of the
+    chessboard, and each edge connects two squares that are a knight's move
+    apart from each other. More specifically, an <A>m</A> by <A>n</A> knight's
+    graph is a knight's graph of an <A>m</A> by <A>n</A> chessboard.</Q>
+    <P/>
+
+    If the optional first argument <A>filt</A> is present, then this should
+    specify the category or representation the digraph being created will
+    belong to. For example, if <A>filt</A> is <Ref Filt="IsMutableDigraph"/>,
+    then the digraph being created will be mutable, if <A>filt</A> is <Ref
+    Filt="IsImmutableDigraph"/>, then the digraph will be immutable.
+    If the optional first argument <A>filt</A> is not present, then <Ref
+      Filt="IsImmutableDigraph"/> is used by default.<P/>
+
+<Example><![CDATA[
+gap> D := KnightsGraph(8, 8);
+<immutable connected symmetric digraph with 64 vertices, 336 edges>
+gap> IsConnectedDigraph(D);
+true
+gap> D := KnightsGraph(3, 3);
+<immutable symmetric digraph with 9 vertices, 16 edges>
+gap> IsConnectedDigraph(D);
+false
+gap> KnightsGraph(IsMutable, 3, 9);
+<mutable digraph with 27 vertices, 88 edges>
+]]></Example>
+  </Description>
+</ManSection>
+<#/GAPDoc>
--- a/doc/z-chap2.xml
+++ b/doc/z-chap2.xml
@@ -87,6 +87,7 @@
     <#Include Label="JohnsonDigraph">
     <#Include Label="PetersenGraph">
     <#Include Label="GeneralisedPetersenGraph">
+    <#Include Label="KnightsGraph">
     <#Include Label="StarDigraph">
   </Section>
 
--- a/gap/examples.gi
+++ b/gap/examples.gi
@@ -400,3 +400,49 @@ function(filt, k)
   SetIsCompleteBipartiteDigraph(D, k > 1);
   return D;
 end);
+
+InstallMethod(KnightsGraphCons,
+"for IsMutableDigraph and two positive integers",
+[IsMutableDigraph, IsPosInt, IsPosInt],
+function(filt, m, n)
+  local D, moveOffSets, coordinates, target, i, j, iPos;
+  D := EmptyDigraph(IsMutableDigraph, m * n);
+  moveOffSets := [[2, 1], [-2, 1], [2, -1], [-2, -1],
+  [1, 2], [-1, 2], [1, -2], [-1, -2]];
+  coordinates := [];
+  for i in [1 .. n] do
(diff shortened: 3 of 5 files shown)


doc/examples.xml
```diff
@@ -309,3 +309,44 @@ gap> GeneralisedPetersenGraph(IsMutableDigraph, 9, 4);
...
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="KnightsGraph">
+<ManSection>
+  <Oper Name="KnightsGraph" Arg="[filt, ]n, k"/>
+  <Returns>A digraph.</Returns>
+  <Description>
+    If <A>n</A> and <A>k</A> are both positive integers, then this operation returns
+    the <E>Knight's Graph</E> for a n by k board. <P/>
```
alex-konovalov, line 319: Do we need to use capitalisation here?

doc/examples.xml
```diff
@@ -309,3 +309,44 @@ gap> GeneralisedPetersenGraph(IsMutableDigraph, 9, 4);
...
+
+    From
+    <URL>https://en.wikipedia.org/wiki/Knight%27s_graph</URL>:
+    <P/>
+
+    <Q>The knights graph for a given n by k board size consists all of n * k vertices, where
```
alex-konovalov, line 325: Please use `<A>` for arguments.

wilfwilson:
Currently the PR is not totally covered by tests, which is why the 'codecov/patch' check is failing.

You can find out (for instance here: https://app.codecov.io/gh/digraphs/Digraphs/compare/425/diff) the bits of your added/changed code that don't have code coverage (you get there by following the "Details" link next to the failing 'codecov/patch' PR check, which will take you to https://github.com/digraphs/Digraphs/pull/425/checks?check_run_id=2026949875 – that gives a summary of the missing lines. If you want to access the details on the Codecov website, which displays things visually, then you will need to log in with your GitHub account on by clicking image in the top-right of the Codecov website, and selecting Log in).

In particular, it's just the error message that isn't tested 🙂 Please add a test that shows that the error message is triggered correctly and the test should pass.

doc/examples.xml
wilfwilson, line 325: This formulation makes me think that you're giving a quotation (which is what `<Q>` tags mean, I think), but it's not a quote! Perhaps you could do something like:

As explained on {URL}, the knight's graph... etc

gap/examples.gi
```diff
@@ -370,3 +370,51 @@ GeneralisedPetersenGraphCons);
...
+end);
+
+InstallMethod(KnightsGraphCons,
+"for IsImmutableDigraph and two positive integers",
+[IsImmutableDigraph, IsPosInt, IsPosInt],
+function(filt, n, k)
+  local D;
+  D := MakeImmutable(KnightsGraphCons(IsMutableDigraph, n, k));
+  SetIsMultiDigraph(D, false);
+  SetIsSymmetricDigraph(D, true);
```
wilfwilson, line 411: Could you work out precisely when `D` is connected? It's probably something like
```suggestion
  SetIsSymmetricDigraph(D, true);
  SetIsConnectedDigraph(D, n > 2 and m > 2 and not (m = 3 and n = 3));
```
That would be nice to add.

gap/examples.gi
```diff
@@ -370,3 +370,51 @@ GeneralisedPetersenGraphCons);
...
+  [1, 2], [-1, 2], [1, -2], [-1, -2]];
+  coordinates := [];
+  p := 0;
+
+  for i in [1 .. k] do
+    for j in [1 .. n] do
+      p := p + 1;
+      coordinates[p] := [i, j];
+    od;
+  od;
```
wilfwilson, line 390: So basically `QuotientRemainder(p, n) = [i, j]` if and only if `coordinates[p] = [i + 1, j]`? (except remainder would be 0 rather than 5)

gap/examples.gi
```diff
@@ -370,3 +370,51 @@ GeneralisedPetersenGraphCons);
...
+
+  for i in coordinates do
+    for j in moveOffSets do
+      targetCoordinate := [i[1] + j[1], i[2] + j[2]];
+      if targetCoordinate in coordinates then
+        DigraphAddEdge(D, [Position(coordinates, i),
+        Position(coordinates, targetCoordinate)]);
+      fi;
+    od;
+  od;
```
wilfwilson, line 400: This looks like a fairly inefficient way of doing things, although I suppose we don't need to be too worried about the performance of these named-digraph creation functions 🙂 

In particular, you shouldn't need to search in `coordinates` for each `targetCoordinate`, because you know that `[a,b] in coordinates` if and only if `a in [1 .. k] and b in [1 .. n]`.

Also, you shouldn't need to search for the position of `i` in `coordinates` (because you're selecting `i` from `coordinates` in the first instance, so you could keep track of where it is when you select it), and you shouldn't need to search for `targetCoordinate` in `coordinates` because you could use my remark about `QuotientRemainder` to compute the position directly (by multiplying and adding).

wilfwilson (commented):
Thanks for this @finnbuck! I've made a few remarks, I hope you don't mind addressing them (or arguing why you don't want to) 🙂.

flsmith (dismissed):
Only some minor changes

gap/examples.gi
```diff
@@ -370,3 +370,51 @@ GeneralisedPetersenGraphCons);
...
+      coordinates[p] := [i, j];
+    od;
+  od;
+
+  for i in coordinates do
+    for j in moveOffSets do
+      targetCoordinate := [i[1] + j[1], i[2] + j[2]];
+      if targetCoordinate in coordinates then
+        DigraphAddEdge(D, [Position(coordinates, i),
+        Position(coordinates, targetCoordinate)]);
```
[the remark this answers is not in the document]
finnbuck: That would probably be more clear, yeah. I think according to the linting tests that would exceed the line length limit though... Is there another way to go about it?

gap/examples.gi
finnbuck: Ok, thanks for the tip, that seems like a good idea! The new version doesn't search through coordinates anymore.

doc/examples.xml
wilfwilson: I agree on the consistency front. Note that here, and in the arguments, we use `n` and `k` for the parameters of the knight's graph, but in the Wikipedia quotation, we use `m` and `n`. I think it would be best for these to match. (Especially as the `m` and `n` in the Wikipedia text already have `<A>` tags around them, which indicates that those are the names of the arguments.

doc/z-chap2.xml
```diff
@@ -87,6 +87,7 @@
     <#Include Label="JohnsonDigraph">
     <#Include Label="PetersenGraph">
     <#Include Label="GeneralisedPetersenGraph">
+    <#Include Label="KnightsGraph">
```
wilfwilson, line 90: When you write a new piece of documentation, you also have to  explicitly link it include the manual by including it in one of the chapters, i.e. in one of the `doc/z-chap*.xml` files. Sometimes it's difficult to work out where to put your documentation, but in this case it's easy - we put it with the other standard examples.

If you don't do this step, then your documentation doesn't end up in the compiled version of the manual, and you can't access the manual entry in GAP by doing (in this case) `?KnightsGraph`.

wilfwilson (approved):
Hi @finnbuck - since #424 was merged, this created merge conflicts between the master b ranch and your PR. I've just fixed them for you, and squashed the commit history into a single commit, and then pushed the resulting rebased branch.

There was one crucial thing that you hadn't done (because you didn't know about!) but I'll comment below so you can see what it was.

finnbuck:
Thanks Wilf! I'll remember to link the documentation to the manual in the future :)

PR #426, opened 2021-03-03
Codecov integration: replace the PR comment by PR statuses

wilfwilson:
See the 'status' bit at the bottom of this PR for the new look (it may not appear immediately, until a CI job has finished). This has the advantage that you don't get email notifications of the Codecov comment, which always initially show that code coverage has declined (even when it hasn't, once all the jobs are complete).

If we want, we eventually can make some or all of these checks be 'required' ones in order for a PR to be merged.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -2,14 +2,20 @@ coverage:
   precision: 2      # Report coverage to 2 decimal places
   round: down       # Round down
   range: "90...100" # Anything below 90% is coloured red (considered bad)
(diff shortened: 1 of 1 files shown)


wilfwilson:
It's not worked so far, I'll investigate.

wilfwilson:
Yes, that does seem to be correct: it won't work until it's merged. See https://github.com/wilfwilson/Digraphs/pull/2 for an example of what it will look like:
Screen Shot 2021-03-03 at 15 40 38

Issue #414, opened 2021-02-19
SetDigraphVertexLabel List Assignment error


LRacine:
**[gap version using:  4.12dev-638-gcd2de0f]**
**@MTWhyte found that this was due to using DigraphDisjointUnion on mutable digraphs causes D!.vertexlabels to be immutable**
After creating a digraph and calling the SetDigraphVertexLabel function on that digraph, a list assignment error appears.
![image](https://user-images.githubusercontent.com/35891709/108508713-1e1f9c80-72b4-11eb-8c98-c916e83eb148.png)

**Steps for reproduction**
1. load the function:
  ```
CreateTree := function(height)
    local graph, size;
    if (height < 2 ) then
      graph := EmptyDigraph(IsMutable, 1);
      return graph;
    fi;
    
    graph := EmptyDigraph(IsMutable, 1); 
    
    DigraphDisjointUnion(graph, CreateTree(height - 1));
    DigraphAddEdge(graph, 1, 2);
    
    size := Size(DigraphVertices(graph));
    
    DigraphDisjointUnion(graph, CreateTree(height - 2));
    DigraphAddEdge(graph, 1, size + 1);
    
    return graph;
end;
  ```
2. Create a digraph i.e G := CreateTree(5).
3. Call SetDigraphVertexLabel i.e SetDigraphVertexLabel(G, 1, "a");

MTWhyte:
I'm in the process of investigating this, and the problem seems to be that where `D` was created from running `DigraphDisjointUnion` on mutable digraphs, `D!.vertexlabels` is immutable. For example,

```
gap> G := CycleDigraph(IsMutable, 5);
<mutable digraph with 5 vertices, 5 edges>
gap> D := DigraphDisjointUnion(G, G);
<mutable digraph with 10 vertices, 10 edges>
gap> D!.vertexlabels;
[ 1 .. 10 ]
gap> IsMutable(last);
false
```

I haven't fixed it yet, but I thought it would be a good idea to note this on the issue.

MTWhyte:
The trouble seems to be in `DigraphDisjointUnion`. Starting on line `436` of `oper.gi` in its definition, there is the following:

```
  if IsMutableDigraph(D) then
    SetDigraphVertexLabels(D, DigraphVertices(D));
    # This above stops D keeping its old vertex labels after being changed
    ClearDigraphEdgeLabels(D);
    return D;
  fi;
```

[...]

PR #427, opened 2021-03-03
Copy immutable lists given to `SetDigraphVertexLabels` (Issue #414)

MTWhyte:
This PR fixes issue #414, per the discussion with @james-d-mitchell and @wilfwilson.

When calling `SetDigraphVertexLabels` with an immutable list, the list is replaced with a mutable copy of itself, so that a digraph's vertex labels can't be immutable.

### DIFF
--- a/doc/labels.xml
+++ b/doc/labels.xml
@@ -24,8 +24,8 @@
     <A>i</A>. <P/>
 
     If <A>digraph</A> is a digraph created from a record with a component
-    <C>vertices</C>, then the labels of the vertices are set to the value of
-    this component.<P/>
+    <C>DigraphVertices</C>, then the labels of the vertices are set to
+    the value of this component.<P/>
 
     Induced subdigraphs, and some other operations which create new digraphs from
     old ones, inherit their labels from their parents.
--- a/doc/labels.xml
+++ b/doc/labels.xml
@@ -71,15 +71,23 @@ gap> DigraphVertexLabel(D, 1);
     copy of the labels of the vertices in <A>digraph</A>.
     <C>SetDigraphVertexLabels</C> can be used to set the labels of the vertices
     in <A>digraph</A> to the list of
-    arbitrary &GAP; objects <A>list</A>. <P/>
+    arbitrary &GAP; objects <A>list</A>, which must be of the same length
+    as the number of vertices of <A>digraph</A>. <P/>
+
+    If the list <A>list</A> is immutable, then the vertex labels are set to a
+    mutable copy of <A>list</A>. Otherwise, the labels are set to exactly
+    <A>list</A>. <P/>
 
     The label of a vertex can be changed an arbitrary number of times. If no
     label has been set for the vertex <A>i</A>, then the default value is
     <A>i</A>. <P/>
 
     If <A>digraph</A> is a digraph created from a record with a component
-    <C>vertices</C>, then the labels of the vertices are set to the value of
-    this component.<P/>
+    <C>DigraphVertices</C>, then the labels of the vertices are set to the
+    value of this component. As in the above, if the component is immutable
+    then the digraph's vertex labels are set to a mutable copy of
+    <C>DigraphVertices</C>. Otherwise, they are set to exactly
+    <C>DigraphVertices</C>. <P/>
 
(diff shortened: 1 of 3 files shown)


wilfwilson (commented):
Thanks! As discussed on Teams, it would be great if you included one or more tests that verify that the previous errors no longer happened, and if you updated the manual entry for `SetDigraphVertexLabels` to document the new behaviour.

tst/standard/labels.tst
```diff
@@ -79,6 +79,14 @@ gap> DigraphVertexLabels(gr);
...
 gap> D := NullDigraph(5);;
 gap> RemoveDigraphVertexLabel(D, 6);
+gap> A := Immutable([1, 2]);
+[ 1, 2 ]
+gap> D := CycleDigraph(2);
+<immutable cycle digraph with 2 vertices>
+gap> SetDigraphVertexLabels(D, A);
+gap> SetDigraphVertexLabels(D, ["a", "b"]);
+gap> DigraphVertexLabels(D);
+[ "a", "b" ]
```
wilfwilson, line 89: This does not capture the previous 'bad' behaviour. Indeed, in the current `stable-1.4` branch, this already works totally fine:
```
gap> A := Immutable([1, 2]);
[ 1, 2 ]
gap> D := CycleDigraph(2);
<immutable cycle digraph with 2 vertices>
gap> SetDigraphVertexLabels(D, A);
gap> SetDigraphVertexLabels(D, ["a", "b"]);
gap> DigraphVertexLabels(D);
[ "a", "b" ]
```
Rather, the problem was that the immutable list of vertex labels meant that you couldn't change _individual_ vertex labels (or proper subsets of labels):
```
gap> A := Immutable([1, 2]);;
gap> DigraphVertexLabels(D);;
gap> SetDigraphVertexLabels(D, A);
gap> SetDigraphVertexLabel(D, 2, "b");
Error, List Assignment: <list> must be a mutable list in
  D!.vertexlabels[v] := name; at /Users/Wilf/gap/pkg/digraphs/gap/labels.gi:35 called from
<function "SetDigraphVertexLabel for a digraph, pos int, object">( <arguments>\
 )
 called from read-eval loop at *stdin*:14
brk>
```
So could you please update the test to reflect this use case, i.e. the test should be:
```
gap> A := Immutable([1, 2]);;
gap> D := CycleDigraph(2);
gap> SetDigraphVertexLabels(D, A);
gap> IsMutable(DigraphVertexLabels(D));
true
gap> SetDigraphVertexLabel(D, 2, "b");
gap> DigraphVertexLabels(D);
[ 1, "b" ]
```

doc/labels.xml
```diff
@@ -73,6 +73,10 @@ gap> DigraphVertexLabel(D, 1);
     in <A>digraph</A> to the list of
     arbitrary &GAP; objects <A>list</A>. <P/>
 
+    If the list <A>list</A> is immutable, then the vertex labels are set to a
+    mutable copy of <A>list</A>. Otherwise, the labels are set to exactly
+    <A>list</A>. <P/>
+
```
wilfwilson, line 79: While you're at it here, could you specific that `<A>list</A>` (when it is introduced) must be the same length of the number of vertices of `<A>digraph</A>`, and also below, when it talks about `<C>vertices</C>`, you could note the analogous thing about mutable copies, there (if indeed a mutable copy is being made in that case; if it's not then it should be!)

wilfwilson (changes requested):

tst/standard/labels.tst
MTWhyte: Ah yes, sorry! It actually appears I made a typo in my discussion on the issue: I'd written `gap> SetDigraphVertexLabel(D, ["a", "b"]);`, and clearly this didn't work because the function should've been `SetDigraphVertexLabels` instead, and I just copied that over from the issue to the tests but with the correct function. I'll fix that now!

wilfwilson (approved):
Great, thank you. Let's hope the tests pass 🙂

wilfwilson (approved):
I've rebased this onto `stable-1.4` rather than `master`, since it's arguably a bugfix, and is more likely to be included in a release of Digraphs soon 🙂 I'll merge if the tests pass.

PR #429, opened 2021-03-03
Update GitHub actions integration

wilfwilson:
This does two things:

1. It implements the linting job along the lines of what Semigroups now does (although, see `.github/workflows/lint.yml`, it perhaps does it a slightly nicer way, combining the two jobs with one matrix).
2. It adds a new `if:` command to the workflows to skip some duplicate jobs on PRs. This means that when a PR is made from a branch in the same repository (e.g. this happens with our `rc-v1.4.1` branches), then the jobs are not run twice (as both a pushed branch, and as a PR). In particular, this skips the jobs associated with the PR, and keeps the jobs for the push. Previously we handled this case with an exception for `rc-v[0-9]+.[0-9]+.[0-9]+` branches, but this is a general solution for any name of branch.
The result is not completely pretty, but [it looks reasonable I think](https://github.com/wilfwilson/Digraphs/pull/1); see https://github.com/wilfwilson/Digraphs/pull/1:

Screen Shot 2021-03-03 at 17 18 36

### DIFF
--- a/.github/workflows/gap.yml
+++ b/.github/workflows/gap.yml
@@ -1,32 +1,12 @@
-name: "CI"
-
-# Trigger the workflow on push or pull request
-on:
-  pull_request:
-  push:
-    branches-ignore:
-      - rc-v[0-9]+.[0-9]+.[0-9]+
+name: "Digraphs tests"
+on: [push, pull_request]
 
 jobs:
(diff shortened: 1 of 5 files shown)


james-d-mitchell (approved):
Approved!

Issue #436, opened 2021-03-10
DigraphCore can give wrong results


wilfwilson:
From the documentation:
> The  core  of  a  digraph  D  is  the  minimal  subdigraph C of D which is a homomorphic image of D. The core of a digraph is unique up to isomorphism.

Let's create the following digraph and compute its core.
```
gap> D := DigraphFromDigraph6String("&GSY??A?SA?O?");
<immutable digraph with 8 vertices, 10 edges>
gap> DigraphCore(D);
[ 1, 2, 3 ]
gap> core := InducedSubdigraph(D, DigraphCore(D));
<immutable digraph with 3 vertices, 3 edges>
```
Alright,let's check that there is actually a homomorphism from `D` to `core`:
```
gap> DigraphHomomorphism(D, core) <> fail;
true
```
Good. But, it's not minimal...
```
gap> smaller_core := InducedSubdigraph(D, [1 ,2]);
<immutable digraph with 2 vertices, 2 edges>
gap> IsCompleteDigraph(smaller_core);
true
gap> DigraphHomomorphism(D, smaller_core) <> fail;
true
```
Indeed, `IsDigraphCore` tells us that the core isn't actually a core, directly:
```
gap> IsDigraphCore(core);
false
```

So unless I'm misunderstanding something, the result of `DigraphCore` for `D` is wrong; in particular, it should be `[1, 2]`.

wilfwilson:
I'm pretty sure that, in general:
```
DigraphCore(D) = ImageSetOfTransformation(RepresentativeOfMinimalIdeal(EndomorphismMonoid(D)));
```
(Not that this is necessarily the best way to check it, but it should give us a good way of writing test cases/working out whether this bug is fixed.)

wilfwilson:
Here's a much simpler example where it goes wrong:
```
gap> D := Digraph([[2], [1, 3], []]);;
gap> DigraphCore(D);
[ 1, 2, 3 ]
gap> DigraphHomomorphism(D, InducedSubdigraph(D, [1, 2]));
Transformation( [ 2, 1, 2 ] )
gap> IsDigraphCore(D);
false
```

wilfwilson:
Resolved in v1.4.1.

PR #437, opened 2021-03-10
Fix DigraphCore

wilfwilson:
See #436.

The generic part of the method for `DigraphCore` assumes a lower bound of there being three vertices in the core.

The core can't contain only one vertex, since this can only happen when the digraph is either empty, or has a loop - but by the time the generic part of the method is encountered, these cases have been excluded. However, I don't see a good reason to rule out the core having only two vertices.

In any case, changing the default lower bound from should not affect mathematical correctness of results - the only problem is that we're widening the search size for the core. But the test cases that I've added show that this is necessary, anyway (at least in general).

### DIFF
--- a/gap/attr.gi
+++ b/gap/attr.gi
@@ -1349,6 +1349,12 @@ function(digraph)
   elif IsCompleteDigraph(digraph) then
     return DigraphVertexLabels(digraph);
   elif IsSymmetricDigraph(digraph) and IsBipartiteDigraph(digraph) then
+    # TODO symmetric is not necessary, you just need bipartite and:
+    # DigraphGirth(digraph) = 2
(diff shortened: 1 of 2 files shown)


wilfwilson:
Oh, perhaps because the core will have two vertices if and only if the digraph is non-empty and bipartite? (I think?)

Maybe this was special case that it was assumed had been implemented, but actually hadn't? Excluding bipartite graphs would allow the general lower bound to be 3, I think.

wilfwilson:
Actually, a fix involving bipartite digraphs that would allow you to keep a lower bound of 3 would require a bit more work. So I'll leave that enhancement for a different PR/someone else to do.

james-d-mitchell (approved):
Also approved

PR #440, opened 2021-03-11
ViewString: special case for trees and forests

wilfwilson:
I realised that undirected trees and forests didn't know that they were bipartite, so I've added this implication via an `InstallTrueMethod`. `UndirectedSpanningTree`s also didn't know that they were undirected trees, so I've added this too.

It then became a bit weird that these digraphs would be shown as `<immutable bipartite symmetric connected  digraph...>` or something like that.

The `ViewString`s now say "...forest digraph..." and "...tree digraph..." - should I drop the 'digraph'? I thought it best to keep it in, so that it's clear to a user that these are digraph objects, but I'm not strongly of that opinion.


james-d-mitchell (approved):
Excellent

PR #442, opened 2021-03-15
display: simplify some `Dot...` methods, and use StringFormatted

wilfwilson:
* This addresses a TODO; `StringFormatted` was only added in GAP 4.10, but we now require it, so we can use it.
* Some of the `else` blocks were unreachable and had syntax errors.
* I've made the code in some of the methods a bit clearer, I think.


james-d-mitchell (approved):
Nice!

PR #445, opened 2021-03-18
Add `BookDigraph`

LRacine:
Another infinite family graph, this adds the Book graph as denoted by: https://mathworld.wolfram.com/BookGraph.html

This graph will be used for the implementation of the StackedBookDigraph later.


doc/examples.xml
```diff
@@ -703,3 +703,45 @@ true
...
+  <Returns>A digraph.</Returns>
+  <Description>
+    The <E>Book digraph</E> is (generally) the cartesian product of the chain digraph
+    (as the "Book Spine") and the star digraph (as the "Pages"). When <A>m</A> is less than 3
+    two special cases occur. When <A>m</A> is 1, a (<A>2</A>, <A>2</A>) complete bipartite
+    graph is returned (the "square" graph). When <A>m</A> is 2, a (<A>2</A>, <A>3</A>)
+    square grid graph is returned (the "domino" graph). For more details on the Book digraph
+    please refer to <URL>https://mathworld.wolfram.com/BookGraph.html</URL>.<P/>
+
+    See <Ref Oper="DigraphCartesianProduct"/>,
```
wilfwilson, line 718: As can be seen on the GitHub Actions artifact (containing the compiled documentation), this reference doesn't resolve (note the `???`).

Screenshot of the documentation:

Screenshot 2021-03-18 at 21 00 10
This is probably because there are actually multiple bits of documentation with the name `DigraphCartesianProduct`, and so you need to specify which one you're referencing.

## Fix

To work out how to properly reference `DigraphCartesianProduct`, we should search the Digraphs package's `doc/` folder for the place where `DigraphCartesianProduct` is documented. We then learn that is is documented in `doc/oper.xml`, as follows:
```
<#GAPDoc Label="DigraphCartesianProduct">
<ManSection>
  <Func Name="DigraphCartesianProduct" Arg="gr1, gr2, ..."
    Label="for a positive number of digraphs"/>
  <Func Name="DigraphCartesianProduct" Arg="list"
    Label="for a list of digraphs"/>
  <Returns>A digraph.</Returns>
  <Description>
```
... and so on.

So in fact it is a `Func` rather than an `Oper`, and there are two different labels that we can reference – since the two versions of `DigraphCartesianProduct` are documented together, it doesn't matter which one we reference, since any link will take us to this same bit of documentation. So let's just pick the first.

Therefore the fix is:
```suggestion
    See <Ref Func="DigraphCartesianProduct" Label="for a positive number of digraphs"/>,
```

doc/examples.xml
```diff
@@ -703,3 +703,45 @@ true
...
+<ManSection>
+  <Oper Name="BookDigraph" Arg="[filt, ]m"/>
+  <Returns>A digraph.</Returns>
+  <Description>
+    The <E>Book digraph</E> is (generally) the cartesian product of the chain digraph
+    (as the "Book Spine") and the star digraph (as the "Pages"). When <A>m</A> is less than 3
+    two special cases occur. When <A>m</A> is 1, a (<A>2</A>, <A>2</A>) complete bipartite
+    graph is returned (the "square" graph). When <A>m</A> is 2, a (<A>2</A>, <A>3</A>)
+    square grid graph is returned (the "domino" graph). For more details on the Book digraph
+    please refer to <URL>https://mathworld.wolfram.com/BookGraph.html</URL>.<P/>
```
wilfwilson, line 716: * In the Digraphs package we seem to capitalise the word 'Cartesian' - see e.g. the documentation for `DigraphCartesianProduct`.
* The Cartesian product of _which_ chain digraph and _which_ star digraph? (i.e. on how many vertices, respectively?) I think this should be documented.
  * (From looking at the link, the numbers are `2` and `<M><A>m</A>+1</M>`, respectively.)
* In fact, after looking at the link, I noticed that it's not the chain digraph that's involved! The chain digraph isn't a symmetric digraph; in particular, `ChainDigraph(2)` just consists of the one edge `1 → 2`. You should define it (and code it) in terms of either the complete digraph on two vertices, or the cycle digraph on two vertices (which are equal).
* I think the information about the 'special cases' is misleading. When I first read this documentation, I assumed that these 'special cases' don't fit the general pattern of being the Cartesian product of the complete digraph on 2 vertices with the star digraph on m+1 vertices. But in fact they do fit in this pattern. So they don't need to be defined separately.
    * If you really want to include this information, I think it should be a separate paragraph, and it should be clear that this is additional information given for interest, not part of the definition. (Something along the lines of "Note that `BookDigraph(1)` is a (2,2)-complete bipartite digraph, and `BookDigraph(2)` is isomorphic to the (2,3)-square grid graph"). But I would recommend just not including this.
    * Note that, in your implementation at least, `BookDigraph(2) <> SquareGridGraph(2,3)`, although they are isomorphic.

gap/examples.gi
```diff
@@ -714,3 +714,41 @@ function(filt, m, n)
...
   return D;
 end);
+
+InstallMethod(BookDigraphCons,
+  "for IsMutableDigraph and one positive integer",
+  [IsMutableDigraph, IsPosInt],
+  function(filt, m)
+    local pages, book, graph;
+    # special cases
+    if m = 1 then
```
wilfwilson, line 724: As noted above, these special cases are unnecessary, and should be removed.

gap/examples.gi
```diff
@@ -714,3 +714,41 @@ function(filt, m, n)
...
+      # "the" square graph
+      return CompleteBipartiteDigraph(filt, 2, 2);
+    elif m = 2 then
+      # domino graph
+      return SquareGridGraph(filt, 2, 3);
+    fi;
+    pages := StarDigraph(filt, m);
```
wilfwilson, line 731: In my opinion (just my preference), for the purposes of maximising clarity, since `filt` should always equal `IsMutableDigraph` in this method, then `IsMutableDigraph` should be used directly, rather than `filt`.

wilfwilson, line 731: Also, as noted above, this does not match the definition on Wolfram MathWorld. It's the star digraph on m+1 vertices, not m.

gap/examples.gi
```diff
@@ -714,3 +714,41 @@ function(filt, m, n)
...
+    book := ChainDigraph(filt, 2);
+    graph := DigraphCartesianProduct(book, pages);
+    return graph;
```
wilfwilson, line 734: No need to assign to a variable just to return it:
```suggestion
    return DigraphCartesianProduct(book, pages);
```

wilfwilson, line 734: Overall, I think the method could be fixed and simplified as follows:
```suggestion
    local book;
    book := CompleteDigraph(IsMutableDigraph, 2);
    return DigraphCartesianProduct(book, StarDigraph(IsMutableDigraph, m + 1));
```

wilfwilson (changes requested):
I found a few problems, sorry... 😬

doc/examples.xml
LRacine: I see, thank you! Is there a way to compile the documentation locally (just curious)?

wilfwilson: Yes, there's a couple of ways 🙂. Within Digraphs, you can run `DigraphsMakeDoc();`. Or, within GAP, you can read in the `makedoc.g` file of the Digraphs package via the `Read` command. 

This produces the HTML and PDF versions of the manual. So in particular I think it'll fail if you don't have `pdflatex` installed.

wilfwilson: When you compile the manual (e.g. locally), there is also a lot of output, and that output contains warnings about broken references like this one.

LRacine: Thank you for this!

gap/examples.gi
LRacine: a lot simpler 😅, thank you for this.

wilfwilson (approved):

wilfwilson:
Looks good thank you 🙂

PR #446, opened 2021-03-19
Improve ChromaticNumber with Brooks' theorem

wilfwilson:
https://en.wikipedia.org/wiki/Brooks'_theorem:
> For any connected undirected graph G with maximum degree Δ, the chromatic number of G is at most Δ, unless G is a complete graph or an odd cycle, in which case the chromatic number is Δ + 1

The greedy colouring algorithm can sometimes give Δ + 1 as an upper bound  (and maybe more?) when Brooks' theorem gives us just Δ. So let's take the minimum of the two in `ChromaticNumber`.


james-d-mitchell (approved):

james-d-mitchell:
This could be merged into stable-1.4, since I don't think it adds any new features.

PR #447, opened 2021-03-19
Improve ViewStrings/TrueMethods for tournaments/cycle digraphs

wilfwilson:
* Some properties were set to be true in the `CycleDigraph` creation function that should rather be `TrueMethods` for all things satisfying `IsCycleDigraph`.
* Strongly connected tournaments are Hamiltonian.
* Stop the `ViewString` '`<(im)mutable cycle tournament with 3 vertices>`' from happening: I think `<(im)mutable cycle digraph with 3 vertices>` is already specific enough, I think.
* Test the number of edges in `IsCycleDigraph` before testing strong connectivity, as it should be the cheaper test.

### DIFF
--- a/gap/digraph.gi
+++ b/gap/digraph.gi
@@ -562,7 +562,8 @@ function(D)
     Append(str, "multi");
   fi;
 
-  if HasIsTournament(D) and IsTournament(D) and n > 1 then
+  if not (HasIsCycleDigraph(D) and IsCycleDigraph(D))
+      and HasIsTournament(D) and IsTournament(D) and n > 1 then
     Append(str, "tournament ");
     display_nredges := false;
   else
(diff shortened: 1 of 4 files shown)


gap/digraph.gi
```diff
@@ -554,7 +554,8 @@ function(D)
     Append(str, "multi");
   fi;
 
-  if HasIsTournament(D) and IsTournament(D) and n > 1 then
+  if not (HasIsCycleDigraph(D) and IsCycleDigraph(D))
```
james-d-mitchell, line 557: What's the overlap between being a tournament and being a cycle, some corner case?

wilfwilson: Yes, a corner case. The only overlap is that a cycle digraph with three vertices is a tournament (although not all tournaments with three vertices are cycle digraphs).

This bit of code stops `<(im)mutable cycle tournament with 3 vertices>` from showing, in favour of the simpler `<(im)mutable cycle digraph with 3 vertices>`, in this case.

flsmith (approved):
Looks good.

I did think about what other properties could be set for cycle digraphs, but the answer seems to be basically all of them, which we probably don't want to do.

wilfwilson:
As far as I'm concerned, the more the merrier 🙂 But people who might want to do that can do it in a separate PR.

PR #451, opened 2021-04-08
There was a subtle mistake in the constructor of CayleyDigraph

jdebeule:
At the point where the labels are set, one cannot just use the given list of generators. Few lines of code were added to correct the problem.


jdebeule:
I made one mistake that was discovered by gaplint. I try to close this PR and make a new one.

PR #454, opened 2021-04-14
Add `BishopsGraph`, `RooksGraph`, and `QueensGraph`

finnbuck:
This pull request adds constructor functions for the [Bishop's](https://mathworld.wolfram.com/BishopGraph.html), [Rook's](https://en.wikipedia.org/wiki/Rook%27s_graph) and [Queen's](https://mathworld.wolfram.com/QueenGraph.html) graphs.


doc/examples.xml
```diff
@@ -529,6 +574,51 @@ true
...
+
+    From
+    <URL>https://mathworld.wolfram.com/RookGraph.html</URL>:
+    <P/>
+
+    <Q>The <C><A>m</A> x <A>m</A></C> rook's graph (confusingly called the 
+    <C><A>m</A> x <A>m</A></C> grid by Brouwer et al.1989, p. 440) and also
+    sometimes known as a lattice graph (e.g., Brouwer) is the graph Cartesian 
+    product <C>K_m square K_n</C> of complete graphs, which is
+    equivalent to the line graph <C>L(K_(m,n))</C> of the complete bipartite graph
```
marinaanagno, line 594: here you could maybe make a comment or reference about the line graph

gap/examples.gi
```diff
@@ -369,6 +369,134 @@ GeneralisedPetersenGraphCons);
...
+  od;
+  for i in [1 .. (m - 1)] do
+    for j in [1 .. (n - 1)] do
+      if IsEvenInt(i + j) and color = "light-square"
+          or IsOddInt(i + j) and color = "dark-square" then
+        v := (i - 1) * n + j + 1;
+        DigraphAddEdge(D2, [v, v + n - 1]);
+      fi;
+    od;
+  od;
```
marinaanagno, line 403:
```suggestion
   for i in [1 .. (m - 1)] do
     for j in [1 .. (n - 1)] do
       if IsEvenInt(i + j) and color = "dark-square"
           or IsOddInt(i + j) and color = "light-square" then
         v := (i - 1) * n + j;
         DigraphAddEdge(D1, [v, v + n + 1]);
       elif IsEvenInt(i + j) and color = "light-square"
           or IsOddInt(i + j) and color = "dark-square" then
         v := (i - 1) * n + j + 1;
         DigraphAddEdge(D2, [v, v + n - 1]);
       fi;
     od;
   od;
```
I feel that this would make it neater. It works on my computer when I tried it :)

doc/examples.xml
```diff
@@ -313,6 +313,51 @@ gap> GeneralisedPetersenGraph(IsMutableDigraph, 9, 4);
...
+    then this operation returns the <E>bishop's graph</E> for an <A>m</A> by
+    <A>n</A> board for a bishop of that square color.<P/>
+
+    From
+    <URL>https://mathworld.wolfram.com/BishopGraph.html</URL>:
+    <P/>
+
+    <Q> A bishop's graph is a graph formed from possible moves of a bishop chess
+    piece, which may make diagonal moves of any length on a chessboard (or any 
+    other board). To form the graph, each chessboard square is considered a vertex,
```
marinaanagno, line 332: I'm not sure I understand the comment "any other board". I suppose you mean not necessarily 8x8 board? Maybe we could make it a bit more precise and say that any other board would again have black and white squares like chessboards etc :)

marinaanagno, line 332:
```suggestion
    other board). To form the graph, each chessboard square is represented by a vertex, 
```

doc/examples.xml
```diff
@@ -313,6 +313,51 @@ gap> GeneralisedPetersenGraph(IsMutableDigraph, 9, 4);
...
+    <A>n</A> board for a bishop of that square color.<P/>
+
+    From
+    <URL>https://mathworld.wolfram.com/BishopGraph.html</URL>:
+    <P/>
+
+    <Q> A bishop's graph is a graph formed from possible moves of a bishop chess
+    piece, which may make diagonal moves of any length on a chessboard (or any 
+    other board). To form the graph, each chessboard square is considered a vertex, 
+    and vertices connected by allowable bishop moves are considered edges. Because
```
marinaanagno, line 333:
```suggestion
    and there is an edge (or path?) between two vertices if and only if one can be reached by the other by an allowable move of a bishop. Because 
```

doc/examples.xml
```diff
@@ -529,6 +574,51 @@ true
...
+    <P/>
+
+    <Q>The <C><A>m</A> x <A>m</A></C> rook's graph (confusingly called the 
+    <C><A>m</A> x <A>m</A></C> grid by Brouwer et al.1989, p. 440) and also
+    sometimes known as a lattice graph (e.g., Brouwer) is the graph Cartesian 
+    product <C>K_m square K_n</C> of complete graphs, which is
+    equivalent to the line graph <C>L(K_(m,n))</C> of the complete bipartite graph
+    <C>K_(m,n)</C>. This is the definition adopted for example by Brualdi and Ryser
+    (1991, p. 153), although restricted to the case <C>m = n</C>. This definition
+    corresponds to the connectivity graph of a rook chess piece (which can
```
marinaanagno, line 597: again I am not sure of what a connectivity graph is, I am aware of the term "graph connectivity" but not really of a connectivity graph :)

marinaanagno (dismissed):

gap/examples.gi
finnbuck: Yes! That's a good point, the way I wrote it was really redundant there. Thanks for pointing that out.

doc/examples.xml
finnbuck: I think I am going to have to re-write these sections in any case, since we might not be allowed to directly quote wolfram alpha!

tomcontileslie (changes requested):
Very nice pull request with some really useful examples. There is very little redundant work thanks to clever use of other line piece graphs, along with graph products, to give the more complicated graphs. Some thought is necessary to decide what's best semantics-wise for the bishop's graph, given that in its current form, output is very convenient but doesn't quite line up with the quoted definition from Wolfram MathWorld. Whether you decide to change the output format, or instead just change the definition you give in the documentation to make it consistent with the implementation, this will be a great feature.

doc/examples.xml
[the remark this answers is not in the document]
finnbuck: Yes, looking over the definitions again I also think some changes need to be made. The second solution seems like a reasonable fix, since it doesn't require making additional functions which do almost identical things and still lines up fairly well with the definition.

doc/examples.xml
```diff
@@ -529,6 +574,51 @@ true
...
+    by <A>n</A> board.<P/>
+
+    From
+    <URL>https://mathworld.wolfram.com/RookGraph.html</URL>:
+    <P/>
+
+    <Q>The <C><A>m</A> x <A>m</A></C> rook's graph (confusingly called the 
+    <C><A>m</A> x <A>m</A></C> grid by Brouwer et al.1989, p. 440) and also
+    sometimes known as a lattice graph (e.g., Brouwer) is the graph Cartesian 
+    product <C>K_m square K_n</C> of complete graphs, which is
```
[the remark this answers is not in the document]
finnbuck: I guess it would probably make more sense to prioritize the aesthetics of the PDF over the aesthetics of the plain text manual, so that sounds good to me!

doc/examples.xml
marinaanagno: sounds good! :)

wilfwilson (commented):
I started to make many small documentation-related changes, but then I realised that it would be quicker and easier for me to just make the changes and push. So I'm going to do just that

gap/examples.gi
```diff
@@ -369,6 +369,183 @@ GeneralisedPetersenGraphCons);
...
+
+InstallMethod(BishopsGraphCons,
+"for IsImmutableDigraph, a string and two positive integers",
+[IsImmutableDigraph, IsString, IsPosInt, IsPosInt],
+function(filt, color, m, n)
+  local D;
+  D := MakeImmutable(BishopsGraphCons(IsMutableDigraph, color, m, n));
+  SetIsMultiDigraph(D, false);
+  SetIsSymmetricDigraph(D, true);
+  SetIsConnectedDigraph(D, true);
```
wilfwilson, line 424: This is not necessarily true:
```
gap> D := BishopsGraph("dark", 1, 6);
<immutable empty digraph with 3 vertices>
gap> IsConnectedDigraph(D);
true
gap> DigraphNrConnectedComponents(D);
3
```

wilfwilson (changes requested):
Thanks for adding these. I haven't looked at the methods closely, but there is one problem that I found (commented on below).

I also think it would be nice (although I don't insist on it) if the vertices were labelled with the co-ordinates of their square. Ideally "A1" etc - but we can't do that since there are only 26 letters but we want to make arbitrarily large graphs 🙂 If it would be easy to add, then, could you please add vertex labels like we have for `KingsGraph`:
```
gap> DigraphVertexLabels(KingsGraph(4, 7));
[ [ 1, 1 ], [ 2, 1 ], [ 3, 1 ], [ 4, 1 ], [ 1, 2 ], [ 2, 2 ], [ 3, 2 ],
  [ 4, 2 ], [ 1, 3 ], [ 2, 3 ], [ 3, 3 ], [ 4, 3 ], [ 1, 4 ], [ 2, 4 ],
  [ 3, 4 ], [ 4, 4 ], [ 1, 5 ], [ 2, 5 ], [ 3, 5 ], [ 4, 5 ], [ 1, 6 ],
  [ 2, 6 ], [ 3, 6 ], [ 4, 6 ], [ 1, 7 ], [ 2, 7 ], [ 3, 7 ], [ 4, 7 ] ]
```

gap/examples.gi
finnbuck: Right, thanks for pointing that out! I'd gotten a bit confused with the definition of a connected graph.

finnbuck:
Hi, thanks for reviewing this pull request @wilfwilson! I just wanted to quickly respond to this comment you sent a few days ago:

> I also think it would be nice (although I don't insist on it) if the vertices were labelled with the co-ordinates of their square. Ideally "A1" etc - but we can't do that since there are only 26 letters but we want to make arbitrarily large graphs. If it would be easy to add, then, could you please add vertex labels like we have for `KingsGraph`:
> 
> ```
> gap> DigraphVertexLabels(KingsGraph(4, 7));
> [ [ 1, 1 ], [ 2, 1 ], [ 3, 1 ], [ 4, 1 ], [ 1, 2 ], [ 2, 2 ], [ 3, 2 ],
>   [ 4, 2 ], [ 1, 3 ], [ 2, 3 ], [ 3, 3 ], [ 4, 3 ], [ 1, 4 ], [ 2, 4 ],
>   [ 3, 4 ], [ 4, 4 ], [ 1, 5 ], [ 2, 5 ], [ 3, 5 ], [ 4, 5 ], [ 1, 6 ],
>   [ 2, 6 ], [ 3, 6 ], [ 4, 6 ], [ 1, 7 ], [ 2, 7 ], [ 3, 7 ], [ 4, 7 ] ]
> ```
> 
> so that the vertex with label `[1,1]` would correspond to the square `A1`, and `[2,4]` to `B4`, etc.

Basically, I was a bit surprised at first that the king's graph had vertex labels, since I do not remember programming these into that function and I could not find any later edits that added them!

After looking at it for a little bit I think the vertex labeling is coming from the use of the Cartesian product function, since the queen's and rook's graphs also have it while the bishop's graph does not (the bishop's graph function is also the only one that does not rely on the Cartesian product of two graphs).

So I've decided for consistency's sake I will try and add the vertex labeling to the bishop's graph as well.

wilfwilson:
Thanks for the reply @finnbuck, I'm sorry I broke your tests but I think I've just fixed them now (I hope). The 'problem' was that I switched the meaning of the arguments `m` and `n` for the Rooks graph, but not for the Bishops graph, and so the Queens graph (which is the combination of the two) ended up being confused. I've put `RooksGraph` back to how it was.

The reason I made the change is that I wanted the `m x n` Rooks graph (whose vertices are already labelled, as you say) to have the vertex with label `[m, n]`. i.e I wanted the 3x2 graph to have a vertex labelled `[3,2]` (corresponding to square C2 on a chessboard). I felt that this was more natural. But currently the 3x2 graph has the square labelled `[2,3]` instead.

The reason that `KingsGraph` has labelled vertices is because it uses `TriangularGridGraph`, and that produces a digraph with the vertices already labelled appropriately. So it wasn't something that you had to add explicitly.

wilfwilson:
So please feel free to give it a go, labelling the vertices of the digraphs. I'm happy to sort it out if it proves too tricky.

If we're going to do it, I really think it should be done consistently with the existing named pieces, so that the `m x n` piece-graph has its labels in the range from `[1,1]` to `[m,n]`, and everything in between, like:
```
gap> DigraphVertexLabels(KingsGraph(2, 3));
[ [ 1, 1 ], [ 2, 1 ], [ 1, 2 ], [ 2, 2 ], [ 1, 3 ], [ 2, 3 ] ]
```
(in particular note it has the label `[2,3]` but not `[3,2]`).

Once the vertex labels are done I'll be happy for this to be merged! Thanks again.

finnbuck:
Apologies for the wait! I think I've just about figured out how to do the vertex labels by now and will hopefully be pushing those changes later this evening...

wilfwilson:
I'm taking a look at this again, thanks @finnbuck

wilfwilson (approved):
Thanks for your patience @finnbuck.

I've updated the pull request so that all five of the of the chess piece graph functions are 'consistent', in the way that they use labels that they use for vertices.

i.e. vertex 1 is always `[1, 1]`,  vertex `2` is `[2, 1]`, vertex `3` is `[3, 1]`, ..., vertex `m` is `[m, 1]`, vertex `m+1` is `[1, 2]`, and so on.

This way, the `QueensGraph(m, n)` contains `KingsGraph(m, n)`, `RooksGraph(m, n)`, and `BishopsGraph(m, n)` as subdigraphs, and the vertex labels are the same in all cases.

I also reworked the documentation a bit.

Let's see if the tests pass!

finnbuck:
Thank you Wilf! Looks like all the checks have indeed passed. I've had a read through the changes and it all looks quite sensible to me. I'll try to keep in mind some of these tricks to make my code more readable and maintainable in the future!

PR #465, opened 2021-05-10
Allow `AsDigraph` to take permutations

wilfwilson:
This is something else that I've found handy for my research. Since I've written the code, I may as well share it, in case it proves useful to others.

### DIFF
--- a/tst/standard/digraph.tst
+++ b/tst/standard/digraph.tst
@@ -447,7 +447,7 @@ gap> gr2 := DigraphByInNeighboursConsNC(IsImmutableDigraph, inn);
 gap> gr2 := DigraphByInNeighbours(IsImmutableDigraph, inn);
 <immutable multidigraph with 20 vertices, 8 edges>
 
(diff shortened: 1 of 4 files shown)


tst/standard/digraph.tst
```diff
@@ -447,7 +447,7 @@ gap> gr2 := DigraphByInNeighboursConsNC(IsImmutableDigraph, inn);
 gap> gr2 := DigraphByInNeighbours(IsImmutableDigraph, inn);
 <immutable multidigraph with 20 vertices, 8 edges>
 
-#  AsDigraph
+#  AsDigraph for a permutation
```
james-d-mitchell, line 450: Permutation -> transformation?

james-d-mitchell (commented):
I’m happy with this mod that one small comment!

tst/standard/digraph.tst
wilfwilson: Well spotted!

Issue #467, opened 2021-05-12
Digraph combining functions modify a list digraph given as an argument


wilfwilson:
According to the documentation, you can either call `DigraphEdgeUnion` with a list of digraphs, or with one digraph per argument, i.e. `DigraphEdgeUnion(D1, D2, ...)` or `DigraphEdgeUnion([D1, D2, ...])`

In the second version, the list is changed in-place. This should not happen.
```
gap> D1 := CycleDigraph(3);; D2 := DigraphReverse(D1);; L := [D1, D2];
[ <immutable cycle digraph with 3 vertices>,
  <immutable digraph with 3 vertices, 3 edges> ]
gap> DigraphEdgeUnion(L) = CompleteDigraph(3);  # correct answer
<immutable digraph with 3 vertices, 6 edges>
gap> D1;  # unchanged
<immutable cycle digraph with 3 vertices>
gap> D2;  # unchanged
<immutable digraph with 3 vertices, 3 edges>
gap> L;  # wrongly changed
[ [ [ 2, 3 ], [ 3, 1 ], [ 1, 2 ] ], [ [ 3 ], [ 1 ], [ 2 ] ] ]
```
This has the consequence that if the list you give is immutable, then an error is given:
```
gap> D1 := CycleDigraph(3);; D2 := DigraphReverse(D1);; L := Immutable([D1, D2]);;
gap> IsMutable(L);
false
gap> CallFuncList(DigraphEdgeUnion, L) = CompleteDigraph(3);  # still correct in the 'multiple arguments' form
true
gap> DigraphEdgeUnion(L);  # bad
Error, List Assignment: <list> must be a mutable list in
  OutNeighboursMutableCopy( arg[1]
 ) at /Users/Wilf/gap/pkg/digraphs/gap/oper.gi:403 called from
DIGRAPHS_CombinationOperProcessArgs( arg
 ); at /Users/Wilf/gap/pkg/digraphs/gap/oper.gi:502 called from
<function "DigraphEdgeUnion">( <arguments> )
 called from read-eval loop at *stdin*:72
type 'quit;' to quit to outer loop
```

wilfwilson:
In addition to `DigraphEdgeUnion`, also `DigraphJoin`, `DigraphCartesianProduct`, `DigraphDisjointUnion`, `DigraphCartesianProduct`, and `DigraphDirectProduct` have this problem.

This is because `DIGRAPHS_CombinationOperProcessArgs` changes its argument in place.

[...]

PR #468, opened 2021-05-12
Stop the 'combination operations' from changing their given list of digraphs

wilfwilson:
By 'combination operations', I mean `DigraphEdgeUnion`, `DigraphJoin`, `DigraphDisjointUnion`, `DigraphDirectProduct`, and `DigraphCartesianProduct`. See #467.

Given a list of digraphs, these operations would try to change this list. Therefore, they would either wrongly mutate this list, if it was mutable, else they would give an error when this failed.

Note: this concerns mutability of lists, not digraphs.

To fix this, I made `DIGRAPHS_CombinationOperProcessArgs` copy its argument, and return that copy when it is finished.

Resolves issue #467.

### DIFF
--- a/gap/oper.gi
+++ b/gap/oper.gi
@@ -386,7 +386,7 @@ InstallMethod(DigraphClosure,
 InstallGlobalFunction(DIGRAPHS_CombinationOperProcessArgs,
 function(arg)
   local copy, i;
-  arg := arg[1];
+  arg := ShallowCopy(arg[1]);
   if IsMutableDigraph(arg[1]) then
     for i in [2 .. Length(arg)] do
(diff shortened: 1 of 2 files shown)


wilfwilson:
FFS, the AppVeyor check _still_ appears here, even though I've revoked its access in Settings, and even though I've removed the `.appveyor.yml` file. Really annoying.

@james-d-mitchell could you please try logging into AppVeyor and deleting the Digraphs project, specifically the one located at https://ci.appveyor.com/project/james-d-mitchell/digraphs-0gul0

wilfwilson:
Okay, seems I just needed to delete a 'web hook'! AppVeyor all gone now, phew.

PR #484, opened 2021-05-24
Reinstate some `Splash` tests in `standard/display.tst`

wilfwilson:
This PR makes it possible to un-comment some commented-out tests in `tst/standard/display.tst`, and to add some new ones for increased code coverage of `gap/display.gi`.

It's impossible or me to get 100% diff coverage, so that PR check will be failing, because I've changed the code that actually does the Splashing (which we can't/don't want to test in the test suites). I've checked that `Splash` still works for `dot` and `latex` files.

Therefore this PR will need to be merged with the failing `Codecov/patch` check.

### DIFF
--- a/.covignore
+++ b/.covignore
@@ -1,3 +1,2 @@
-display.tst
 utils.tst
 io.tst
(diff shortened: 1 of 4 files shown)


wilfwilson:
It's going to be impossible or me to get 100% diff coverage, because I've changed the code that actually does the Splashing (which we can't/don't want to test in the test suites). I've checked that `Splash` still works for `dot` and `latex` files.

wilfwilson:
The good news 😇 :
> `codecov/project — 97.40% (+0.47%)`

PR #489, opened 2021-06-01
oper: add IsDigraphPath

james-d-mitchell:
Adds `IsDigraphPath` missing ~~tests and doc so far~~ nothing.

### DIFF
--- a/doc/oper.xml
+++ b/doc/oper.xml
@@ -1382,7 +1382,9 @@ gap> VerticesReachableFrom(D, 5);
 
     The method for <C>DigraphPath</C> has worst case complexity of <M>O(m +
       n)</M> where <M>m</M> is the number of edges and <M>n</M> the number of
-    vertices in <A>digraph</A>.
+      vertices in <A>digraph</A>.<P/>
+
+      See also <Ref Oper="IsDigraphPath"/>.
 
 <Example><![CDATA[
 gap> D := Digraph([[2], [3], [2, 3]]);
@@ -2129,5 +2131,57 @@ gap> data;
 [ 4 ]
 ]]></Example>
 </Description>
+</ManSection>
+<#/GAPDoc>
+
+<#GAPDoc Label="IsDigraphPath">
+<ManSection>
+  <Oper Name="IsDigraphPath" Arg="D, v, a"/>
+  <Oper Name="IsDigraphPath" Arg="D, list" Label="for a digraph and list"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    This function returns <K>true</K> if the arguments <A>v</A> and <A>a </A>
+    describe a path in the digraph <A>D</A>.
+
+    A directed path (or directed cycle) of non-zero length <C>n-1</C>,
+    <M>(v_1, e_1, v_2, e_2, ..., e_{n-1}, v_n)</M>,
+    is represented by a pair of lists <C>[<A>v</A>, <A>a</A>]</C> as follows:
+
+    <List>
+      <Item>
+        <A>v</A> is the list <M>[v_1, v_2, ..., v_n]</M>.
+      </Item>
+      <Item>
+        <A>a</A> is the list of positive integers <M>[a_1, a_2, ..., a_{n-1}]</M>
+        where for each each <M>i &lt; n</M>, <M>a_i</M> is the position of
+        <M>v_{i+1}</M> in
+        <C>OutNeighboursOfVertex(</C><A>D</A><C>,</C><M>v_i</M><C>)</C>
+        corresponding to the edge <M>e_i</M>.  This can be useful if the
+        position of a vertex in a list of out-neighours is significant, for
+        example in orbit digraphs.
+      </Item>
+    </List>
+
+    If the arguments to <C>IsDigraphPath</C> are a digraph <A>D</A> and list
+    <A>list</A>, then this is equivalent to calling <C>IsDigraphPath(<A>D</A>,
+    <A>list</A>[1], <A>list</A>[2])</C>.
+
+    <Example><![CDATA[
+gap> D := Digraph(IsMutableDigraph, Combinations([1 .. 5]), IsSubset);
+<mutable digraph with 32 vertices, 243 edges>
+gap> DigraphReflexiveTransitiveReduction(D);
+<mutable digraph with 32 vertices, 80 edges>
+gap> MakeImmutable(D);
+<immutable digraph with 32 vertices, 80 edges>
+gap> IsDigraphPath(D, [32, 31, 33], [1, 1]);
+false
+gap> IsDigraphPath(D, [1], []);
+true
+gap> IsDigraphPath(D, [6, 9, 16, 17], [3, 3, 2]);
+true
+gap> IsDigraphPath(D, DigraphPath(D, 6, 1));
+true
+]]></Example>
+  </Description>
 </ManSection>
 <#/GAPDoc>
--- a/gap/oper.gi
+++ b/gap/oper.gi
@@ -1333,6 +1333,37 @@ function(D, u, v)
   return out;
 end);
 
+InstallMethod(IsDigraphPath, "for a digraph and list",
+[IsDigraph, IsList],
+function(D, digraph_path_output)
+  if Length(digraph_path_output) <> 2 then
+    DError(["the 2nd argument (a list) must have length 2, ",
+            "but found length {}"],
+           Length(digraph_path_output));
+  fi;
+  return IsDigraphPath(D, digraph_path_output[1], digraph_path_output[2]);
+end);
+
+InstallMethod(IsDigraphPath, "for a digraph, homog. list, and homog. list",
+[IsDigraph, IsHomogeneousList, IsHomogeneousList],
+function(D, nodes, edges)
+  local a, i;
+  if Length(nodes) <> Length(edges) + 1 then
+    DError(["the 2nd and 3rd arguments (lists) are incompatible, ",
+            "expected 3rd argument of length {}, got {}"],
+           Length(nodes) - 1,
+           Length(edges));
+  fi;
+  a := OutNeighbours(D);
+  for i in [1 .. Length(edges)] do
+    if nodes[i] > Length(a) or edges[i] > Length(a[nodes[i]])
+        or a[nodes[i]][edges[i]] <> nodes[i + 1] then
+      return false;
+    fi;
+  od;
+  return true;
+end);
+
 InstallMethod(DigraphShortestPath,
 "for a digraph by out-neighbours and two pos ints",
 [IsDigraphByOutNeighboursRep, IsPosInt, IsPosInt],
--- a/gap/utils.gi
+++ b/gap/utils.gi
@@ -592,3 +592,25 @@ function(blist)
   od;
   return nr + 1;   # to be in [1 .. 2 ^ n]
 end);
+
+InstallGlobalFunction(DError,
+function(arg)
+  local msg;
+  if not (IsString(arg[1]) or IsList(arg[1])) then
+    Error("expected a string or a list as the 1st argument");
+  elif IsList(arg[1]) and not ForAll(arg[1], IsString) then
+    Error("expected a list of strings as the 1st argument");
+  fi;
+  if IsList(arg[1]) then
+    arg[1] := Concatenation(arg[1]);
+  fi;
+  msg := CallFuncList(StringFormatted, arg);
+  RemoveCharacters(msg, "\\\n");
+  ErrorInner(
+      rec(context := ParentLVars(GetCurrentLVars()),
+          mayReturnVoid := false,
+          mayReturnObj := false,
+          lateMessage := "type 'quit;' to quit to outer loop",
+          printThisStatement := false),
+      [msg]);
+end);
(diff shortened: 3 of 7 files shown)


doc/oper.xml
```diff
@@ -2131,3 +2133,47 @@ gap> data;
 </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="IsDigraphPath">
+<ManSection>
+  <Oper Name="IsDigraphPath" Arg="D, v, a"/>
```
wilfwilson, line 2139: I’m happy for this version of arguments to exist but I think that `IsDigraphPath` should also (and in my mind this would be the main one) have a version that accepts the output of `DigraphPath` as input. ie if there was a path from 1 to 2 in D then `IsDigraphPath(D, DigraphPath(D, 1, 2));` would return true.

wilfwilson (commented):
I would like to review this when I’m back at work next week

doc/oper.xml
james-d-mitchell: Done, although I left the main one as being the one that accepts 3 args, apart from anything, to avoid creating a list where one isn't required. Happy to revisit if necessary!

gap/utils.gi
```diff
@@ -577,3 +577,27 @@ function(nr, n)
   od;
   return x;
 end);
+
+InstallGlobalFunction(DError,
```
wilfwilson, line 581: I'm happy for this to be introduced, although I have a reservation that I'll detail in a minute. But at the very least I think the introduction of this function should be mentioned in the commit message; preferably in a separate commit; and ideally in a separate PR (although no big deal on that front).

The potential pitfall is about what happens when the first variable that you introduce after the intended error message is itself a string:
```
gap> variable := "wilf";;
gap> DError("Hi {} you've done something wrong!", variable);
Error, out of bounds -- used 1 replacement fields without id when there are only 0 arguments at /Users/Wilf/gap/lib/string.gi:1353 called from
CallFuncList( PrintToFormatted,
 Concatenation( [ OutputTextString( str, false ), s ], data )
 ); at /Users/Wilf/gap/lib/string.gi:1395 called from
CallFuncList( StringFormatted, arg{[ i - 1 .. Length( arg ) ]}
 ) at /Users/Wilf/gap/pkg/digraphs/gap/utils.gi:592 called from
<function "DError">( <arguments> )
```
Note also:
```
gap> DError("Hi {} you've done something wrong!", variable, fail);
Error, Hi fail you've done something wrong!wilf called from
not in any function at *stdin*:2
```
`StringFormatted` itself doesn't suffer from this problem because it requires the main error message string to be the first argument.

I like your idea of introducing this function, and of allowing the error message to be broken into multiple smaller strings. But if we choose this option, then it means that we need to be careful. Here are some options I see.

1. Be careful never to use a string for the first replacement argument (sounds too risky and having a variable be a string is realistic use case in my opinion).
2. We introduce a dummy separator argument to mark the boundary between message and replacement (at least required in cases where the first variable _could_ be a string), e.g.
```
DError(msg1, msg2, msg3, DIGRAPHS_SEPARATOR, var1, var2)
```
3. We 'give up' and use `Concatenation` around the message arguments so that they all fit into the first argument (uglier but works):
```
DError(Concatenation(msg1, msg2, msg3), var1, var2)
```
4. We 'give up' and fit everything into the first argument, but we start using multiline strings (would require some additional processing; see this example where it doesn't look quite right; might also require changes to the linter):
```
gap> DError("""this is the start of a very long error message and
> I am going to break it over multiple
> ({}) lines because I am cool""", 3);
Error, this is the start of a very long error message andI am going to break it over \
multiple(3) lines because I am cool called from
not in any function at *stdin*:36
```
Any other ideas?

gap/oper.gi
```diff
@@ -1333,6 +1333,55 @@ function(D, u, v)
...
+        err := "the range []";
+      fi;
+      DError("the 3rd argument (a list) must consist of values in ",
+             err,
+             " but found {1} in position {2}",
+             edges[i],
+             i,
+             Length(a[nodes[i]]));
+    elif a[nodes[i]][edges[i]] <> nodes[i + 1] then
+      return false;
```
wilfwilson, line 1379: I disagree with the philosophy of your implementation (of course it's just my opinion so feel free to disagree).

In my opinion the error should only be given here for a 'type' error: so at the start of the method, you should check that the lists have compatible lengths and both consist of positive integers, and give an error if not.

Then, I think that if one of the vertices isn't in the digraph, or if the indexed out-neighbour doesn't actually exist, the result should be `false` rather than an error. At this point in the method, the input is at least a path in _some_ digraph (maybe a very big and weird one, but still), and the question is: is it a path in _this_ digraph: yes or no?

For me this fits in with other ways that GAP works, e.g.:
```
gap> T5 := FullTransformationMonoid(5);;
gap> P := PartitionMonoid(5);;
gap> IsSubsemigroup(T5, P);
false
```
(`false` rather than `Error...`).

A bit artificial, but a possible use case could be filtering a list of a paths of a big digraph `D1` by whether or not they also exist in a sub digraph `D2` of `D1`, something like `Filtered(paths, x -> IsDigraphPath(D2, x))` which probably wouldn't work in your current implementation.

```suggestion
    if nodes[i] > Length(a) or edges[i] > Length(a[nodes[i]]) or a[nodes[i]][edges[i]] <> nodes[i + 1] then
      return false;
```

doc/oper.xml
```diff
@@ -2131,3 +2133,55 @@ gap> data;
...
+  <Oper Name="IsDigraphPath" Arg="D, list" Label="for a digraph and list"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    This function returns <K>true</K> if the arguments <A>v</A> and <A>a </A>
+    describe a path in the digraph <A>D</A>.
+
+    A directed path (or directed cycle) of non-zero length <C>n-1</C>,
```
wilfwilson, line 2146: I don't like the duplication of this block of text just for reasons of maintainance.

I'd say that the practical definition of should be given here, as it is, and that the documentation for `DigraphPath` should point here. So something like:
```
This function returns <K>true</K> if the arguments <A>v</A> and <A>a</A> describe a path in the digraph <A>D</A>.

The mathematical definition of a directed path and a directed cycle is given
in Section <Ref Subsect="Definitions" Style="Text"/>.
<P/>

In practice, a directed path (or directed cycle) of non-zero length <C>n-1</C>...
```

wilfwilson (commented):
Thanks for this @james-d-mitchell and sorry for being so critical in my review about such a small feature.

I totally understand why you added this, because a couple of days before I saw this PR I actually made my own implementation (visible https://github.com/wilfwilson/Digraphs/commit/4d2ca6a4ad5d6b36405a728db0b50134e67b5a19) but I didn't get around to making my own PR.

In my implementation I also included a version that just let you specify a list of vertices and it would return `true` iff each subsequent vertex was an out-neighbour of the last. I think I would find this useful. But perhaps this would be better under a different name, and in any case I'm happy to add it later in a separate PR. What do you think?

gap/utils.gi
```diff
@@ -577,3 +577,27 @@ function(nr, n)
...
+  i := 1;
+  while i <= Length(arg) and IsString(arg[i]) do
+    Append(err, arg[i]);
+    i := i + 1;
+  od;
+  if i > 1 then
+    arg[i - 1] := err;
+    err := [CallFuncList(StringFormatted, arg{[i - 1 .. Length(arg)]})];
+    RemoveCharacters(err[1], "\n");
+    RemoveCharacters(err[1], "\\");
```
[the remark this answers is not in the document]
james-d-mitchell: Aha, nice one, I'd actually prefer not to have to do this, but for whatever reasons `StringFormatted` inserts line breaks (which is an undocumented, and, at least for me, unwanted, feature!).

gap/utils.gi
james-d-mitchell: This is a good point, I also wondered about strings as variables, and somehow convinced myself that it shouldn't be an issue. I think you've convinced me the other way around now :) 

I'm happy to introduce this in a separate PR too, no problem. 

About what to do to resolve the issues of strings as variables: 

1. I agree we shouldn't do this
2+3. We could do something between 2 and 3: `DError([msg1, msg2, msg3], var1, var2);` minimally different, and avoids using the too long `Concatenation`? Alternatively, we could consume msg1, msg2,.. until we hit a non-string or we hit `DIGRAPHS_SEPARATOR`, thereby allowing `DError` to be used without `DIGRAPHS_SEPARATOR` if none of the variables is a string, and use it if the first variable is a string. 
4. I'm not enamoured with this option mostly because the resulting strings will be badly formatted in the code.

gap/oper.gi
james-d-mitchell: This makes sense, I'm happy to make this change!

doc/oper.xml
james-d-mitchell: Also good

james-d-mitchell: I'll try to make the changes today and re-push.

gap/utils.gi
wilfwilson: I'm happy to not do 1 and 4 and I like both of your suggestions for something between 2 and 3. I think my preference is for the list version, which would permit these variations:
```
DError([msg1, msg2, msg3], var1, var2);
DError([msg1], var1, var2);
DError(msg1, var1, var2);
```
which adds minimal ugliness.

gap/utils.gi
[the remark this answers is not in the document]
wilfwilson: I'm pretty sure @ChrisJefferson added `StringFormatted`, maybe we should ask him to add a `StringNotFormatted` too 😉

doc/oper.xml
james-d-mitchell: I've only got back to this today, and I think I've made all the changes you requested @wilfwilson, and although I thought I understood your comment above (the one "In practice", that I'm replying to) but now I don't. What would follow "In practice" precisely?

wilfwilson: Hey @james-d-mitchell, happy to try to clarify what I meant, and the way that I think about things.

The text at `<Ref Subsect="Definitions" Style="Text"/>` in the introduction contains (amongst other definitions) the following sentences:
> A directed walk on a digraph is a sequence of alternating vertices and edges (v_1, e_1, v_2, e_2, ..., e_n-1, v_n) such that each edge e_i has source v_i and range v_i+1. A directed path is a directed walk where no vertex (and hence no edge) is repeated

For me, this is the 'mathematical definition'.

However, in a practical sense in GAP/Digraphs, there is no such object as a 'sequence'; a vertex is a positive integer, and an edge is a list of two positive integers. So even if you were happy to consider a sequence as being another name for a list, then the literal interpretation of the mathematical definition would be that an object representing a digraph path in the Digraphs package would be something like `[5, [5,3], 3, [3,2], 2, [2, 6], 6]`. But that's not how we've decided to represent paths in the Digraphs package, we do in differently (and more sensibly). So in that sense, we have a practical definition that is separate from the mathematical definition.

The practical definition is what is (currently) given in the manual entry for `DigraphPath`. i.e. it's a pair of lists, the first one giving the list of vertices that occur, the second one giving the positions of subsequent vertices in the out-neighbours of the previous.

In my opinion, we should only give this 'practical definition' once (since it's easier to maintain one copy of a fiddly definition in the documentation); but instead of it being in the manual entry for `DigraphPath`, as it is now, I think it should instead be in the manual entry for `IsDigraphPath`. Then the manual entry for `DigraphPath` should no longer contain the full definition, but it should instead have a reference to the manual entry for `IsDigraphPath` in its place.

To make it more concrete:
```
< Manual entry for IsDigraphPath >

This function returns <K>true</K> if the arguments <A>v</A> and <A>a</A>
describe a path in the digraph <A>D</A>.
<P/>

The mathematical definition of a directed path and a directed cycle is given
in Section <Ref Subsect="Definitions" Style="Text"/>.
<P/>

In practice, a directed path (or directed cycle) of non-zero length
<C>n-1</C> (v_1, e_1, v_2, e_2, ..., e_n-1, v_n),
is represented by a pair of lists [v,a] as follows:
* v is the list [v_1, v_2, ..., v_n].
* ... [continue with the rest of the text from the manual entry for `DigraphPath`]
```
and then the manual entry for `DigraphPath` should look something like:
```
Returns:  A pair of lists, or fail.

  If  there  exists  a non-trivial directed path (or a non-trivial cycle, in the case that u = v) from
  vertex  u  to  vertex v in the digraph digraph, then this operation returns such a directed path (or
  directed cycle). Otherwise, this operation returns fail.

  See Section 'Definitions' for the mathematical definition of a directed path and a directed cycle,
  and see IsDigraphPath for the way in which paths and cycles are represented in the Digraphs package.
```

wilfwilson: In short, what should follow "In practice," is the paragraph-bulletpoint-bulletpoint-paragraph that currently forms the second half of the manual entry for `DigraphPath`.

wilfwilson: Sorry @james-d-mitchell I didn't do a great job of this 😬. Having looked at the diff and remembering the context more, I think I can say it better:

I am suggesting that you add:
```
<P/>

The mathematical definition of a directed path and a directed cycle
is given in Section <Ref Subsect="Definitions" Style="Text"/>.
<P/>

In practice,
```
between the first and second paragraphs that you've already included in this PR.

And also I'm suggesting that you shorten the documentation for `DigraphPath`, by pointing to `IsDigraphPath` rather than repeat the same information.

wilfwilson (approved):
This is good, thank you for making those changes. The final documentation thing is minor, and I'll make that change separately if I feel like it. Or just leave it, because it's already good as-is, cheers.

PR #494, opened 2021-06-09
Add more implications (including some implications of falsity)

wilfwilson:
I learned recently (see the discussion starting https://github.com/gap-system/gap/pull/4546#discussion_r646906841) that you can use
```
InstallTrueMethod( HasIsX, IsY );
```
to, in practice, automatically set `IsX(D) = false` when `IsY(D) = true` becomes known. The only pitfall is that if `IsX(D)` is already (i.e. wrongly) set to `true`, then this true method doesn't do anything; in particular it doesn't complain. So this doesn't help us to catch bugs (where properties get set incorrectly), but we weren't catching them anyway, and it does offer useful functionality for storing learned information.

Therefore I have added some such '`InstallFalseMethod`'s to Digraphs. This was made easier with the addition of `DigraphHasVertices` and `DigraphHasEdges`, which allows me to easily exclude some trivial exceptions which would otherwise stop certain implications being added.

I also added immediate methods for these things so that they are set to true or false once the attributes `DigraphNrVertices` and `DigraphNrEdges` are known; similarly, I added immediate methods for `DigraphNrVertices` and `DigraphNrEdges` once the attributes `DigraphVertices` and `DigraphEdges` are known.

I still need to add documentation and tests, and possibly I'll come up with some more `InstallTrueMethod`s to add.

### DIFF
--- a/gap/attr.gi
+++ b/gap/attr.gi
@@ -1080,7 +1080,7 @@ function(D)
   # alter the answer for the diameter/girth if necessary.  This function is
   # called, if appropriate, by DigraphDiameter and DigraphUndirectedGirth.
 
-  if DigraphNrVertices(D) = 0 and IsImmutableDigraph(D) then
+  if DigraphHasNoVertices(D) and IsImmutableDigraph(D) then
     SetDigraphDiameter(D, fail);
     SetDigraphUndirectedGirth(D, infinity);
     return rec(diameter := fail, girth := infinity);
@@ -1491,10 +1491,10 @@ end);
 
 InstallMethod(DegreeMatrix, "for a digraph", [IsDigraph],
 function(D)
-  if DigraphNrVertices(D) = 0 then
-    return [];
+  if DigraphHasAVertex(D) then
+    return DiagonalMat(OutDegrees(D));
   fi;
-  return DiagonalMat(OutDegrees(D));
+  return [];
 end);
 
 InstallMethod(LaplacianMatrix, "for a digraph", [IsDigraph],
@@ -1815,7 +1815,7 @@ function(D)
     SetDigraphAddAllLoopsAttr(D, C);
     SetIsReflexiveDigraph(C, true);
     SetIsMultiDigraph(C, ismulti);
-    SetDigraphHasLoops(C, DigraphNrVertices(C) > 0);
+    SetDigraphHasLoops(C, DigraphHasAVertex(C));
   fi;
   return C;
 end);
@@ -2264,7 +2264,7 @@ InstallMethod(UndirectedSpanningForest, "for a digraph by out-neighbours",
 [IsDigraphByOutNeighboursRep],
 function(D)
   local C;
-  if DigraphNrVertices(D) = 0 then
+  if DigraphHasNoVertices(D) then
     return fail;
   fi;
   C := MaximalSymmetricSubdigraph(D)!.OutNeighbours;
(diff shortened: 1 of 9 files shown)


gap/attr.gi
```diff
@@ -530,14 +530,7 @@ function(D)
 end);
 
 InstallMethod(DigraphNrEdges, "for a digraph", [IsDigraphByOutNeighboursRep],
-function(D)
-  local m;
-  m := DIGRAPH_NREDGES(D);
-  if IsImmutableDigraph(D) then
-    SetIsEmptyDigraph(D, m = 0);
```
wilfwilson, line 537: TODO: nonempty digraphs now automatically know that they are nonempty once they know they number of edges, because of a true method, but empty digraphs do not know that they are empty, even if their number of edges is zero. So perhaps I want to add something like this back in (or find another way of doing it).

gap/prop.gi
```diff
@@ -162,10 +189,6 @@ function(D)
   return DigraphNrConnectedComponents(D) = 1;
 end);
 
-InstallImmediateMethod(IsAcyclicDigraph, "for a reflexive digraph",
-IsReflexiveDigraph, 0,
-D -> DigraphNrVertices(D) = 0);
```
wilfwilson, line 167: TODO: Check that by deleting this, we aren't losing anything important.

gap/prop.gi
```diff
@@ -307,10 +330,6 @@ end);
 InstallMethod(IsReflexiveDigraph, "for a digraph", [IsDigraph],
 D -> ForAll(DigraphVertices(D), x -> IsDigraphEdge(D, x, x)));
 
-InstallImmediateMethod(DigraphHasLoops, "for a reflexive digraph",
```
wilfwilson, line 310: TODO: Check that by deleting this, we aren't losing anything important.

gap/attr.gi
wilfwilson: Probably I just want there to be an `IsEmptyDigraph` immediate method for `HasDigraphNrEdges`, doing the same (well, opposite) as the new immediate method for `DigraphHasEdges`.

james-d-mitchell:
This is interesting @wilfwilson, I generally like the idea, and see this as a major improvement, but I'm vaguely worried that this feature relies on a (possibly) undocumented implementational detail of the type system in GAP (that if the default value of the bit corresponding to a `IsSomething` is `0`, and if we set `HasIsSomething` to `true` but don't set the value of `IsSomething` it will be `false`).  If this isn't likely to change in GAP in the future, then I'm only vaguely worried, and will forget soon 👍

wilfwilson:
Fair concern! I can't really imagine this is the sort of thing that would ever change about GAP.

But even if it did, this trick is used in GAP itself, and now in at least one other package (polycyclic). So any change to GAP would have to be (and would be) either backwards compatible, or postponed until all packages using this trick were changed to use some other system.

gap/grahom.gi
```diff
@@ -109,10 +109,10 @@ function(D, n)
 
   # Only the null D with 0 vertices can be coloured with 0 colours
```
wilfwilson, line 110:
```suggestion
  # Only the null digraph with 0 vertices can be coloured with 0 colours
```

wilfwilson:
I've distilled this into its essence I think, maybe one day I'll make future improvements in a separate PR.

As mentioned above, I do not think that this is the kind that of thing that will be changed in future versions of GAP. But I've added extensive tests for the implications of falsity, so we would immediately catch it if any of them ever stops working. There's no risks that this will give us wrong answers; I suppose the risk is just that the true methods won't do anything any more. If that ever becomes the case we can just remove them.

PR #495, opened 2021-06-10
Use macros to reduce amounts of repeated text in the manual

### LINKED ISSUE #478: Replace the oft-repeated text `If the optional first argument <A>filt</A> is present...` in the manual with an include
The following text is included in many manual entries, which makes maintenance annoying:
```
    If the optional first argument <A>filt</A> is present, then this should
    specify the category or representation the digraph being created will
    belong to. For example, if <A>filt</A> is <Ref Filt="IsMutableDigraph"/>,
    then the digraph being created will be mutable, if <A>filt</A> is <Ref
    Filt="IsImmutableDigraph"/>, then the digraph will be immutable. 
    If the optional first argument <A>filt</A> is not present, then <Ref
      Filt="IsImmutableDigraph"/> is used by default.<P/>
```
It's possibly that there are already subtle differences between these!

Perhaps these could all be replaced by an 'entity' in `doc/main.xml` (e.g. `MUTABILITYSPIEL`) that can include via `&MUTABILITYSPIEL;` as appropriate:
https://github.com/digraphs/Digraphs/blob/46f5163e97985fb0553dd2349859d2796d6769b5/doc/main.xml#L2-L27

wilfwilson:
Specifically:
* &STANDARD_FILT_TEXT;
* &MUTABLE_RECOMPUTED_PROP;
* &MUTABLE_RECOMPUTED_ATTR;

Also fixes some typos in `DigraphByInNeighbours` and adjusts some whitespace (😬).

These things are done in separate commits.

Resolves #478.

### DIFF
--- a/doc/digraph.xml
+++ b/doc/digraph.xml
@@ -368,11 +368,12 @@ gap> DigraphByEdges(IsMutableDigraph,
   <C>[1 .. Length(<A>list</A>)]</C>, then this function returns the
   digraph with vertices <M>E^0=</M><C>[1 .. Length(<A>list</A>)]</C>, and
   edges corresponding to the entries of <A>list</A>. More precisely, there is an
-  edge with source vertex <C>i</C> and range vertex <C>j</C> if <C>i</C> is list
-  <C><A>list</A>[j]</C>. <P/>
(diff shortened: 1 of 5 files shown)


james-d-mitchell (approved):
Very good!

PR #499, opened 2021-09-15
Dfs

james-d-mitchell:
This PR fixes some bugs introduced in #459

Issue #487, opened 2021-05-28
Wrong result from `DigraphPath` in the master branch


wilfwilson:
This example came from the failing Semigroups package tests https://github.com/semigroups/Semigroups/pull/746/checks?check_run_id=2688637629 although there are likely to be smaller examples.
```
gap> D := Digraph([
>   [ 2, 3, 4, 5, 5 ], [ 6, 3, 4, 7, 5 ], [ 8, 9, 10, 8, 11 ],
>   [ 12, 13, 14, 15, 16 ], [ 2, 13, 4, 12, 17 ], [ 6, 9, 4, 16, 11 ],
>   [ 18, 13, 4, 12, 8 ], [ 8, 19, 10, 19, 20 ], [ 8, 9, 10, 8, 21 ],
>   [ 12, 13, 14, 15, 16 ], [ 22, 13, 14, 12, 16 ], [ 23, 13, 24, 12, 8 ],
>   [ 19, 9, 19, 8, 24 ], [ 19, 13, 19, 15, 16 ], [ 21, 19, 24, 19, 20 ],
>   [ 25, 13, 10, 12, 8 ], [ 26, 13, 10, 12, 17 ], [ 6, 3, 4, 7, 27 ],
>   [ 19, 19, 19, 19, 19 ], [ 28, 13, 19, 12, 16 ], [ 29, 13, 14, 12, 16 ],
>   [ 23, 3, 24, 7, 30 ], [ 29, 9, 14, 16, 24 ], [ 12, 19, 14, 19, 19 ],
>   [ 8, 8, 10, 24, 15 ], [ 8, 8, 10, 24, 31 ], [ 30, 19, 4, 19, 20 ],
>   [ 19, 8, 19, 24, 12 ], [ 23, 9, 24, 16, 21 ], [ 6, 13, 4, 12, 17 ],
>   [ 32, 13, 24, 12, 17 ], [ 29, 3, 14, 7, 7 ] ]);;
gap> path := DigraphPath(D, 5, 5);;
gap> # Check the value against the definition - wrong
gap> ForAll([1 .. Length(path[2])], i -> OutNeighbours(D)[path[1][i]][path[2][i]] = path[1][i + 1]);
false
gap> # Check whether the vertices are even correct (with just the edges being wrong) - no
gap> ForAll([1 .. Length(path[2])], i -> path[1][i + 1] in OutNeighbours(D)[path[1][i]]);
false
gap> # This is the first spot where it goes wrong
[...]

PR #459, opened 2021-04-29
ExecuteDFS (generic depth first search)

LRacine:
In this pull request, I submit my work for VP2215.
I did this to resolve the issue noted in https://github.com/digraphs/Digraphs/issues/393

From this pull request you will see that I have created:
- A generic depth first search algorithm _ExecuteDFS_ that takes a record, data object, start vertex,
preorder function, postorder function, ancestor function and a cross function.
- _NewDFSRecord_ which creates a record for the  _ExecuteDFS_ function that stores the preorder, postorders, parent,
current and child attributes
- _DFSDefault_ which acts as a default argument for the _ExecuteDFS_ function

The generic depth first search algorithm works as follows:

```
While there are nodes in our stack
  pop node from stack
   If we are backtracking on this node
     PostOrderFunc()
     Set and save the post order number for the node
   elif we’ve visited it but we are not backtracking
     Continue;
   else
     Set and save the pre order number for the node
     Add it back to the stack for backtracking
 
   Return if the user want to

   Go through the node’s neighbours
     if the neighbour is unvisited
       Set its parent as the node
       Add it to the stack
     elif we haven’t backtracked on the neighbour
        AncestorFunc()
     elif we visited the neighbour before the node
       CrossFunc()
     Return if the user want to
```

I also refactored the existing DFS implementations in:
- UndirectedSpanningForest
- DigraphPath
- DigraphTopologicalSort
- DigraphLongestDistanceFromVertex
- IsAntisymmetricDigraph
- IsAcyclicDigraph
- DIGRAPHS_ArticulationPointsBridgesStrongOrientation
- VerticesReachableFrom
- DominatorTree

PR #505, opened 2021-10-27
Re-add PR #459: ExecuteDFS (generic depth first search)

wilfwilson:
This reverts #504, in other words, it adds again the code that was in #459.

Since the code in this PR is buggy (see #487), it was removed from the `main` branch. I do not imagine that this PR _itself_ will ever be merged (at least not as-is), but I have pushed this branch and opened this PR so that we do not lose track of this code.

@james-d-mitchell has started PR #499 to try to fix this, which I have taken the liberty of rebasing on top of this branch. When #499 is merged, this PR should then be closed, and the associated branch deleted.


wilfwilson:
Ah good to know, so that will supersede whatever is in here? Great!

PR #512, opened 2021-11-03
Implement `DigraphContractEdge`

hrj4:


gap/oper.gi
```diff
@@ -309,6 +309,36 @@ function(D, u, v)
...
+    ErrorNoReturn("the 1st argument <D> must be a digraph with no multiple ",
+                  "edges,");
+  elif not (IsPosInt(src) and src in DigraphVertices(D)) then
+    ErrorNoReturn("the 2nd argument <src> must be a vertex of the ",
+                  "digraph <D> that is the 1st argument,");
+  elif not (IsPosInt(ran) and ran in DigraphVertices(D)) then
+    ErrorNoReturn("the 3rd argument <ran> must be a vertex of the ",
+                  "digraph <D> that is the 1st argument,");
+  elif not ([src,ran] in DigraphEdges(D)) then
+    ErrorNoReturn("there must be an edge between ",src," and ",ran);
```
wilfwilson, line 327:
```suggestion
  elif not ([src, ran] in DigraphEdges(D)) then
    ErrorNoReturn("there must be an edge between ", src, " and ", ran);
```

gap/oper.gi
```diff
@@ -309,6 +309,36 @@ function(D, u, v)
...
 
+InstallMethod(DigraphContractEdge,
+"for a mutable digraph by out-neighbours and two positive integers",
+[IsMutableDigraph and IsDigraphByOutNeighboursRep, IsPosInt, IsPosInt],
+function(D, src, ran)
+  local V, P;
+  if IsMultiDigraph(D) then
+    ErrorNoReturn("the 1st argument <D> must be a digraph with no multiple ",
+                  "edges,");
+  elif not (IsPosInt(src) and src in DigraphVertices(D)) then
```
wilfwilson, line 320: Given the filter of this method, `src` and `ran` are guaranteed to be positive integers, so I will remove this check. In any case, if `src` ever wasn't a positive integer, then it wouldn't be a member of `DigraphVertices(D)`, and so there still wouldn't be a need to check whether it was a positive integer first.
```suggestion
  elif not src in DigraphVertices(D) then
```

gap/oper.gi
```diff
@@ -309,6 +309,36 @@ function(D, u, v)
...
+    ErrorNoReturn("the 2nd argument <src> must be a vertex of the ",
+                  "digraph <D> that is the 1st argument,");
+  elif not (IsPosInt(ran) and ran in DigraphVertices(D)) then
```
wilfwilson, line 323:
```suggestion
  elif not ran in DigraphVertices(D) then
```

gap/oper.gi
```diff
@@ -309,6 +309,36 @@ function(D, u, v)
...
+    ErrorNoReturn("the 3rd argument <ran> must be a vertex of the ",
+                  "digraph <D> that is the 1st argument,");
+  elif not ([src, ran] in DigraphEdges(D)) then
```
wilfwilson, line 326: For a single edge, the following should be much more efficient (it typically avoids having to construct the whole set of edges from the list of out-neighbours).
```suggestion
  elif not IsDigraphEdge(D, src, ran) then
```

doc/oper.xml
```diff
@@ -110,6 +110,49 @@ true]]></Example>
...
+    If <A>digraph</A> belongs to <Ref
+      Filt="IsMutableDigraph"/>, then the edge is contracted directly on
+    <A>digraph</A>. If <A>digraph</A> belongs to <Ref
+      Filt="IsImmutableDigraph"/>, an immutable copy of <A>digraph</A> with
+    the edge contracted is returned.
+    <P/>
+
+    Note that if <A>digraph</A> belongs to <Ref Filt="IsImmutableDigraph"/>,
+    then a new copy of <A>digraph</A> will be returned even if <A>[src, ran]</A>
+    does not define an edge of <A>digraph</A>.<P/>
```
wilfwilson, line 133: This does not match the implementation: currently an error occurs in this situation. Which behaviour do you want?

gap/oper.gi
```diff
@@ -309,6 +309,36 @@ function(D, u, v)
...
+    ErrorNoReturn("there must be an edge between ", src, " and ", ran);
```
wilfwilson, line 327: I think this should be more precise (currently is doesn't explicitly say which direction the edge should be)
```suggestion
    ErrorNoReturn("[<src>, <ran>] must be an edge of the digraph <D> ",
                  "(the 2nd, 3rd, and 1st arguments, respectively),");
```

wilfwilson (changes requested):

wilfwilson (commented):
Sorry, I forgot to say: thank you for this pull request, it'll be a nice feature to have! I just have a few suggestions/questions.

hrj4:
Hi Wilf, thanks for reviewing the pull request and for flagging the parts needing improvement. Hopefully I've resolved the things which needed resolving now... I'm one of the VIP students this semester and this was my first attempt at implementing a new function so I appreciate the detailed explanations of what I could've done better.

james-d-mitchell:
This feature was included in #618, so I'm going to close this.

PR #513, opened 2021-11-03
Added IsPermutationDigraph

baydrea:


doc/prop.xml
```diff
@@ -810,6 +810,34 @@ true
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="IsPermutationDigraph">
+<ManSection>
+  <Prop Name="IsPermutationDigraph" Arg="digraph"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    This property is <K>true</K> if the digraph <A>digraph</A> is
+    functional and each node has only one in neighbour. <P/>
```
wilfwilson, line 819: We use the term "in-neighbour" (with hyphen) in the Digraphs package 🙂 
```suggestion
    functional and each node has only one in-neighbour. <P/>
```

gap/prop.gi
```diff
@@ -259,6 +259,10 @@ InstallMethod(IsFunctionalDigraph, "for a digraph by out-neighbours",
 [IsDigraphByOutNeighboursRep],
 D -> ForAll(OutNeighbours(D), x -> Length(x) = 1));
 
+InstallMethod(IsPermutationDigraph, "for a digraph by out-neighbours",
+[IsDigraphByOutNeighboursRep],
+D -> IsFunctionalDigraph(D) and ForAll(InNeighbours(D), x -> Length(x) = 1));
```
wilfwilson, line 264: What you wrote is entirely valid, but I think this way should be slightly more efficient: constructing the in-neighbours of a digraph can be somewhat expensive, but the method for `DigraphSources` uses something called "blists" which I reckon will be a bit faster in the average case.
```suggestion
D -> IsFunctionalDigraph(D) and IsEmpty(DigraphSources(D)));
```

wilfwilson (commented):
Thanks for this contribution, I like it! I've noticed just a couple of really minor things that I'll change, with explanations (including the addition of a `TrueMethod`: this teaches Digraphs that every permutation digraph is necessarily a functional digraph).

wilfwilson (approved):

Issue #518, opened 2021-11-24
Segfault from calling `OutNeighbours` with the wrong argument in GAP


wilfwilson:
```
gap> OutNeighbours([CycleDigraph(4)]);
/usr/local/bin/gap: line 3: 25995 Segmentation fault: 11
```

james-d-mitchell:
Oops, this is caused by checking the value of `CALL_1ARGS` which is an `Obj` not a `bool`, fixed it in #519

wilfwilson:
Resolved in v1.5.1.

PR #519, opened 2021-11-25
kernel: fix check in OutNeighbours

james-d-mitchell:
This PR fixes Issue #518.

### DIFF
--- a/src/digraphs.c
+++ b/src/digraphs.c
@@ -86,7 +86,7 @@ Obj FuncOutNeighbours(Obj self, Obj D) {
   if (!RNamOutNeighbours) {
     RNamOutNeighbours = RNamName("OutNeighbours");
   }
-  if (!CALL_1ARGS(IsDigraph, D)) {
+  if (CALL_1ARGS(IsDigraph, D) != True) {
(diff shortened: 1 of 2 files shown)


wilfwilson (approved):
Cheers

james-d-mitchell:
The CI fails because of some issues with the wording of errors, any idea how I can add this test and get around this @wilfwilson ?
~~~~
Error, expected a digraph, not a homogeneous plain list
Error, expected a digraph, not a list (plain,hom)
~~~~

wilfwilson:
I've pushed a slightly different test, I hope you don't mind. If you do `OutNeighbours(OutNeighbours)` then you get a consistent error message across GAP versions:
```
Error, expected a digraph, not a function
```
Once we require GAP 4.11, these kinds of problems will be straightforward to work around because you can use the syntax `#@if`, `#@else`, and `#@endif` in test files to test stuff conditionally, and to expect different outputs.

james-d-mitchell:
> I've pushed a slightly different test, I hope you don't mind.

Thanks! 

> Once we require GAP 4.11, these kinds of problems will be straightforward to work around because you can use the syntax `#@if`, `#@else`, and `#@endif` in test files to test stuff conditionally, and to expect different outputs.

I didn't know this! Awesome!

PR #523, opened 2021-12-06
UnweightedBellmanFord

baydrea:


james-d-mitchell:
Thanks for the PR @baydrea ! Before I review this, can you please try to fix the CI? Everything has passed except linting, you can install gaplint by doing:
~~~
python3 -m pip install gaplint
~~~
then run it on the file `gap/attr.gi`  by doing (from within the `digraphs` directory):
~~~
gaplint gap/attr.gi
~~~
and then fix the errors indicated.

doc/attr.xml
```diff
@@ -902,6 +902,36 @@ gap> DigraphShortestDistances(D);
...
+  <Attr Name="UnweightedBellmanFord" Arg="digraph","source"/>
+  <Returns>A list of integers or <K>fail</K>.</Returns>
+  <Description>
+    If <A>digraph</A> is a digraph with <M>n</M> vertices, then this
+    function returns a list with two sublists of <M>n</M> entries, where each entry is
+    either a non-negative integer, or <K>fail</K>.  <P/>
+
+    If there is a directed path from <A>source</A> to vertex <C>i</C>, then
+    the value of the entry i in the first sublist is the length of the shortest such directed
+    path and the entry i in the second sublist is the vertex preceding the vertex of that entry. If no such directed path exists,     then the value of i is <C>fail</C>.  We use the convention that the distance from every vertex to
```
james-d-mitchell, line 916: I'm not sure what `the vertex preceding the vertex of that entry` means, can you please clarify? And also please wrap the line so that it is at most 80 chars long.

doc/attr.xml
```diff
@@ -902,6 +902,37 @@ gap> DigraphShortestDistances(D);
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="UnweightedBellmanFord">
+<ManSection>
+  <Attr Name="UnweightedBellmanFord" Arg="digraph","source"/>
+  <Returns>A list of integers or <K>fail</K>.</Returns>
```
flsmith, line 908: This is not what the rest of the documentation claims as the return value

doc/attr.xml
```diff
@@ -902,6 +902,37 @@ gap> DigraphShortestDistances(D);
...
+  <Description>
+    If <A>digraph</A> is a digraph with <M>n</M> vertices, then this
```
flsmith, line 910: and `<C>source</C>` is?

doc/attr.xml
```diff
@@ -902,6 +902,37 @@ gap> DigraphShortestDistances(D);
...
+    function returns a list with two sublists of <M>n</M> entries, where each entry is
```
flsmith, line 911: "...a list `<C>L = [distances, predecessors]</C>`..." so they can be referred to later

doc/attr.xml
```diff
@@ -902,6 +902,37 @@ gap> DigraphShortestDistances(D);
...
+    either a non-negative integer, or <K>fail</K>.  <P/>
+
+    If there is a directed path from <A>source</A> to vertex <C>i</C>, then for each i-th entry the first sublist contains
```
flsmith, line 914: `<C>i</C>th`

doc/attr.xml
```diff
@@ -902,6 +902,37 @@ gap> DigraphShortestDistances(D);
...
+<ManSection>
+  <Attr Name="UnweightedBellmanFord" Arg="digraph","source"/>
+  <Returns>A list of integers or <K>fail</K>.</Returns>
+  <Description>
+    If <A>digraph</A> is a digraph with <M>n</M> vertices, then this
+    function returns a list with two sublists of <M>n</M> entries, where each entry is
+    either a non-negative integer, or <K>fail</K>.  <P/>
+
+    If there is a directed path from <A>source</A> to vertex <C>i</C>, then for each i-th entry the first sublist contains 
+    the length of the shortest directed path to that i-th vertex and second sublist contains the vertex preceding that i-th
```
flsmith, line 915: This could be clearer - maybe something like "... vertex `<C>i</C>`, then `<C>distances[i]</C>` contains the length of the shortest path from `<C>source</C>` to vertex <C>i</C>, and `<C>predecessors[i]</C>` contains the vertex before `<C>i</C>` on such a shortest path."

doc/attr.xml
```diff
@@ -902,6 +902,37 @@ gap> DigraphShortestDistances(D);
...
+  <Attr Name="UnweightedBellmanFord" Arg="digraph","source"/>
+  <Returns>A list of integers or <K>fail</K>.</Returns>
+  <Description>
+    If <A>digraph</A> is a digraph with <M>n</M> vertices, then this
+    function returns a list with two sublists of <M>n</M> entries, where each entry is
+    either a non-negative integer, or <K>fail</K>.  <P/>
+
+    If there is a directed path from <A>source</A> to vertex <C>i</C>, then for each i-th entry the first sublist contains 
+    the length of the shortest directed path to that i-th vertex and second sublist contains the vertex preceding that i-th 
+    vertex. If no such directed path exists, then the value of i is <C>fail</C>.
```
flsmith, line 916: "the value of `<C>distances[i]</C>` and `<C>predecessors[i]</C>` is `<C>fail</C>`.

doc/attr.xml
```diff
@@ -902,6 +902,37 @@ gap> DigraphShortestDistances(D);
...
+  <Returns>A list of integers or <K>fail</K>.</Returns>
+  <Description>
+    If <A>digraph</A> is a digraph with <M>n</M> vertices, then this
+    function returns a list with two sublists of <M>n</M> entries, where each entry is
+    either a non-negative integer, or <K>fail</K>.  <P/>
+
+    If there is a directed path from <A>source</A> to vertex <C>i</C>, then for each i-th entry the first sublist contains 
+    the length of the shortest directed path to that i-th vertex and second sublist contains the vertex preceding that i-th 
+    vertex. If no such directed path exists, then the value of i is <C>fail</C>.  
+    We use the convention that the distance from every vertex to
```
flsmith, line 917: "any" rather than "every", and no need for "for all vertices..."

gap/attr.gi
```diff
@@ -700,6 +700,52 @@ end);
 # returns the vertices (i.e. numbers) of <D> ordered so that there are no
 # edges from <out[j]> to <out[i]> for all <i> greater than <j>.
 
+InstallMethod(UnweightedBellmanFord, "for a digraph by out-neighbours",
+[IsDigraph, IsPosInt],
+function(digraph, source)
+  local distance, n, predecessor, i, inf, u, v, edge, w;
+  n := DigraphNrVertices(digraph);
+  # wouldn't  work for weighted digraphs
```
flsmith, line 708: This comment isn't necessary - the name of the function is enough.

gap/attr.gi
```diff
@@ -700,6 +700,52 @@ end);
...
+  inf := n + 1;
+  distance := List([1 .. n], x -> 0);
+  predecessor := List([1 .. n], x -> 0);
+  for i in DigraphVertices(digraph) do
```
flsmith, line 712: This loop is unnecessary - you should just do `distance := List([1 .. n], x -> inf)` to initialise the `distances` list to contain the value `inf`, and the same for `predecessor` (which already contains all `0`s from the line above).

Better would be to use the descriptive function `ListWithIdenticalEntries(n, inf)`.

gap/attr.gi
```diff
@@ -700,6 +700,52 @@ end);
...
+    distance[i] := inf;
+    predecessor[i] := 0;
+  od;
+  distance[source] := 0;
+  for i in [1 .. n - 1] do
+    for edge in DigraphEdges(digraph) do
+      u := edge[1];
+      v := edge[2];
+      # only works for unweighted graphs, w needs to be changed into a variable
```
flsmith, line 721: But for this function which is only supposed to do the unweighted version, is there any reason to have this comment or use the variable `w`? Particularly since `w` is not very descriptive.
There is some sort of argument for using a constant `edge_weight := 1` defined near the start of the function, but most people would probably just replace `w` by `1` everywhere.

gap/attr.gi
```diff
@@ -700,6 +700,52 @@ end);
...
+      fi;
+    od;
+  od;
+  for edge in DigraphEdges(digraph) do
+    u := edge[1];
+    v := edge[2];
+    # only works for unweighted graphs, w needs to be changed into a variable
+    w := 1;
+    if distance[u] + w < distance[v] then
+      Print("Graph contains a negative-weight cycle");
```
flsmith, line 735: We almost never want our functions to `Print` things. In the unweighted case, there's no need for this check is there? It's impossible to have a negative-weight cycle when all edges have weight 1.

gap/attr.gi
```diff
@@ -700,6 +700,52 @@ end);
...
+    fi;
+  od;
+  for i in DigraphVertices(digraph) do
+    if distance[i] >= inf then
```
flsmith, line 739: Can this value really ever be greater than `inf`?

gap/attr.gi
```diff
@@ -700,6 +700,52 @@ end);
...
+      distance[i] := fail;
+    fi;
+    if predecessor[i] = 0 then
```
flsmith, line 742: I would be tempted to use `inf` to represent failure for both `distance` and `predecessor`, rather than `inf` and `0` - especially since they can only fail at the same time. Since they can only fail at the same time, you could also replace the two loops with one that looks like
```
for i in DigraphVertices(digraph) do
    if distance[i] = inf or predecessor[i] = inf then
        ...
```

gap/attr.gi
```diff
@@ -700,6 +700,52 @@ end);
...
+      if distance[u] + w < distance[v] then
+        distance[v] := distance[u] + w;
+        predecessor[v] := u;
+      fi;
+    od;
+  od;
+  for edge in DigraphEdges(digraph) do
+    u := edge[1];
+    v := edge[2];
+    # only works for unweighted graphs, w needs to be changed into a variable
```
flsmith, line 732: As above.

flsmith (changes requested):
Thanks for the pull request. I've left a few things that should be addressed before we merge this in.

PR #526, opened 2022-01-10
digraph: add AsDigraph for partial perms and fix doc

james-d-mitchell:
This PR adds an `AsDigraph` method for partial perms, and cleans up the documentation for `AsDigraph` as a bonus.


wilfwilson (approved):
Nice

PR #528, opened 2022-02-02
Add `IsDistributiveLatticeDigraph` property

MTWhyte:
**Note: This PR is on top of #538, which is in turn on top of #533.**

This PR adds the property `IsDistributiveLatticeDigraph`, which determines whether or not a digraph is a distributive lattice digraph. A digraph `D` is a distributive lattice digraph if it is a lattice digraph, and the operations `PartialOrderDigraphMeetOfVertices` and `PartialOrderDigraphJoinOfVertices` distributive over each other.

A lattice is distributive if and only if it the 'diamond' (`M3`) and 'pentagon' (`N5`) lattices described [here](https://en.wikipedia.org/wiki/Distributive_lattice#Characteristic_properties) are not sublattices. The method uses this characterisation.

Given a lattice digraph `D`, every lattice embedding from `M3` or `N5` into `D` is necessarily a digraph embedding of their corresponding partial order digraphs. The embeddings of `M3` and `N5` into `D` as digraphs are found, and it is determined whether or not they are lattice embeddings -- that is, the operations `PartialOrderDigraphMeetOfVertices` and `PartialOrderDigraphJoinOfVertices` are respected by the embeddings.

If it is found that one such digraph embedding is a lattice embedding, `IsDistributiveLatticeDigraph` returns `false`. Otherwise, it returns `true`.

I tested this approach against the simple approach of iterating through triplets of vertices, and found it to be faster. Producing 2500 random lattices with `RandomLattice(30)` (each call yields a random lattice on somewhere between 30 and 2*30 = 60 vertices), and comparing the runtimes for each method gives the following:

image

Each point along the horizontal represents an individual lattice (the same being used for both methods). I'm referring to the lattice embedding method as the "homomorphism method". This axis is ordered by runtime for the trivial method (blue), and the garbage collection time is filtered out.

Comparing the methods on 1000 random lattices generated by `RandomLattice(n)` for `n` in the range `[9 .. 30]` gives the following:

image

I've also compared the methods on all lattices on 11 and 12 points, and the homomorphism method was better on these. I'm working on a lattice-specific homomorphism finding algorithm, which should make this approach faster still.

### DIFF
--- a/doc/prop.xml
+++ b/doc/prop.xml
@@ -1311,7 +1311,8 @@ true
     Equivalently, a <E>distributive lattice digraph</E> is a <E>lattice digraph</E>
     in which the <E>lattice digraphs</E> representing <M>M3</M> and <M>N5</M> are
     not embeddable as lattices
-    (see <URL>https://en.wikipedia.org/wiki/Distributive_lattice)</URL>.<P/>
+    (see <URL>https://en.wikipedia.org/wiki/Distributive_lattice</URL> and
+    <Ref Prop="IsLatticeEmbedding"/>).<P/>
 
     &MUTABLE_RECOMPUTED_PROP;
 
--- a/doc/prop.xml
+++ b/doc/prop.xml
@@ -1295,6 +1295,44 @@ true
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="IsDistributiveLatticeDigraph">
+<ManSection>
+  <Prop Name="IsDistributiveLatticeDigraph" Arg="digraph"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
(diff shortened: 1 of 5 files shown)


MTWhyte:
I see the code coverage test has failed (oddly, this didn't happen in draft mode). I'll fix this shortly.

MTWhyte:
I've (hopefully) fixed the code coverage error by removing some lines which could never actually be run.

james-d-mitchell:
I've labelled this PR as work in progress, because I think that's accurate, if not, @MTWhyte please let me know!

MTWhyte:
**Todo** There should be a separate method for `LatticeEmbedding`.

MTWhyte:
I have updated the method to use the lattice embedding function from #538. This should not be merged before #533 and #538.

PR #530, opened 2022-03-13
Fix compatibility with grape

fingolfin:
When starting GAP with `-A`, then first loading digraphs and afterwards
loading grape, the following error was produced:

    Error, variable `Vertices' is not bound to an operation at GAPROOT/lib/oper.g:940 called from
    <function "DeclareOperation">( <arguments> )
     called from read-eval loop at GAPROOT/pkg/grape-4.8.5/lib/grape.g:566
    you can 'quit;' to quit to outer loop, or
    you can 'return;' to continue

### DIFF
--- a/init.g
+++ b/init.g
@@ -32,7 +32,7 @@ if not IsBound(IsGraph) then
   IsGraph := ReturnFalse;
 fi;
 if not IsBound(Vertices) then
-  Vertices := IdFunc;
+  DeclareOperation("Vertices", [IsRecord]);
 fi;
 if not IsBound(Adjacency) then
   Adjacency := IdFunc;


wilfwilson (approved):
I think this is fine; @james-d-mitchell? (the purpose of this line was to allow the conversion of Grape graphs into Digraphs objects; but to do it gracefully in the case that Grape was not available).

fingolfin:
It would be really nice to have this merged and in a release of this package, as it breaks e.g. the tests for the `agt` package under certain circumstances (simply because it depends on  both digraphs and grape, and if those two are loaded in the wrong order, then BOOM).

wilfwilson:
Sure. It's a minor change so I'm happy to merge, I imagine we'll be able to make a release soon.

PR #531, opened 2022-03-17
Random filters

KamranKSharma:


doc/digraph.xml
```diff
@@ -765,23 +765,56 @@ gap> List(D, x -> Size(x));
...
+      
+    The other implemented filters are as follows:
+      <Ref Filt='IsConnectedDigraph'/>, <Ref Filt='IsSymmetricDigraph'/>, <Ref Filt='IsAcyclicDigraph'/>, <Ref Filt='IsEulerianDigraph'/>, <Ref Filt='IsHamiltonianDigraph'/>.<P/>
+      
+    For <Ref Filt='IsConnectedDigraph'/>, a random tree is first created independent of the value of <A>p</A>,
+      guaranteeing connectivity (with n-1 edges), and then edges are added between the remaining pairs
+      of vertices with probability approximately <A>p</A>.<P/>
+      
+    For <Ref Filt='IsHamiltonianDigraph'/>, a random Hamiltonian Cycle is first created independent of the
+      value of <A>p</A> (with n edges), and then edges are added between the remaining pairs of vertices
```
mtorpey, line 777: `n` should be `<A>n</A>` (since it's an Argument).

doc/digraph.xml
```diff
@@ -765,23 +765,56 @@ gap> List(D, x -> Size(x));
...
+      <A>p = 1</A> does not necessarily guarantee a complete digraph. Instead, it guarantees that all edges 
+      considered up to the point where the cycle stops, are added.<P/>
+      
+    For <Ref Filt='IsAcyclicDigraph'/> and <Ref Filt='IsSymmetricDigraph'/>, edges are added between any 
+      pairs of vertices with probability approximately <A>p</A>.<P/>
 
     If <A>n</A> is a positive integer, then this function returns a random
     digraph with <A>n</A> vertices and without multiple edges. The result
-    may or may not have loops.<P/>
+    may or may not have loops. If using <Ref Filt='IsAcyclicDigraph'/>, the resulting graph
```
mtorpey, line 793: Well spotted!  This is a good addition.

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
   return DigraphNC(IsMutableDigraph, RANDOM_DIGRAPH(n, Int(p * 10000)));
 end);
 
+BindGlobal("DIGRAPHS_FillOutGraph", function(n, p, adjacencyList)
```
mtorpey, line 1426: It'd be good to have a comment here explaining briefly what `DIGRAPHS_FillOutGraph` is supposed to do.

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+
+    local i, j;
+
+    for i in [1 .. n] do
+        for j in [1 .. n] do
+            if (not (j in adjacencyList[i])) then
+                 if Float(Random([0 .. 99])/100) < Float(p) then
```
mtorpey, line 1433: Similarly, you might want to create `[0 .. 99]` somewhere, since this is created every time.

I don't think `Float` is needed here, is it?  GAP is happy to deal with rationals.

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+    fi;
+  
+    vertices:= [1 .. n];
+    
+    # Starting from a random vertex, we create a Hamiltonian cycle
+    startVertex := Remove(vertices, Random(vertices));
+    hamiltonianCycle := [startVertex];
+    
+    # While there are remaining n-1 vertices to be added to the Hamiltonian cycle
+    for x in [1 .. n-1] do
```
mtorpey, line 1473: Put spaces around minus signs: `n - 1`.

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+
+end);
+
+InstallMethod(RandomDigraphCons,
+"for IsHamiltonianDigraph, a positive integer, and a float",
+[IsHamiltonianDigraph, IsPosInt, IsFloat],
+function(filt, n, p)
+    local adjacencyList, vertices, startVertex, hamiltonianCycle, x, i, j, graph;
+  
+    adjacencyList := [];
```
mtorpey, line 1450: Instead of `[]`, you should use `EmptyPlist(n)`, since we know that we're going to add `n` elements to the list.  This seems weird, but it tells GAP to allocate the right amount of memory and avoids waste.

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+  
+    for i in [1 .. n] do
+        Add(adjacencyList, []);
```
mtorpey, line 1453: We don't know how big this will be, so `[]` is fine here.

gap/digraph.gi
mtorpey: Several more cases of `Float` below, which should also be checked.

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+    od;
+  
+    # Edge Case
+    if n = 1 then
+        if Float(Random([0 .. 99])/100) < Float(p) then
+            Add(adjacencyList[1], 1);
+            return Digraph(adjacencyList);
```
mtorpey, line 1460: Put this return line after the `fi`.  Then it's only needed once!

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+        else
+            return Digraph(adjacencyList);
+        fi;
+    fi;
+  
+    vertices:= [1 .. n];
```
mtorpey, line 1466: Watch the spacing around the `:=`

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+    # Starting from a random vertex, we create a Hamiltonian cycle
+    startVertex := Remove(vertices, Random(vertices));
+    hamiltonianCycle := [startVertex];
+    
+    # While there are remaining n-1 vertices to be added to the Hamiltonian cycle
+    for x in [1 .. n-1] do
+        # Create a random edge from the last vertex in the cycle, to random vertex not in the cycle
+        i := hamiltonianCycle[x];
+        j := Random(vertices);
+        Remove(vertices, Position(vertices, j));
```
mtorpey, line 1477: To avoid a search here, replace these two lines with:
```
j := Remove(vertices, Random([1 .. Length(vertices)]));
```
It looks worse, but the linear search is expensive.

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+        else
+            return Digraph(adjacencyList);
+        fi;
+    fi;
+  
+    vertices:= [1 .. n];
+    
+    # Starting from a random vertex, we create a Hamiltonian cycle
+    startVertex := Remove(vertices, Random(vertices));
+    hamiltonianCycle := [startVertex];
```
mtorpey, line 1470: Again, we might be better off with:
```
hamiltonianCycle := EmptyPlist(n);
hamiltonianCycle[1] := startVertex;
```
even though this is weird.

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+    if n = 1 then
+        if Float(Random([0 .. 99])/100) < Float(p) then
+            Add(adjacencyList[1], 1);
+            return Digraph(adjacencyList);
+        else
+            return Digraph(adjacencyList);
+        fi;
+    fi;
+
+    vertices:= [1 .. n];
```
mtorpey, line 1511: Move this earlier and use it in line 1497.

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+    # Starting from a random vertex, we create an Eulerian circuit
+    startVertex := Random(vertices);
+    circuitVertex := startVertex;
+    continueCircuit := true;
+
+    # This will keep track of the vertices left to consider when deciding whether to extend the cycle. For example,
+    # if we want to extend from the current circuitVertex, we can check - verticesToConsider[circuitVertex] -
+    # to see which vertices it hasn't previously considered.
+    verticesToConsider := [];
+    for i in [1 .. n] do
```
mtorpey, line 1522: This should be `for i in vertices do`.

However, don't use `vertices` in the following line (inside the loop).  We want each list to be created new.

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+        Add(verticesToConsider, [1 .. n]);
+    od;
+
+    # Eulerian graphs must be connected, so we store a list of connected vertices and check if a vertex isn't in the list
+    connectedVertices := [startVertex];
+
+    while continueCircuit do
+        while Length(verticesToConsider[circuitVertex]) > 0 do
+            # From the remaining vertices to check, select a random one to see if an edge will be added
+            i := Random(verticesToConsider[circuitVertex]);
```
mtorpey, line 1532: As above, better to select a random *position* rather than a random *element*.  Then we don't have to use `Position` below, which is expensive.

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+    if circuitVertex <> startVertex then
+        # If an edge already exists between the finish vertex and the start vertex,
+        # we must find another path to the start vertex to avoid generating a multidigraph.
+        while (startVertex in adjacencyList[circuitVertex]) do
+            # Here we consider all possible edges again, including previously rejected edges, to guarantee a path
+            # back to the start
+            verticesToCheck := [1 .. n];
+            while Length(verticesToCheck) > 0 do
+                i := Random(verticesToCheck);
+                Remove(verticesToCheck, Position(verticesToCheck, i));
```
mtorpey, line 1567: Again, avoid using `Position`.

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+
+    # While there are n-1 remaining vertices to be added to the tree
+    for x in [1 .. n-1] do
+        # Create an edge from a random vertex in the tree, to a random vertex outside of it
+        i := Random(tree);
+        j := Random(vertices);
+        Remove(vertices, Position(vertices, j));
+        Add(tree, j);
+        Add(adjacencyList[i], j);
+    od;
```
mtorpey, line 1608: James was suggesting that this should be a separate function for generating random trees.  But for now I suggest we leave it here, and remember for later.

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+    return graph;
+end);
+
+InstallMethod(RandomDigraphCons,
+"for IsConnectedDigraph, a positive integer, and a float",
+[IsConnectedDigraph, IsPosInt, IsFloat],
+function(filt, n, p)
+    local adjacencyList, vertices, startVertex, tree, x, i, j, graph;
+
+    adjacencyList := [];
```
mtorpey, line 1588: Again, instead of `[]` we can use `EmptyPlist(n)`.

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+    return graph;
+end);
+
+InstallMethod(RandomDigraphCons,
+"for IsAcyclicDigraph, a positive integer, and a float",
+[IsAcyclicDigraph, IsPosInt, IsFloat],
+function(filt, n, p)
+    local adjacencyList, vertices, i, j, graph;
+
+    adjacencyList := [];
```
mtorpey, line 1622: `EmptyPlist(n)`

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+
+    vertices:= [1 .. n];
+
+    # We shuffle the vertices and treat the order of the vertices as a hierarchy where all vertices to the right of
+    # a vertex in the list are potential edges. This idea is that the edges flow downwards in the hierarchy, avoiding cycles.
+    Shuffle(vertices);
+
+    # From position i in the list, we consider all vertices j to the right of it
+    for i in [1 .. n] do
+        for j in [i+1 .. n] do
```
mtorpey, line 1636: Spacing around the `+` symbol

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+    return graph;
+end);
+
+InstallMethod(RandomDigraphCons,
+"for IsSymmetricDigraph, a positive integer, and a float",
+[IsSymmetricDigraph, IsPosInt, IsFloat],
+function(filt, n, p)
+    local adjacencyList, vertices, i, j, graph;
+
+    adjacencyList := [];
```
mtorpey, line 1653: `EmptyPlist(n)`

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+                else
+                    Add(adjacencyList[i], j);
+                    Add(adjacencyList[j], i);
+                fi;
+            fi;
+        od;
+    od;
+
+    graph := Digraph(adjacencyList);
+    return graph;
```
mtorpey, line 1676: Just `return Digraph(adjacencyList);` should be fine.

mtorpey (changes requested):
A bunch of changes requested here, but all minor.

One general comment for something that comes up everywhere: don't call `Digraph(...)`.  Instead use `DigraphNC(...)`.  It's the same thing but without the user-level input checks, which aren't required if we're generating adjacency lists algorithmically.

gap/digraph.gi
mtorpey: Looks like you do need Float after all!

But get rid of `Float(p)` everywhere, since `p` should always be a float.

gap/digraph.gi
mtorpey: Spacing around the `:=`

gap/digraph.gi
```diff
@@ -1368,6 +1423,259 @@ function(filt, n, p)
...
+    od;
+
+    vertices:= [1 .. n];
+
+    # Starting from a random vertex, we first create a tree to guarantee connectivity
+    startVertex := Remove(vertices, Random(vertices));
+    tree := [startVertex];
+
+    # While there are n-1 remaining vertices to be added to the tree
+    for x in [1 .. n-1] do
```
mtorpey, line 1601: `n - 1` with spaces

gap/digraph.gi
mtorpey: But use `DigraphNC` everywhere.

mtorpey (approved):
I realised that I can push changes to this branch, so I did the required linting myself and have rebased.  This now looks like it'll pass, and I'm happy with it!

I'll merge it once the final checks are done.

Issue #428, opened 2021-03-03
Implement a function to 'join' two graphs by a common subgraph


bspiers:
Implement a function that takes as input two graphs and a specified common subgraph of the two graphs, with output being a single graph with all vertices and edges of the two original graphs preserved except that the common subgraphs of the two graphs are 'collapsed' (for example, taking the cyclic graph with 3 vertices twice and letting the common subgraph be a single edge would result in the 2-fan graph https://mathworld.wolfram.com/FanGraph.html).

Something similar already exists in the form of DigraphEdgeUnion, but the exact action above requires defining the second graph with a lot of extra vertices with no edges incident to them (including in the common subgraph) in order to get the desired result.

PR #532, opened 2022-03-17
Initial version of AmalgamDigraphs and AmalgamDigraphsIsomorphic.

finnbuck:
Based on [Issue #428](https://github.com/digraphs/Digraphs/issues/428). This pull request adds two functions: AmalgamDigraphs and AmalgamDigraphsIsomorphic. 

Description:

_The AmalgamDigraphs function takes as input two digraphs and two lists of vertices corresponding to two identical subdigraphs of the two input digraphs. It returns a new digraph which consists of the two input digraphs joined together by their common subdigraph in such a way that the edge connectivity between the vertices in the new digraph matches the edge connectivity of the two input digraphs._

_AmalgamDigraphsIsomorphic functions similarly, but the two lists of vertices given as inputs need only describe subdigraphs that are isomorphic to one another rather than have them be precisely equal. AmalgamDigraphsIsomorphic then rearranges the second list of subdigraph vertices so that the induced subdigraphs become equal and then calls AmalgamDigraphs._

_Both functions return a tuple of size two, with the first element being the output digraph and the second being a record which maps each vertex number in the second input digraph to its new vertex number in the output digraph. The mapping of the vertices of the first input digraph can be  seen as the identity mapping._

(The documentation for these two functions still needs to be written.)


gap/oper.gi
```diff
@@ -765,6 +765,91 @@ function(D1, D2, edge_function)
...
+  subLength := Length(subdigraphVertices1);
+
+  # 'map' is a mapping from the vertices of D2 to the vertices of the
+  # final output graph. The idea is to map the subdigraph vertices of D2
+  # onto the subdigraph vertices of D1 and then map the rest of the vertices
+  # of D2 to other (higher) values. The mapping from D1 to the output graph
+  # can be understood as the identity mapping.
+  map := rec();
+
+  for vertex in [1 .. subLength] do
```
james-d-mitchell, line 816:
```suggestion
  for vertex in [1 .. Length(subdigraphVertices1)] do
```

james-d-mitchell: The variable is only used once so you can skip it.

gap/oper.gi
```diff
@@ -765,6 +765,91 @@ function(D1, D2, edge_function)
...
+    map.(subdigraphVertices2[vertex]) := subdigraphVertices1[vertex];
+  od;
+
+  vertexList := Difference(DigraphVertices(D2), subdigraphVertices2);
```
james-d-mitchell, line 820: And you can delete lines 820 to 826.

gap/oper.gi
```diff
@@ -765,6 +765,91 @@ function(D1, D2, edge_function)
...
+  size := DigraphNrVertices(D1);
+  iterator := 1;
+  for vertex in vertexList do
+    map.(vertex) := iterator + size;
+    iterator := iterator + 1;
+  od;
+
+  # The problem with adding edges to the output graph was that the
+  # edges of of the subdigraph were added twice, creating multiple
```
james-d-mitchell, line 829:
```suggestion
  # edges of the subdigraph were added twice, creating multiple
```

gap/oper.gi
```diff
@@ -765,6 +765,91 @@ function(D1, D2, edge_function)
...
+
+InstallMethod(AmalgamDigraphs,
+"for a digraph, a digraph, a list, and a list",
+[IsDigraph, IsDigraph, IsList, IsList],
+function(D1, D2, subdigraphVertices1, subdigraphVertices2)
+  local D, map, vertex, vertexList, size, iterator, edgeList, subLength;
+
+  if not InducedSubdigraph(D1, subdigraphVertices1) =
+         InducedSubdigraph(D2, subdigraphVertices2) then
+    ErrorNoReturn(
```
james-d-mitchell, line 800:
```suggestion
    ErrorNoReturn("the subdigraph induced by the 3rd argument (a list) in the 1st argument (a digraph) does not equal the subdigraph induced by the 4th argument (a list) in the 2nd argument (a digraph)");
```

gap/oper.gi
```diff
@@ -765,6 +765,91 @@ function(D1, D2, edge_function)
...
+      "the two subdigraphs must be equal.");
```
james-d-mitchell, line 801:
```suggestion

```

gap/oper.gi
```diff
@@ -765,6 +765,91 @@ function(D1, D2, edge_function)
...
+  fi;
+
+  # Create a mutable copy so that the function also works on
+  # immutable input digraphs.
+  D := DigraphMutableCopy(D1);
```
james-d-mitchell, line 806: Maybe you only want to copy if `D1` is not mutable using `DigraphMutableCopyIfImmutable`?

gap/oper.gi
```diff
@@ -765,6 +765,91 @@ function(D1, D2, edge_function)
...
+  return AmalgamDigraphs(D1, D2, subdigraphVertices1, newSubdigraphVertices2);
+end);
+
+InstallMethod(AmalgamDigraphs,
+"for a digraph, a digraph, a list, and a list",
+[IsDigraph, IsDigraph, IsList, IsList],
+function(D1, D2, subdigraphVertices1, subdigraphVertices2)
+  local D, map, vertex, vertexList, size, iterator, edgeList, subLength;
+
+  if not InducedSubdigraph(D1, subdigraphVertices1) =
```
james-d-mitchell, line 798: If `D1` or `D2` is mutable, then `InducedSubdigraph` will modify them in place here, which I don't think you'd want!

gap/oper.gi
```diff
@@ -765,6 +765,91 @@ function(D1, D2, edge_function)
...
+  od;
+
+  # The problem with adding edges to the output graph was that the
+  # edges of of the subdigraph were added twice, creating multiple
+  # edges between certain pairs of points. A quick and readable fix
+  # would have been to use DigraphRemoveAllMultipleEdges, but I decided
+  # to check each of the edges being added to see if they were already
+  # in the subdigraph. This way the function does not end up adding edges
+  # only to delete them later.
+  edgeList := ShallowCopy(DigraphEdges(D2));
```
james-d-mitchell, line 835: I'd be tempted to just go with `DigraphRemoveAllMultipleEdges` here or maybe:

~~~
DigraphAddVertices(D, DigraphNrVertices(D2) - Length(subdigraphVertices2));

for edge in DigraphEdges(D2) do
  if not (edge[1] in subdigraphVertices2 and edge[2] in subdigraphVertices2) then
    DigraphAddEdge(D, edge);
  fi;
od
~~~

gap/oper.gi
```diff
@@ -765,6 +765,91 @@ function(D1, D2, edge_function)
...
+  for vertex in subdigraphVertices2 do
+    newSubdigraphVertices2[
+      Position(subdigraphVertices2, vertex) ^ transformation] := vertex;
+  od;
+
+  return AmalgamDigraphs(D1, D2, subdigraphVertices1, newSubdigraphVertices2);
+end);
+
+InstallMethod(AmalgamDigraphs,
+"for a digraph, a digraph, a list, and a list",
```
james-d-mitchell, line 793: Thinking about this some more, I think there should be 5 arguments for this function:

1, 2: `D1` and `D2`
3, 4: `S` induced subdigraph of `D1` and `D2` respectively
5, 6: `map1` and `map2` transformation objects that embed `S` into `D1` and `S` into `D2`, respectively. 

This won't require many changes and is the most flexible way to specify this function, I think.

james-d-mitchell (changes requested):
@finnbuck looks great! Lots of small things to fix, but generally this is good! Some additional things:
* please also add some tests when the arguments are mutable digraphs, I don't think this'll work quite like you expect (see comments in the PR)
* Add a version that takes 4 digraphs, `D1`, `D2`, and `S` and a transformation `map1` (embedding `S` in `D1`), then computes `map2` (an embedding of `S` in `D2`) and calls `DigraphAmalgam(D1, D2, S, map1, map2);`
* Add a version that takes 3 digraphs: `D1`, `D2` and `S` and computes `map1` and `map2` (embeddings into `D1` and `D2` resp), and calls `DigraphAmalgam(D1, D2, S, map1, map2);`
* Add a version that takes 2 digraphs `D1` and `D2`, and calls `x := MaximumCommonSubdigraph(D1, D2);` and then returns `DigraphAmalgam(D1, D2, x[1], x[2], x[3]);`

finnbuck:
@james-d-mitchell thanks for the review, I'll get to work on that!

gap/oper.gi
finnbuck: That would work, would the idea be to change D1 in place if it is mutable?

gap/oper.gi
finnbuck: Ok, so based on the code here and some of the comments above, I just wanted to double check quickly if I understood the overall idea of the function correctly.

Essentially, I can't seem to find a way to construct the map in such a way that it could be represented as a transformation and without using something like the lines 820-826 and I also noticed this piece of code seems to be placing the edges differently from the way I would have expected. So I'm wondering if I've understood everything right.

As far as I understand it, if I want to add vertices to D  in order to create the output graph, then I need to have a way to map the non-subdigraph vertices of D2 to the newly added vertices (which will have a higher index than the number of vertices in D1). This is sort of what I had in mind when constructing the 'map' and is also what I intended to do with the lines 820-826 above. It also means that the map variable can't be represented as a Transformation as far as I know, since there will often be entries in map that will be larger than the length of the map (since map needs to have the same number of entries as D2).

PR #533, opened 2022-03-24
Improve `IsJoinSemilatticeDigraph` and similar methods

MTWhyte:
This PR implements an algorithm described in _Free Lattices_ by Ralph Freese et. al. (1995 edition), which determines whether or not a given partial order digraph is a meet/join semilattice, simultaneously computing the table of meets/joins.

There is an existing algorithm for this, added in #59, which I believe is fairly similar to what I have done here. I modified this existing method in order to simultaneously compute the meet/join tables, and compared the two. The existing method and that in this PR were very similar in terms of computation time, this one maybe being slightly faster, which I have discussed with @james-d-mitchell. Another advantage of the method in this PR is that there's an immediate textbook reference to it, which could be useful in the future.

The most significant modification is in the method for the internal function `DIGRAPHS_IsMeetJoinSemilattice`. Around this, the user functions have been altered so that `DIGRAPHS_IsMeetJoinSemilattice` is run only once, computing and storing both the properties and the tables. The functions `PartialOrderDigraphJoinOfVertices` and `PartialOrderDigraphMeetOfVertices` will use the join/meet tables of a join/meet semilattice digraph, if they are known.

<details>
  <summary>Todo list</summary>
  
  ## Todo
- [x] If a digraph has its adjacency matrix, use this to determine whether elements are less than or equal to each other. This can be achieved by using `IsDigraphEdge(D, u, v)` in place of `u in nbs[v]`.

- [x] Add speed timing graphs to PR text or the wiki

- [x] Add tests for new behaviour
- [x] Documentation for `JoinSemilatticeDigraphJoinTable` and `MeetSemilatticeDigraphJoinTable`
- [x] Add reference to _Free Lattices_ Listing 11.9 as a comment above `DIGRAPHS_IsMeetJoinSemilattice`  
</details>


gap/prop.gi
```diff
@@ -38,58 +38,92 @@ D -> IsConnectedDigraph(D) and IsEmpty(ArticulationPoints(D)));
...
+  for i in [1 .. N] do
+    ord[P[i]] := i;
+  od;
+
+  S := [];
+
+  for x in P do
+    tab[x, x] := x;
+    for y in S do
+      T := [];  # EmptyPlist didn't improve speed
```
MTWhyte, line 74: I meant to remove the comments here from `prop.gi` and add them as GitHub comments. I will remove them later.

MTWhyte:
**Error with mutable digraphs (note to self, mainly)**

I have noticed some strange behaviour with mutable digraphs, in what I have written. I get the following:

```
gap> D := Digraph(IsMutable, [[2, 3], [4], [4], []]);
<mutable digraph with 4 vertices, 4 edges>
gap> DigraphReflexiveTransitiveClosure(D);
<mutable digraph with 4 vertices, 9 edges>
gap> JoinSemilatticeDigraphJoinTable(D);
fail
gap> IsJoinSemilatticeDigraph(D);
false
gap> MeetSemilatticeDigraphMeetTable(D);
[ [ 1, 1, 1, 1 ], [ 1, 2, 1, 2 ], [ 1, 1, 3, 3 ], [ 1, 2, 3, 4 ] ]
gap> IsMeetSemilatticeDigraph(D);
true
gap> D := DigraphImmutableCopy(D);
<immutable digraph with 4 vertices, 9 edges>
gap> IsJoinSemilatticeDigraph(D);
true
gap> JoinSemilatticeDigraphJoinTable(D);
[ [ 1, 2, 3, 4 ], [ 2, 2, 4, 4 ], [ 3, 4, 3, 4 ], [ 4, 4, 4, 4 ] ]
```

This indicates something has gone wrong with how mutable digraphs are handles, which I shall investigate.

MTWhyte:
### Some work regarding computation times

Here are some graphs regarding the speeds of the methods I tested. The first graph refers to the _New method_, the _Existing adaptation_, and the _Naïve method_. These are described as follows:

- **_New method:_** The algorithm in this PR
- **_Existing adaptation:_** Adapting the old `DIGRAPHS_IsMeetJoinSemilattice`, storing meets/joins when they are found (Code is in the appendix)
- **_Naïve method:_** Constructing meet/join tables of lattices, by iterating through pairs `u` and `v` of vertices, using `PartialOrderDigraphJoinOfVertices(D, u, v)` (and analogously for meets). The commutativity of meets and joins was used, so each pair of vertices as a set was only considered once.

image

The methods seem to have similar complexity. It's worth noting that naïve method does not determine whether or not its argument is a meet/join semilattice, whereas the other two methods do this concurrently with computing the tables. The naïve method in green is really just there for illustration, but the principal question was which of the other two methods is better.

image

The above graph compares the time between the new method and the existing adaptation for determining the property _IsJoinSemilatticeDigraph_, but not the tables themselves. The existing adaptation seems marginally faster in this situation, I think because it can avoid adding a lot of stuff to lists in this case, whereas the new method must add a lot of stuff to lists as it computes the join table in the very process of determining whether its argument is a join semilattice.

In discussion with @james-d-mitchell, it was decided that it's better to just automatically compute the meet/join tables for meet/join semilattices in conjunction with determining whether or not they are in fact meet/join semilattices. This makes the first of the above graphs more relevant than the second.

image

This graph above runs the same process as the first graph on random lattices, but on a larger range of vertices. It appears that for larger `n`, the new method is marginally faster than the existing adaptation. Running a similar process on random digraphs at large, rather than those which are necessarily lattices, gives the following:

image

In all, the new method and the existing adaptation are reasonably similar, with the new method perhaps taking the lead as the number of vertices increases. The claimed time complexity of the new method is `O(n^2.5)`, where `n` is the number of vertices of the digraph. Doing computations for random lattices and random digraphs, I get the following

image
image

In both, I have fitted a line to some of the data where it looks reasonably linear. It being a log-log plot, the gradient of this line should be the complexity. This suggests that the measured time complexity is about 2. It's worth noting that this was just a rough measurement: `RandomDigraph(n)` returns a digraph with `n` vertices, but `RandomLattice(n)` returns a digraph with somewhere between `n` and `2 n` vertices. I've plotted this as if `n` is the number of vertices, but since the two are reasonably proportional it shouldn't be that important.



#### Appendix: Code for the _Existing adaptation_

```
DIGRAPHS_IsMeetJoinSemilattice2 := function(nbs, topo)
  local i, j, k, n, x, len, table;

  # The IsHomogenousList assumes the list <nbs> contains no or only empty lists
  n := Length(nbs);
  if topo <> false then
    table := List([1 .. n], x -> []);
  fi;
  for i in [1 .. n] do
    if topo <> false then
      table[topo[i]][topo[i]] := topo[i];
    fi;
    for j in [i + 1 .. n] do
      if j in nbs[i] or i in nbs[j] then
        if topo <> false then
          if j in nbs[i] then
            table[topo[i]][topo[j]] := topo[j];
            table[topo[j]][topo[i]] := topo[j];
          else
            table[topo[i]][topo[j]] := topo[i];
            table[topo[j]][topo[i]] := topo[i];
          fi;
        fi;
        continue;
      fi;
      x := Intersection(nbs[i], nbs[j]);
      if IsEmpty(x) then
        return false;
      fi;
      len := Length(x);
      k   := x[len];  # Check whether <k> is the meet of <i> and <j>
      if Length(nbs[k]) < len then
        if topo <> false then
          return fail;
        fi;
        return false;
      fi;
      if topo <> false then
        table[topo[i]][topo[j]] := topo[k];
        table[topo[j]][topo[i]] := topo[k];
      fi;
    od;
  od;
  if topo <> false then
    return table;
  fi;
  return true;
end;

IsJoinSemilattice2 := function(D)
  local topo;
  if not IsPartialOrderDigraph(D) then
    return false;
  fi;
  topo := DigraphTopologicalSort(D);
  D    := DigraphMutableCopy(D);
  D    := OnDigraphs(D, PermList(topo) ^ -1);
  Apply(D!.OutNeighbours, Set);
  return DIGRAPHS_IsMeetJoinSemilattice2(D!.OutNeighbours, false);
end;

JoinTable2 := function(D)
  local topo;
  if not IsPartialOrderDigraph(D) then
    return false;
  fi;
  topo := DigraphTopologicalSort(D);
  D    := DigraphMutableCopy(D);
  D    := OnDigraphs(D, PermList(topo) ^ -1);
  Apply(D!.OutNeighbours, Set);
  return DIGRAPHS_IsMeetJoinSemilattice2(D!.OutNeighbours, topo);
end;
```

gap/prop.gi
```diff
@@ -38,58 +38,99 @@ D -> IsConnectedDigraph(D) and IsEmpty(ArticulationPoints(D)));
...
+  for i in [1 .. N] do
+    ord[P[i]] := i;
+  od;
+
+  S := [];
+
+  for x in P do
+    tab[x, x] := x;
+    for y in S do
+      T := [];
```
MTWhyte, line 79: Using `EmptyPlist` didn't improve speed

gap/prop.gi
```diff
@@ -38,58 +38,99 @@ D -> IsConnectedDigraph(D) and IsEmpty(ArticulationPoints(D)));
...
+      for z in U[x] do
+        Add(T, tab[y, z]);
```
MTWhyte, line 81: Using `SetAdd` didn't improve speed

gap/oper.gi
```diff
@@ -2187,7 +2192,12 @@ InstallMethod(PartialOrderDigraphMeetOfVertices,
 "for a digraph and two positive integers",
 [IsDigraph, IsPosInt, IsPosInt],
 function(D, i, j)
-  local x, nbs, intr;
+  local x, nbs, intr, tab;
+
+  if HasMeetSemilatticeDigraphMeetTable(D) then
+    tab := MeetSemilatticeDigraphMeetTable(D);
+    return tab[i, j];
```
james-d-mitchell, line 2199:
```suggestion
    return MeetSemilatticeDigraphMeetTable(D)[i, j];
```

gap/attr.gi
```diff
@@ -2675,3 +2675,27 @@ function(D)
   D := DigraphReflexiveTransitiveReduction(DigraphMutableCopyIfMutable(D));
   return DIGRAPHS_NonSemimodularPair(InNeighbours(D)) = fail;
 end);
+
+InstallMethod(JoinSemilatticeDigraphJoinTable, "for a digraph",
```
james-d-mitchell, line 2679: I completely understand the rationale for the name, but I think `DigraphJoinTable` would better!

gap/prop.gi
```diff
@@ -38,58 +38,99 @@ D -> IsConnectedDigraph(D) and IsEmpty(ArticulationPoints(D)));
...
+InstallMethod(DIGRAPHS_IsMeetJoinSemilatticeDigraph, "for a digraph and a bool",
+[IsDigraph, IsBool],
+function(D, join)
+  local P, U, ord, tab, S, N, i, x, T, l, q, z, y;
+
+  if not IsPartialOrderDigraph(D) then
+    return false;
+  fi;
+
+  D   := DigraphImmutableCopyIfMutable(D);
```
james-d-mitchell, line 52:
```suggestion
  D   := DigraphMutable(D);
```

gap/prop.gi
```diff
@@ -38,58 +38,99 @@ D -> IsConnectedDigraph(D) and IsEmpty(ArticulationPoints(D)));
...
+
+  if IsMultiDigraph(D) then
+    D := DigraphRemoveAllMultipleEdges(D);
```
james-d-mitchell, line 55:
```suggestion
    DigraphRemoveAllMultipleEdges(D);
```

gap/prop.gi
MTWhyte: I think this is replaced by what we discussed later on

gap/prop.gi
```diff
@@ -38,58 +38,99 @@ D -> IsConnectedDigraph(D) and IsEmpty(ArticulationPoints(D)));
...
-      x := Intersection(nbs[i], nbs[j]);
-      if IsEmpty(x) then
-        return false;
-      fi;
-      len := Length(x);
-      k := x[len];  # Check whether <k> is the meet of <i> and <j>
-      if Length(nbs[k]) < len then
+# The method below is based on Listing 11.9 of 'Free Lattices'
+# by Ralph Freese et. al.
+InstallMethod(DIGRAPHS_IsMeetJoinSemilatticeDigraph, "for a digraph and a bool",
```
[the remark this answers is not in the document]
MTWhyte: Done

gap/attr.gi
MTWhyte: Done

gap/prop.gi
```diff
@@ -38,60 +38,103 @@ D -> IsConnectedDigraph(D) and IsEmpty(ArticulationPoints(D)));
...
-  topo := Reversed(DigraphTopologicalSort(D));
-  D := OnDigraphs(DigraphMutableCopyIfMutable(D), PermList(topo) ^ -1);
-  list := InNeighboursMutableCopy(D);
-  Apply(list, Set);
-  return DIGRAPHS_IsMeetJoinSemilatticeDigraph(list);
+  copy   := DigraphMutableCopyIfMutable(D);
+  P      := Reversed(DigraphTopologicalSort(D));
+  U      := InNeighbours(DigraphReflexiveTransitiveReduction(copy));
+  tab    := DIGRAPHS_MeetJoinTable(D, P, U, false);
+  SetDigraphMeetTable(D, tab);
```
MTWhyte, line 126: Check that `D` is immutable first.

MTWhyte: Done

gap/prop.gi
```diff
@@ -38,60 +38,103 @@ D -> IsConnectedDigraph(D) and IsEmpty(ArticulationPoints(D)));
...
-  topo := DigraphTopologicalSort(D);
-  D := DigraphMutableCopy(D);
-  D := OnDigraphs(D, PermList(topo) ^ -1);
-  Apply(D!.OutNeighbours, Set);
-  return DIGRAPHS_IsMeetJoinSemilatticeDigraph(D!.OutNeighbours);
+  copy   := DigraphMutableCopyIfMutable(D);
+  P      := DigraphTopologicalSort(D);
+  U      := OutNeighbours(DigraphReflexiveTransitiveReduction(copy));
+  tab    := DIGRAPHS_MeetJoinTable(D, P, U, true);
+  SetDigraphJoinTable(D, tab);
```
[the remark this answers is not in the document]
MTWhyte: Done

MTWhyte:
*Testing note*

For good measure in testing the correctness of this work, I have compared it to more elementary methods for computing the meet and join tables of lattices, and found them to give the same results on all 37622 lattices on 11 vertices. The elementary functions described are as follows:

```
JoinTable := function(D)
  local N, tab, i, j;
  N   := DigraphNrVertices(D);
  tab := List([1 .. N], x -> []);
  for i in [1 .. N] do
    for j in [i .. N] do
      tab[i][j] := PartialOrderDigraphJoinOfVertices(D, i, j);
      tab[j][i] := tab[i][j];
    od;
  od;
  return tab;
end;

MeetTable := function(D)
  local N, tab, i, j;
  N   := DigraphNrVertices(D);
  tab := List([1 .. N], x -> []);
  for i in [1 .. N] do
    for j in [i .. N] do
      tab[i][j] := PartialOrderDigraphMeetOfVertices(D, i, j);
      tab[j][i] := tab[i][j];
    od;
  od;
  return tab;
end;
```

This was run on the current `master`, so before any changes to `PartialOrderDigraphJoinOfVertices` and `PartialOrderDigraphMeetOfVertices`.

doc/z-chap5.xml
```diff
@@ -28,6 +28,7 @@
     <#Include Label="IsPreorderDigraph">
     <#Include Label="IsPartialOrderDigraph">
     <#Include Label="IsMeetSemilatticeDigraph">
+    <#Include Label="DigraphMeetTable">
```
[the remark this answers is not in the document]
MTWhyte: I'm not actually sure. I understand that `DigraphMeetTable` is the 'label' for the documentation entry discussing both `DigraphMeetTable` and `DigraphJoinTable`. I've done this based on how `IsLatticeDigraph`, `IsMeetSemilatticeDigraph` and `IsJoinSemilattice` digraph are documented. Calling these up in the documentation gives
```
6.3-3 IsMeetSemilatticeDigraph

  ‣ IsMeetSemilatticeDigraph( digraph ) ───────────────────────────── property
  ‣ IsJoinSemilatticeDigraph( digraph ) ───────────────────────────── property
  ‣ IsLatticeDigraph( digraph ) ───────────────────────────────────── property
  Returns:  true or false.
```
In the same sort of style, just having `DigraphMeetTable` point at the documentation for both the meet and join table gives
```
6.3-4 DigraphMeetTable

  ‣ DigraphMeetTable( digraph ) ──────────────────────────────────── attribute
  ‣ DigraphJoinTable( digraph ) ──────────────────────────────────── attribute
  Returns:  A matrix or fail.
```

I don't know if this is optimal, but I think it's how to get the same style.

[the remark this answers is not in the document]
flsmith: Ah, I see.

tst/standard/attr.tst
```diff
@@ -2804,6 +2804,36 @@ false
...
+<immutable digraph with 4 vertices, 2 edges>
+gap> D := DigraphAddEdge(D, [4, 3]);
+<immutable digraph with 4 vertices, 3 edges>
+gap> D := DigraphReflexiveTransitiveClosure(D);
+<immutable preorder digraph with 4 vertices, 8 edges>
+gap> DigraphJoinTable(D);
+[ [ 1, 2, 3, 3 ], [ 2, 2, 3, 3 ], [ 3, 3, 3, 3 ], [ 3, 3, 3, 4 ] ]
+gap> DigraphMeetTable(D);
+fail
+
```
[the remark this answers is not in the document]
flsmith: One last comment @MTWhyte - it would be nice to have these tests also explicitly check the known value of these joins/meets.

flsmith (approved):

MTWhyte:
Thanks @flsmith, I've added explicit meet/join checking now. It occurred to me that, for the larger example I added, the explicit checking I was doing was in the wrong order. Since this PR makes `PartialOrderDigraphJoinOfVertices` look up the appropriate entry of `DigraphJoinTable` if it is known, I've changed it to compute the test meets/joins before the table, so the other method is used. (I guess this doesn't matter for the mutable example, but I think it makes more sense to keep everything in the same 'style'.)

james-d-mitchell:
@wilfwilson any ideas why the cygwin job is failing? Seems to be an issue with the CI rather than with this PR.

wilfwilson:
Yes it's a problem with `gap-actions/setup-gap` as linked in the GitHub interface above this comment. I've tried something and hoping it works.

wilfwilson:
@james-d-mitchell It's getting further than it was before, hopefully it's sorted.

PR #535, opened 2022-03-29
More fixes for loading grape after digraphs

fingolfin:
When loading digraphs before grape, then some code kept printing a warning
saying "Grape is not loaded", even though grape was loaded. This patch fixes
that.

Sorry, I didn't notice this one before. 

Note that even with this patch, it will skip some test if one now tries to run the Digraphs test suite, That's because `DIGRAPHS_OmitFromTests` is initialized once and then not updated. This could be fixed, e.g. by removing `DIGRAPHS_OmitFromTests` and modifying the code which currently uses it to directly use `DIGRAPHS_IsGrapeLoaded()` -- the code would arguably be simpler, too. But I don't know if you have plans to use `DIGRAPHS_OmitFromTests` more or have other needs for it, so I didn't do that in this PR.

### DIFF
--- a/gap/grape.gi
+++ b/gap/grape.gi
@@ -132,7 +132,7 @@ function(D)
          "the Grape graph will have fewer\n#I  edges than the original,");
   fi;
 
-  if not DIGRAPHS_IsGrapeLoaded then
+  if not DIGRAPHS_IsGrapeLoaded() then
     Info(InfoWarning, 1, "Grape is not loaded,");
   fi;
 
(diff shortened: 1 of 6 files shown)


wilfwilson (approved):

wilfwilson:
Thanks @fingolfin and thanks for noticing about `DIGRAPHS_OmitFromTests`. I made #536 to keep track of that. Probably we'll adopt your suggestion.

PR #538, opened 2022-04-01
Add `LatticeDigraphEmbedding` method

MTWhyte:
This PR adds a method for finding an embedding from a lattice digraph into another lattice digraph.

A lattice embedding is an injective lattice homomorphism. In terms of partial order digraphs (of which lattice digraphs are a type) a lattice homomorphism is a digraph homomorphism, which respects the meets and joins of all pairs of vertices.

There were two competing methods for `LatticeDigraphEmbedding`. The one in this pull request is based on the digraph homomorphism finding algorithm, where the notion that "meets must maps to meets, and joins must map to joins" is enforced. The other method called `RepresentativesDigraphsEmbeddings`, iterating through vertices to search for a digraph embedding which respects the meets and joins. When I update this PR, I will include some timing graphs comparing the two. It was decided in conversation with @james-d-mitchell that the lattice-specific method of this PR was the way to go.

**The meet/join table functions of #533 are used here, so this shouldn't be merged before #533.**

**Todo**
- [x] Check things work for mutable digraphs
- [x] Perhaps provide better explanation in the documentation
- [x] Add timing graphs to PR
- [x] Consider optimisations
- [x] Change output to a transformation, rather than a list

### DIFF
--- a/gap/grahom.gi
+++ b/gap/grahom.gi
@@ -903,9 +903,15 @@ InstallMethod(IsLatticeHomomorphism,
 function(L1, L2, map)
   local N1, N2, x, y, meet1, meet2, join1, join2;
 
+  N1 := DigraphNrVertices(L1);
+  N2 := DigraphNrVertices(L2);
+
+  if LargestMovedPoint(map) > N1 then
+    return false;
+  fi;
+
   # We compute the join/meet table to avoid having to do this twice if L1 or L2
   # is mutable
-
   join1 := DigraphJoinTable(L1);
   meet1 := DigraphMeetTable(L1);
 
@@ -919,9 +925,6 @@ function(L1, L2, map)
   if join2 = fail or meet2 = fail then
     ErrorNoReturn("the 2nd argument (a digraph) must be a lattice digraph");
   fi;
-
-  N1 := DigraphNrVertices(L1);
-  N2 := DigraphNrVertices(L2);
 
   if Maximum(ImageSetOfTransformation(map, N1)) > N2 then
     return false;
--- a/gap/grahom.gi
+++ b/gap/grahom.gi
@@ -762,7 +762,7 @@ function(L1, L2)
   in1, in2, mask, defined, not_in_image, FindNextAmong, Recurse;
 
   p := PermList(Reversed(DigraphWelshPowellOrder(L1)));
-  L1 := OnDigraphs(L1, p);
+  L1 := OnDigraphs(L1, p ^ -1);
 
   # We compute the join/meet table to avoid having to do this twice if L1 or L2
   # is mutable
@@ -890,7 +890,7 @@ function(L1, L2)
 
   if Recurse(1, 1, false) then
     Append(map, [N1 + 1 .. N2]);
-    return Transformation(map) ^ p;
+    return p ^-1  * Transformation(map);
   fi;
   return fail;
 end);
--- a/gap/grahom.gi
+++ b/gap/grahom.gi
@@ -925,3 +925,23 @@ function(L1, L2, map)
   od;
   return true;
 end);
+
+InstallMethod(IsLatticeEndomorphism, "for a digraph and a transformation",
+[IsDigraph, IsTransformation],
+{L, map} -> IsLatticeHomomorphism(L, L, map));
+
+InstallMethod(IsLatticeEpimorphism,
+"for a digraph, a digraph, and a transformation",
+[IsDigraph, IsDigraph, IsTransformation],
+function(L1, L2, map)
+  return IsLatticeHomomorphism(L1, L2, map)
+    and OnSets(DigraphVertices(L1), map) = DigraphVertices(L2);
+end);
+
+InstallMethod(IsLatticeEmbedding,
+"for a digraph, a digraph, and a transformation",
+[IsDigraph, IsDigraph, IsTransformation],
+function(L1, L2, map)
+  return IsLatticeHomomorphism(L1, L2, map)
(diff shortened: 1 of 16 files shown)


MTWhyte:
Based on what exists for digraphs in general, I think it would be a good idea to add an `IsLatticeHomomorphism` method, which would work by iterating over all pairs of vertices and checking their meets and joins are respected by the argument map. This would naturally fit in with this PR, though also doing it at a smaller separate PR seems reasonable.

MTWhyte:
This is mainly just a note to myself, but when I write `IsLatticeHomomorphism` and similar, I don't think I even need to check if it is a digraph homomorphism. The fact it is a digraph homomorphism will, I think (but will double check) be guaranteed by the fact that meets and joins are respected.

MTWhyte:
The latest collection of commits adds a method `IsLatticeHomomorphism`, which takes two lattices and a transformation/permutation and tests whether the map is a lattice homomorphism. The 'friends' of `IsLatticeHomomorphism` in the style of `IsDigraphHomomorphism` are also added.

MTWhyte:
The following is a comparison of computation times between the embedding method implemented in this PR, and a more naïve version based on finding _digraph_ embeddings and checking which if any are homomorphisms of _lattices_.

[3d-embeddings.pdf](https://github.com/digraphs/Digraphs/files/8521783/3d-embeddings.pdf)

Each point in the horizontal plane represents a pair of digraphs. We try and embed the domain into the range. Both domain and range lattices are generated by `RandomLattice(n)`, with the value `n` being marked by the horizontal axes. The vertical axis represents the difference in computation times between the two methods (with a 20 second timeout), where a positive value means the method in this PR is faster. As there is a 20 second timeout, the time differences in the vicinity of 20 seconds may in fact be a great deal larger.

MTWhyte:
In considering optimisations, @james-d-mitchell considered relabelling vertices of the domain lattice according to their Welsh Powell order. This represents an improvement in speed, which gets better with increasing domain size.




In the above graphs, each point in the horizontal direction represents a pair of graphs; one generated by `RandomLattice(60)`, the other generated by `RandomLattice(n)` where `n` is `10`, `20` or `30` depending on the graph. The graph pairs are ordered along the horizontal axis by their 'no ordering' computation time.

gap/grahom.gi
```diff
@@ -754,3 +754,233 @@ function(D1, D2)
...
+  if Maximum(ImageSetOfTransformation(map, N1)) > N2 then
+    return false;
+  fi;
+  # The above checks if the <x ^ map> and <y ^ map> entries of
+  # meet2 and join2 exist
+
+  for x in [1 .. N1] do
+    for y in [1 .. N1] do
+      if meet2[x ^ map, y ^ map] <> meet1[x, y] ^ map
+          or join2[x ^ map, y ^ map] <> join1[x, y] ^ map then
```
flsmith, line 938: Again, you might not want to implement this - but do you really have to check this for every pair of elements? I would guess that you can do something with irreducible elements here.

tst/standard/grahom.tst
```diff
@@ -2498,6 +2498,211 @@ Error, the 1st argument (a digraph) must not satisfy IsMultiDigraph
...
+gap> LatticeDigraphEmbedding(G, D);
+fail
+gap> D := CycleDigraph(5);
+<immutable cycle digraph with 5 vertices>
+gap> G := Digraph([[1]]);
+<immutable digraph with 1 vertex, 1 edge>
+gap> LatticeDigraphEmbedding(D, G);
+Error, the 1st argument (a digraph) must be a lattice digraph
+gap> LatticeDigraphEmbedding(G, D);
+Error, the 2nd argument (a digraph) must be a lattice digraph
```
flsmith, line 2576: It would be good to have a few more examples on larger lattices - you could test them with `IsDigraphEmbedding`.

flsmith (commented):
Looks pretty good. A few comments, but you probably only want to address the ones about the documentation/tests.

MTWhyte:
Thanks for the review @flsmith. I've made the documentation and tests changes now. The other suggestions are good, though as you say I might want to leave them for the time being for the purposes of getting this work ready for assessment for the VIP.

gap/grahom.gi
```diff
@@ -754,3 +754,233 @@ function(D1, D2)
                         Transformation([1 .. Size(embedding2)], embedding2)];
 
 end);
+
+InstallMethod(LatticeDigraphEmbedding, "for a pair of digraphs",
```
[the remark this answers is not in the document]
james-d-mitchell: Yes, but it's more straightforward to write GAP code so we did this first :)

gap/grahom.gi
```diff
@@ -754,3 +754,233 @@ function(D1, D2)
...
+    if next = fail then
+      return false;
+    fi;
+
+    value := Position(conditions[depth][next], true, 0);
+    while value <> fail do
+      map[next]             := value;
+      defined[next]         := true;
+      not_in_image[value]   := false;
+      conditions[depth + 1] := StructuralCopy(conditions[depth]);
```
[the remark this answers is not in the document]
james-d-mitchell: I think the answer is "no", but again that's an optimisation we can make later.

gap/grahom.gi
james-d-mitchell: That's a good question, I don't immediately know the answer, but will add a comment to the code to indicate that this should be investigated!

james-d-mitchell (approved):
I'm made a couple of minor changes to the doc, and added a couple of comments related to the comments @flsmith made in his review. If this passes, I'll merge it.

Issue #500, opened 2021-09-17
Adding edges one-by-one is too slow (because of edge labels)


wilfwilson:
In my opinion, adding edges to a digraph one-by-one is too slow (even to a mutable digraph, which we are starting to do more often, especially with our standard examples; this was even the point of adding mutable digraphs).

For example, let's use this method to try to construct a chain digraph with only 100,000 vertices and therefore 99,999 edges:
```gap
gap> n := 10 ^ 5;; D1 := EmptyDigraph(IsMutableDigraph, n);;
gap> for i in [1 .. n - 1] do DigraphAddEdge(D1, i, i + 1); od; time;
280944
```
Nearly 5 minutes is way too long! By contrast, doing this with the `ChainDigraph` function takes very little time:
```gap
gap> D2 := ChainDigraph(n);; time;
17
gap> D1 = D2;
true
```
When I interrupt the slow computation, I always get a backtrace that looks like this, so I presume that the slow part of adding the edge is the accounting to do with edge labels:
```
^CError, user interrupt in
  p := Position( OutNeighboursOfVertex( D, v ), w )
 ; at /Users/Wilf/gap/pkg/digraphs/gap/labels.gi:96 called from
SetDigraphEdgeLabel( D, src, ran, 1
 ); at /Users/Wilf/gap/pkg/digraphs/gap/oper.gi:188 called from
DigraphAddEdge( D1, i, i + 1 ); at *stdin*
```
Therefore I think we need a better implementation of edge labels!

Anyway, in this example, I have never explicitly given this digraph any edge labels, so it is frustrating that the package is spending so long worrying about keeping track of the information that every edge has label 1.

james-d-mitchell:
I'm just observing this now too in an example I am trying to run, I think we need something like `HasEdgeLabels` so that we can check if they exist and if they don't, then don't try setting them. I may do something about this, time permitting!

james-d-mitchell:
Ah wait this already exists and is called `HaveEdgeLabelsBeenAssigned`, so I guess this is a matter of just using this more.

PR #540, opened 2022-04-06
Disable edge labels if not already set in some cases

james-d-mitchell:
This PR is related to #500, and ensures that edge labels aren't set when taking the reflexive closure of a digraph if the original digraph doesn't have any edge labels.


gap/attr.gi
```diff
@@ -1787,30 +1787,23 @@ function(D)
...
   C := DigraphMutableCopyIfImmutable(D);
-  list := C!.OutNeighbours;
-  Assert(1, IsMutable(list));
   for v in DigraphVertices(C) do
-    if not v in list[v] then
-      Add(list[v], v);
-      if not ismulti then
-        SetDigraphEdgeLabel(C, v, v, 1);
-      fi;
+    if not IsDigraphEdge(D, v, v) then
```
james-d-mitchell, line 1798: By way of explanation, we use `IsDigraphEdge(D, v, v)` because `D` might know, for example, its adjacency matrix and this is hopefully then used by `IsDigraphEdge`, whereas `C` is mutable so doesn't have any attributes.

gap/attr.gi
```diff
@@ -1787,30 +1787,23 @@ function(D)
...
+      DigraphAddEdge(C, v, v);
```
james-d-mitchell, line 1799: `DigraphAddEdge` only sets edge labels if `D` already has edge labels, so we avoid the overhead described in #500, by doing this.

wilfwilson (approved):
Fine by me.

In general, I would prefer methods to return the identical object if nothing has changed, but I realise that that's separate topic, and the package and the documentation are inconsistent about this. So I don't mind you making that change.

Issue #539, opened 2022-04-02
Allow disabling `-march=native` without disabling intrinsics?


fingolfin:
For creating portable binaries, it is important to not use `-march=native`. But right now to disable this without patching, one has to use `--with_intrinsics` but that actually disables more things that are perfectly fine to use in portable binaries, yet slow down things.

So perhaps an option could be added which disabled only `-march=native`?

(To be honest, I would do the opposite and not use `-march=native` by default, but that's just me, clearly you have decided otherwise :-) ).

james-d-mitchell:
Thanks @fingolfin, I'll be happy to remove the `-march=native` by default, I'm not sure that this was a very well reasoned decision, and given the difficulties it's causing, better to just remove it.

james-d-mitchell:
This was removed in v1.5.3 I think

PR #541, opened 2022-04-06
build: remove the default flag -march=native

james-d-mitchell:
This PR resolves Issue #539


james-d-mitchell:
@fingolfin does this fix the issue?

Issue #517, opened 2021-11-19
`String` for a digraph can sometimes wrongly return `"CycleDigraph(n)"`


wilfwilson:
The problem is that `IsCycleDigraph` is a property concerning _isomorphism type_, whereas `CycleDigraph(n)` creates a _specific_ cycle digraph. So this is fine:
```
gap> D1 := CycleDigraph(6);;
gap> String(D1);
"CycleDigraph(6)"
gap> D1 = EvalString(String(D1));
true
```
But this should return a different string:
```
gap> D2 := DigraphReverse(D1);;
gap> IsCycleDigraph(D2);
true
gap> String(D2);
"CycleDigraph(6)"
gap> D2 = EvalString(String(D2));
false
```

james-d-mitchell:
This will be resolved in v1.6.0 (or earlier).

PR #542, opened 2022-04-06
digraph: fix String method for chains/cycles

james-d-mitchell:
This PR resolves Issue #517, @wilfwilson is this the fix you envisaged?


wilfwilson (approved):
Yep! Nice

PR #543, opened 2022-04-07
Add DigraphRandomWalk

mtorpey:
This adds a new operation that executes a random walk on a graph, starting at a vertex, and having the desired length.

I haven't done doc or tests yet, but does this look like the right sort of thing?

Output is the same as DigraphPath: a pair `[v,a]` where
- `v` is the ordered list of vertices in the path;
- `a` is the list of indices of edges followed from each vertex (helps distinguish edges in multidigraphs).

```gap
gap> D := Digraph([[1, 2], [3], [2, 4], [1], [2, 4]]);                
<immutable digraph with 5 vertices, 8 edges>
gap> DigraphRandomWalk(D, 1, 4);
[ [ 1, 2, 3, 2, 3 ], [ 2, 1, 1, 1 ] ]
```

If we reach a sink, we cut the path short, so the final length might be less than requested.

### DIFF
--- a/doc/oper.xml
+++ b/doc/oper.xml
@@ -1535,6 +1535,36 @@ gap> DigraphShortestPath(D, 1, 1);
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="DigraphRandomWalk">
+<ManSection>
+  <Oper Name="DigraphRandomWalk" Arg="digraph, v, t"/>
+  <Returns>A pair of lists.</Returns>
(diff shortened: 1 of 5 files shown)


flsmith:
As discussed, Michael plans to rename this length to something less expectation-loaded which forces users to read the documentation, so that surprises from shorter-length paths being returned are minimised.

flsmith (approved):
Looks good!

PR #548, opened 2022-06-12
DigraphAbsorptionProbabilities

mtorpey:
This pull request adds the `DigraphAbsorptionProbabilities` attribute.  Here's an explanation:

> A random walk of infinite length through a digraph may pass through several different strongly connected components (SCCs). However, with probability 1 it must eventually reach an SCC which it can never leave, because the SCC has no out-edges leading to other SCCs. We may say that such an SCC has **absorbed** the random walk, and we can calculate the probability of each SCC absorbing a walk starting at a given vertex.
>
> DigraphAbsorptionProbabilities returns an $m \times n$ matrix `mat`, where $m$ is the number of vertices in `digraph` and $n$ is the number of strongly connected components. Each entry `mat[i][j]` is a rational number representing the probability that an unbounded random walk starting at vertex `i` will be absorbed by strongly connected component `j` – that is, the probability that it will reach the component and never leave.

The algorithm is based on established theory for [absorbing Markov chains](https://brilliant.org/wiki/absorbing-markov-chains/).  Once the strongly connected components are known, it pretty much comes down to one matrix multiplication.

This is the result of me looking into the game [Soccer Dice](https://boardgamegeek.com/boardgame/42785/soccer-dice), since I wanted to know the chance of a player scoring a goal starting from the black die.  The answer is $\frac{1020}{4493}$, as shown in the test file!


gap/attr.gi
```diff
@@ -196,6 +196,96 @@ function(I, subtracted_set, size_bound)
...
+    for v in comp do
+      if ForAny(OutNeighboursOfVertex(D, v), w -> not w in comp) then
+        is_sink_comp[i] := false;
+        Append(transient_vertices, comp);
+        break;
+      fi;
+    od;
+  od;
+  sink_comps := Positions(is_sink_comp, true);
+  nr_transients := Length(transient_vertices);
```
james-d-mitchell, line 233: You could shorten this section of code by just taking the quotient of the original digraph by the strongly connected components, and then running `DigraphSinks`, and using the labels of the quotient graph to recover the actual strongly connected components.

mtorpey: This is done in c28a75b3.

PR #549, opened 2022-06-25
Add Kamran to PackageInfo.g

mtorpey:
This is for his work on random digraphs that was recently merged (PR #531).

Is this the only place we add people?


james-d-mitchell:
Thanks @mtorpey, I think you need to add it to `doc/z-chap0.xml` also.

PR #555, opened 2022-08-08
Add a workaround for macOS code signing issues.

fingolfin:
Copying digraph.so over an existing digraph.so confuses macOS' code
signing system, leading to segfaults when trying to load that
digraph.so. As a workaround, delete the old digraph.so before copying
the new one into place.


james-d-mitchell:
Also amazing, thanks so much @fingolfin !

PR #563, opened 2022-12-05
Update CI

fingolfin:
... to get rid of pesky warnings in the GH actions webview


james-d-mitchell:
Thanks for this too! @fingolfin

PR #564, opened 2022-12-05
Change `CayleyDigraph` to use `AsSet`

fingolfin:
... instead of `AsList`, to produce a stable output that does not depend on the specific GAP version, random sources or anything else.

This should fix CI tests for latest GAP master, where the output of `AsList` for permutation groups changed: it used to match `AsSet` but doesn't anymore. See https://github.com/gap-system/gap/pull/5216 for details.

### DIFF
--- a/gap/grape.gi
+++ b/gap/grape.gi
@@ -104,7 +104,7 @@ function(G, gens)
   fi;
 
   # vertex i in the Cayley digraph corresponds to elts[i].
-  elts := AsList(G);
+  elts := AsSet(G);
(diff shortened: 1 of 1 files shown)


fingolfin:
@james-d-mitchell thanks for the quick merge. Would it be possible to get a new digraphs release with this and the previous minor tweaks? If it is made before December 14, then it'll get into GAP 4.12.2, too :-).

james-d-mitchell:
Sure, I can make a release today, after 3

PR #569, opened 2022-12-08
Fix some typos found by codespell

fingolfin:


james-d-mitchell:
Seems like some of the typos are in error messages which is why the test are failing, o/w this is fantastic, didn't know about codespell, but am very happy to have these changes. I'm not 100% sure it makes sense to correct spelling in the comments of the vendored bliss and planarity, any chance that those could be removed from the PR?

fingolfin:
I agree that nothing under `extern` should be modified, and will remove this -- I didn't pay enough attention, sorry.

One issue is that sometimes `homogeneous` and sometimes `homogenous` -- I thought both are correct and synonyms, but in truth, it seems that while both are valid words, they mean different things, at least according to https://writingexplained.org/homogenous-vs-homogeneous

I'll fix these to be consistent

PR #571, opened 2023-02-03
Avoid src/ prefix for GAP headers

fingolfin:
... so that future GAP versions can stop installing /usr/local/include/gap/src


fingolfin:
@james-d-mitchell this seems to be working fine now after I rebased. Not sure why CI failed before (random fluke perhaps?)

james-d-mitchell:
Thanks @fingolfin

PR #572, opened 2023-02-25
Use same URL as in my other packages

olexandr-konovalov:


james-d-mitchell:
Thanks @olexandr-konovalov

PR #579, opened 2023-04-21
Fix test files: strings and unbinds

mtorpey:
Nothing controversial here.

- Unbind variables correctly in all files (I've got a script that helped here)
- Fix bad message in DigraphsTestExtreme
- Correct the STOP_TEST messages in a couple of files where it said the wrong filename.


wilfwilson (approved):
Nice one! Thanks

PR #582, opened 2023-04-23
Standardise the way that `UndirectedSpanningForest` works in relation to mutability

wilfwilson:
The currently-existing methods for `UndirectedSpanningForest` in `attr.gi` date from quite a while ago, during a time when we dealt with mutability in a mix of different ways.

(`UndirectedSpanningForest` is a bit special, since it delegates to a C method that takes and returns a list of adjacencies - I think that's why it's been untouched for a while.)

In this PR I've made `UndirectedSpanningForest` a bit more normal, I think. The setup now has a similar structure to `UndirectedSpanningTree`, which comes just after it in the file. In particular, this lets us remove the use of a `SUM_FLAGS` and of having to explicitly do `SetUndirectedSpanningForestAttr`.

There is now the slightest bit of code duplication, but in my opinion this is a price worth paying for the increased clarity of the result.

### DIFF
--- a/gap/attr.gi
+++ b/gap/attr.gi
@@ -2250,35 +2250,37 @@ InstallMethod(DigraphReflexiveTransitiveReductionAttr,
 "for an immutable digraph",
 [IsImmutableDigraph], DigraphReflexiveTransitiveReduction);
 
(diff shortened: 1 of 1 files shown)


james-d-mitchell:
Thanks @wilfwilson good change!

PR #583, opened 2023-04-24
Add immediate methods for connected and strongly connected graphs

mtorpey:
A few observations that can get us certain properties for free:
- `IsStronglyConnectedDigraph` implies `IsConnectedDigraph`;
- One `DigraphStronglyConnectedComponent` implies `IsStronglyConnectedDigraph`;
- One `DigraphConnectedComponent` implies `IsConnectedDigraph`.

This PR adds a couple of quick immediate methods that give us these.

### DIFF
--- a/gap/prop.gi
+++ b/gap/prop.gi
@@ -137,6 +137,10 @@ InstallMethod(IsMeetSemilatticeDigraph, "for a digraph",
 [IsDigraph],
 D -> DIGRAPHS_IsMeetSemilatticeAndMeetTable(D)[1]);
 
+InstallImmediateMethod(IsStronglyConnectedDigraph,
+IsDigraph and HasDigraphStronglyConnectedComponents, 0,
+D -> Length(DigraphStronglyConnectedComponents(D).comps) = 1);
+
 InstallMethod(IsStronglyConnectedDigraph, "for a digraph by out-neighbours",
 [IsDigraphByOutNeighboursRep],
 D -> IS_STRONGLY_CONNECTED_DIGRAPH(OutNeighbours(D)));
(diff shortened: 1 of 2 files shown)


gap/prop.gi
```diff
@@ -208,6 +212,12 @@ function(D)
   return true;
 end);
 
+InstallTrueMethod(IsConnectedDigraph, IsStronglyConnectedDigraph);
```
wilfwilson, line 215: We tend to put true methods in the `.gd` file. In particular, we already have this true method in `prop.gd`, so please don't duplicate it here:

https://github.com/digraphs/Digraphs/blob/33c3b9c6197c6dd6c33e04e66fdd4bff7d85c987/gap/prop.gd#L84

gap/prop.gi
```diff
@@ -137,6 +137,10 @@ InstallMethod(IsMeetSemilatticeDigraph, "for a digraph",
 [IsDigraph],
 D -> DIGRAPHS_IsMeetSemilatticeAndMeetTable(D)[1]);
 
+InstallImmediateMethod(IsStronglyConnectedDigraph,
+HasDigraphStronglyConnectedComponents, 0,
```
wilfwilson, line 141: A different, or additional, approach could be to set the value of this attribute whenever `DigraphStronglyConnectedComponents`, respectively `DigraphConnectedComponents`, is computed. But I'm happy with immediate methods, and I'm happy to leave this as it is.

This might be something from a cargo cult, but I have a vague memory of it being important make it ultra explicit that a method that only applies to, say, digraphs should explicitly say that it requires a digraph. In that vain, I suggest changing this line to:
```suggestion
IsDigraph and HasDigraphStronglyConnectedComponents, 0,
```
and changing the other such line in this PR correspondingly.

I don't think there's any harm in doing it this way anyway.

gap/prop.gi
mtorpey: Nice!  Will remove.

gap/prop.gi
mtorpey: Good enough for me.  Will change!

mtorpey:
Cheers for the suggestions.  Fixed and rebased!

james-d-mitchell:
Nice one thanks @wilfwilson and @mtorpey !

PR #585, opened 2023-05-10
Add AbsorptionExpectedSteps

mtorpey:
As the doc says:

    A random walk of infinite length through a finite digraph may pass through
    several different strongly connected components (SCCs).  However, with
    probability 1 it must eventually reach an SCC which it can never leave,
    because the SCC has no out-edges leading to other SCCs.  When this happens,
    we say the walk has been <E>absorbed</E>.
    <P/>

    <C>DigraphAbsorptionExpectedSteps</C> returns a list <C>L</C> with length
    equal to the number of vertices of <A>digraph</A>, where <C>L[v]</C> is the
    expected number of steps a random walk should take before it is absorbed.  –
    that is, the expected number of steps to reach an SCC that it can never
    leave.

This is the sister function to `DigraphAbsorptionProbabilities` and uses a lot of shared code, which I've abstracted into a new `DIGRAPHS_AbsorbingMarkovChain`; this could be used for even more things later.


james-d-mitchell (commented):
Generally looks good. Three things:

1. ~~I'm pretty sure that the absorbing scc is always the last one on the output of strongly connected components `comp`~~ @mtorpey  rightly points out that there can be several absorbing sccs and so this comment is nonsense.
2. Can you just double check (and maybe add a test) that this works correctly when the input digraph is mutable? 
3. Can you please regenerate the unbind local variables statements at the end of the test file using the script in `etc`?

mtorpey:
1. Ignoring.
2. Added test in the last commit.
3. Tried this, but no useful changes.

I also rebased.

PR #601, opened 2024-01-05
Apply some more linting rules

james-d-mitchell:


gap/io.gi
```diff
@@ -908,24 +909,23 @@ function(func, s)
...
   local D;
   D := MakeImmutable(DigraphFromSparse6String(IsMutableDigraph, s));
   SetIsSymmetricDigraph(D, true);
   return D;
 end);
 
 InstallMethod(DigraphFromSparse6String, "for a function and a string",
-[IsFunction, IsString],
-DigraphFromSparse6StringCons);
+[IsFunction, IsString], DigraphFromSparse6StringCons);
```
mtorpey, line 928: Why?

gap/isomorph.gi
```diff
@@ -292,23 +284,13 @@ function(D)
...
 
 InstallMethod(BlissAutomorphismGroup, "for a digraph and vertex coloring",
-[IsDigraph, IsHomogeneousList],
-function(D, colors)
-  return BLISS_DATA(D,
-                    colors,
-                    fail,
-                    "AutomorphismGroup")[1];
-end);
+[IsDigraph, IsHomogeneousList], {D, colors} -> BLISS_DATA(D, colors, fail)[1]);
```
mtorpey, line 287: We definitely don't want the last argument in these?

mtorpey (approved):

gap/io.gi
james-d-mitchell: As I think Reimer Behrens once said: "vertical screen real estate is at a premium", so one less line is a good thing from this perspective.

gap/isomorph.gi
james-d-mitchell: Definitely don't, I think this was a remnant of some earlier code.

PR #609, opened 2024-01-31
[Draft] Implement IsOrderIdeal

DanielPointon:


doc/oper.xml
```diff
@@ -1395,6 +1395,38 @@ gap> VerticesReachableFrom(D, 5);
 ]]></Example>
   </Description>
 </ManSection>
+
+<ManSection>
+  <Oper Name="VerticesReachableFrom" Arg="digraph, root"/>
```
james-d-mitchell, line 1400: I think it'd be better if the 2nd argument was called `list`.

doc/oper.xml
```diff
@@ -1395,6 +1395,38 @@ gap> VerticesReachableFrom(D, 5);
...
+  <Returns>A list.</Returns>
+  <Description>
+    This operation returns a list of the vertices  <A>v</A>, for which there exists
+    a non-trivial directed walk from any of the provided <A>root</A>s to vertex <A>v</A> in the digraph
```
james-d-mitchell, line 1404: `root` is a list of vertices, so this doesn't really make sense as written, I'm afraid.

doc/oper.xml
```diff
@@ -1395,6 +1395,38 @@ gap> VerticesReachableFrom(D, 5);
...
+    <A>digraph</A>. See Section <Ref Subsect="Definitions" Style="Number" />
+    for the definition of a non-trivial directed walk.
+    <P/>
```
james-d-mitchell, line 1407: The manual entry should include some indication of what happens if the list of roots contains values that are not vertices of `digraph`.

doc/oper.xml
james-d-mitchell: And the first argument be called `D` for consistency with the other entry you're adding here.

doc/oper.xml
```diff
@@ -2257,3 +2289,30 @@ true
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="IsOrderIdeal">
+<ManSection>
+  <Oper Name="IsOrderIdeal" Arg="D, subset" Label="for a digraph and list"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    This function returns <K>true</K> if the specified subset is "downwards" closed, i.e. contains every node less than the given nodes.
```
james-d-mitchell, line 2298: The manual entry should indicate that `D` should satisfy `IsPartialOrderDigraph` and say what happens if `D` does not satisfy `IsPartialOrderDigraph`.

doc/oper.xml
```diff
@@ -2257,3 +2289,30 @@ true
...
+
+    <Example><![CDATA[
```
james-d-mitchell, line 2300: For consistency, can you please rename all the variables in this example to `D` instead of `TestPartialOrderDigraph2` etc?

gap/oper.gi
```diff
@@ -2043,6 +2043,59 @@ function(D, root)
...
+      fi;
+    od;
+
+    visited_as_ints := [];
+
+    for i in [1 .. N] do
+      if visited[i] then;
+        Add(visited_as_ints, i);
+      fi;
+    od;
```
james-d-mitchell, line 2069:
```suggestion
    visited_as_ints := ListBlist([1 .. N], visited);
```

gap/oper.gi
```diff
@@ -2043,6 +2043,59 @@ function(D, root)
...
 end);
 
+InstallMethod(VerticesReachableFrom, "for a digraph and a list of vertices",
+[IsDigraph, IsList],
+function(D, roots)
+    local accessible_from_node, candidate, root, i, visited, visited_as_ints, N;
+    N := DigraphNrVertices(D);
+    visited := BlistList([1 .. N], []);
+    for root in roots do
+      if not(visited[root]) then
```
james-d-mitchell, line 2053:
```suggestion
      if not visited[root] then
```
Please remove these extraneous brackets.

gap/oper.gi
```diff
@@ -2043,6 +2043,59 @@ function(D, root)
...
+        accessible_from_node := VerticesReachableFrom(D, root);
+        for candidate in accessible_from_node do
+          if not(visited[candidate]) then
```
james-d-mitchell, line 2056:
```suggestion
          if not visited[candidate] then
```

gap/oper.gi
```diff
@@ -2043,6 +2043,59 @@ function(D, root)
...
+
+    return visited_as_ints;
+end);
+
+InstallMethod(IsOrderIdeal, "for a digraph and a list of vertices",
+[IsDigraph, IsList],
+# Check if digraph represents a partial order
+function(D, roots)
+  local reachable_vertices, vertex_in_subset, i, N;
+  if not(IsPartialOrderDigraph(D)) then
```
james-d-mitchell, line 2079:
```suggestion
  if not IsPartialOrderDigraph(D) then
```

gap/oper.gi
```diff
@@ -2043,6 +2043,59 @@ function(D, root)
...
+    ErrorNoReturn("the 1st element (a digraph) must be a partial order digraph");
```
james-d-mitchell, line 2080: Please add a test that covers these lines, you can just a the gap input to the test file:
~~~
gap> IsOrderIdeal(D, [1, 2, 3]); # where D is not a partial order 
~~~
and then do 
~~~
Test("tst/standard/oper.tst", rec(rewriteToFile := true));
~~~
and GAP will add the expected output to the test file for you.

james-d-mitchell (changes requested):
@DanielPointon This looks really very good for a first PR, there are a number of minor comments that I've pointed out in comments. I also have one more major comment: I think the complexity of the `VerticesReachableFrom` method that you've implemented is unreasonably high (big-O of the number of vertices + the number of edges squared,
think of the example in the image attached where the roots are `[6 .. 10]` rather than number of vertices + number of edges that could be achieved as follows). I suggest replacing your method for `VerticesReachableFrom(D, list)` with an adjusted version of the method for

 `VerticesReachableFrom(D, vertex)`, and then replacing the method for the latter with a call to `VerticesReachableFrom(D, [vertex]);` this will have better complexity, and reduce duplication of the code. 

[vizpicture.pdf](https://github.com/digraphs/Digraphs/files/14122833/vizpicture.pdf)

gap/oper.gi
DanielPointon: Resolved - thanks!

gap/oper.gi
DanielPointon: Cool - thanks!

gap/oper.gi
DanielPointon: Neat - thanks!

doc/oper.xml
DanielPointon: Good point - thanks! Resolved

doc/oper.xml
DanielPointon: Resolved

doc/oper.xml
```diff
@@ -1395,6 +1395,38 @@ gap> VerticesReachableFrom(D, 5);
 ]]></Example>
   </Description>
 </ManSection>
+
+<ManSection>
+  <Oper Name="VerticesReachableFrom" Arg="D, list"/>
+  <Returns>A list.</Returns>
+  <Description>
+    This operation returns a list of the vertices  <A>v</A>, for which there exists
```
james-d-mitchell, line 1403:
```suggestion
    This operation returns a list of the vertices  <C>v</C>, for which there exists
```

`A` is for "argument"

doc/oper.xml
```diff
@@ -1395,6 +1395,38 @@ gap> VerticesReachableFrom(D, 5);
...
+    a non-trivial directed walk from any of the elements of the provided list to vertex <A>v</A> in the digraph
```
james-d-mitchell, line 1404:
```suggestion
    a non-trivial directed walk from any of the vertices in <A>list</A> to vertex <C>v</C> in the digraph
```

doc/oper.xml
```diff
@@ -2257,3 +2289,30 @@ true
...
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="IsOrderIdeal">
+<ManSection>
+  <Oper Name="IsOrderIdeal" Arg="D, subset" Label="for a digraph and list"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    This function returns <K>true</K> if the specified subset is "downwards" closed, i.e. contains every node less than the given nodes. 
+    The function can only be used on digraphs satisfying <K>IsPartialOrderDigraph</K> and will throw an error if passed a digraph that is not a partial order digraph.
```
james-d-mitchell, line 2299:
```suggestion
    The function can only be used on digraphs satisfying <Ref Prop="IsPartialOrderDigraph"/> and will throw an error if passed a digraph that is not a partial order digraph.
```
`K` is for keyword

doc/oper.xml
```diff
@@ -2257,3 +2289,30 @@ true
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="IsOrderIdeal">
+<ManSection>
+  <Oper Name="IsOrderIdeal" Arg="D, subset" Label="for a digraph and list"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    This function returns <K>true</K> if the specified subset is "downwards" closed, i.e. contains every node less than the given nodes.
```
james-d-mitchell, line 2298:
```suggestion
    This function returns <K>true</K> if the specified subset is "downwards" closed, i.e. contains every vertex less than the given vertices. 
```

doc/oper.xml
```diff
@@ -1395,6 +1395,38 @@ gap> VerticesReachableFrom(D, 5);
...
   </Description>
 </ManSection>
+
+<ManSection>
+  <Oper Name="VerticesReachableFrom" Arg="D, list"/>
+  <Returns>A list.</Returns>
+  <Description>
+    This operation returns a list of the vertices  <A>v</A>, for which there exists
+    a non-trivial directed walk from any of the elements of the provided list to vertex <A>v</A> in the digraph
+    <A>digraph</A>. See Section <Ref Subsect="Definitions" Style="Number" />
```
james-d-mitchell, line 1405:
```suggestion
    <A>D</A>. See Section <Ref Subsect="Definitions" Style="Number" />
```

gap/oper.gi
```diff
@@ -2005,42 +2005,82 @@ end);
...
+[IsDigraph, IsList],
+function(D, roots)
+  local N, index, visited, queue_tail, queue,
+  root, element, neighbour, graph_out_neighbors, node_neighbours;
+
+  N := DigraphNrVertices(D);
+  visited := BlistList([1 .. N], []);
+
+  for root in roots do
+    if 0 = root or root > N then
```
james-d-mitchell, line 2029:
```suggestion
    if not IsPosInt(N) or 0 = root or root > N then
```

gap/oper.gi
```diff
@@ -2005,42 +2005,82 @@ end);
...
+                    "argument (a digraph)");
+    fi;
+  od;
+
+  graph_out_neighbors := OutNeighbors(D);
+  queue := EmptyPlist(N);
+  queue_tail := 0;
+  for root in roots do
+    queue_tail := queue_tail + 1;
+    queue[queue_tail] := root;
```
james-d-mitchell, line 2041: I'm not sure I see why you are doing this here, why not just do `queue := roots`? Or are you trying to accommodate the situation when `roots` has holes (i.e. `roots = [1,, 2]`)? If the latter, then I think the better approach would be to just require that `roots` satisfies `IsDenseList` and to error out if it does not.

gap/oper.gi
```diff
@@ -2005,42 +2005,82 @@ end);
...
+end);
+
+InstallMethod(VerticesReachableFrom, "for a digraph and a list of vertices",
+[IsDigraph, IsList],
+function(D, roots)
+  local N, index, visited, queue_tail, queue,
+  root, element, neighbour, graph_out_neighbors, node_neighbours;
+
+  N := DigraphNrVertices(D);
+  visited := BlistList([1 .. N], []);
```
james-d-mitchell, line 2026: This line should be moved until after the argument checking, o/w it is created for nothing.

gap/oper.gi
```diff
@@ -2005,42 +2005,82 @@ end);
...
+  return ListBlist([1 .. N], visited);
+end);
+
+InstallMethod(IsOrderIdeal, "for a digraph and a list of vertices",
+[IsDigraph, IsList],
+# Check if digraph represents a partial order
+function(D, roots)
+  local reachable_vertices, vertex_in_subset, i, N;
+  if not IsPartialOrderDigraph(D) then
+    ErrorNoReturn("the 1st element (a digraph) must be a partial order digraph");
```
james-d-mitchell, line 2067:
```suggestion
    ErrorNoReturn("the 1st argument (a digraph) must be a partial order digraph");
```

gap/oper.gi
```diff
@@ -2005,42 +2005,82 @@ end);
...
+  fi;
+
+  N := Length(roots);
+  vertex_in_subset := BlistList([1 .. N], []);
+  reachable_vertices := VerticesReachableFrom(D, roots);
+  for i in roots do
+    vertex_in_subset[i] := true;
+  od;
```
james-d-mitchell, line 2075:
```suggestion
  vertex_in_subset := BlistList([1 .. N], roots);
  reachable_vertices := VerticesReachableFrom(D, roots);
```

james-d-mitchell: Or you could replace the entirety of the end of this function by:
~~~
return Length(reachable_vertices) = Length(roots) and not ForAny(reachable_vertices, x -> vertex_in_subset[x]);
~~~

james-d-mitchell (changes requested):
Looks good! Only minor quibbles left, please resolve these, then we can merge this I think.

gap/oper.gi
DanielPointon: The idea here was for efficiency, queue := roots would create a list that would need resizing, which would (I think) have some time-cost associated with resizing the list, creating the list like this means we don't have to resize the array every time which I thought would save time, correct me if I'm wrong :)

DanielPointon:
@james-d-mitchell this is ready! (I think)

doc/oper.xml
```diff
@@ -2257,3 +2268,30 @@ true
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="IsOrderIdeal">
+<ManSection>
+  <Oper Name="IsOrderIdeal" Arg="D, subset" Label="for a digraph and list"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    This function returns <K>true</K> if the specified subset is "downwards" closed, i.e. contains every vertex less than the given vertices.
```
james-d-mitchell, line 2277:
```suggestion
    This function returns <K>true</K> if the specified subset is "downwards" closed, i.e. contains every vertex less than the given vertices in the order defined by <A>D</A>. 
```

james-d-mitchell (dismissed):

PR #611, opened 2024-02-05
Print empty lists with two spaces for consistency

zickgraf:
as a preparation for https://github.com/gap-system/gap/pull/5418


james-d-mitchell:
Thanks for the PR @zickgraf, I'm happy enough to merge this but won't until https://github.com/gap-system/gap/pull/5418 is merged, so that we don't break the development version of GAP.

PR #618, opened 2024-02-21
Add DigraphContractEdge

saffronmciver:
Initial implementation of ContractEdge (see [Contract Edge Wikipedia Article](https://en.wikipedia.org/wiki/Edge_contraction); as discussed, this implementation should not create multiple edges). Mutable and Immutable methods exist for this. 

Edge labels should be handled if `DigraphRemoveEdge` were removing them, but this has not been tested yet because of this issue. 

Contract edge takes a non looping edge `(u, v)`, and contracts this into a single vertex `w`. The edge is removed, vertex labels are combined (in a list), and incident edges to `(u, v)` become incident to `w`. `MultiDigraphs` are not accepted into this function. If an edge `(v, u)` exists, this is also contracted (removed). 

Documentation has not been written yet (this is in progress).

### DIFF
--- a/gap/oper.gi
+++ b/gap/oper.gi
@@ -473,7 +473,7 @@ function(D, u, v)
   # add vertex w with combined labels from u and v
   NewDigraph := DigraphAddVertex(NewDigraph, [DigraphVertexLabel(D, u), DigraphVertexLabel(D, v)]); # add vertex w
 
-  vertices := DigraphVertices(D);
+  vertices := DigraphVertices(NewDigraph);
   w := vertices[Length(vertices)]; # w is the new vertix identifier 
 
   # Handle loops from edges u or w, with the same source / range
--- a/gap/oper.gi
+++ b/gap/oper.gi
@@ -377,67 +377,74 @@ InstallMethod(DigraphClosure,
 [IsImmutableDigraph, IsPosInt],
 {D, k} -> MakeImmutable(DigraphClosure(DigraphMutableCopy(D), k)));
 
+DIGRAPHS_CheckContractEdgeDigraph := function(D, u, v)
+  if not IsDigraphEdge(D, u, v) then  # Check if [u, v] is an edge in digraph D
+    ErrorNoReturn("expected an edge between the 2nd and 3rd arguments ",
+                  "(vertices) ", u, " and ", v, " but found none");
+  elif IsMultiDigraph(D) then
+    ErrorNoReturn("The 1st argument (a digraph) must not satisfy ",
+                   "IsMultiDigraph");
+  elif u = v then  # edge should not be contracted if u = v
+    ErrorNoReturn("The 2nd argument <u> must not be equal to the 3rd ",
+                   "argument <v>");
+  fi;
+
+end;
+
 InstallMethod(DigraphContractEdge,
 "for a mutable digraph and two positive integers",
 [IsMutableDigraph, IsPosInt, IsPosInt],
 function(D, u, v)
   local w, neighbours, transformations_to_edge, transformations_to_old_edges,
-  old_edge, new_edge, neighbour, t, copy, vertices;
+  old_edge, new_edge, neighbour, t, vertices, edge_label;
 
-  if not IsDigraphEdge(D, u, v) then  # Check if [u, v] is an edge in digraph D
-    ErrorNoReturn("u, v is not an edge of D");
-  fi;
+  DIGRAPHS_CheckContractEdgeDigraph(D, u, v);
 
-  if IsMultiDigraph(D) then  # multi digraphs cannot have their edges contracted
-    ErrorNoReturn("Cannot contract an edge for a MultiDigraph");
-  fi;
-
-  if u = v then  # edge should not be contracted if u = v
-    ErrorNoReturn("Cannot contract an edge with the same source and range");
-  fi;
-
-  if IsDigraphEdge(D, v, u) then  # if (v, u) is an edge,
-    DigraphRemoveEdge(D, v, u);   # disallow loops, so also remove (v, u)
+  # if (v, u) is an edge, disallow loops, so remove (v, u)
+  if IsDigraphEdge(D, v, u) then
+    DigraphRemoveEdge(D, v, u);
   fi;
 
   # remove the contracted edge
   DigraphRemoveEdge(D, u, v);
 
-  copy := DigraphImmutableCopy(D);  # create a copy of the original digraph,
-                                    # with removed edges, to use as a reference
-                                    # for edge labels
+  # Find vertex neighbours of u and v to construct new incident edges of w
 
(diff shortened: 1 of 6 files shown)


gap/oper.gi
```diff
@@ -409,7 +407,8 @@ function(D, u, v)
   
   DigraphAddVertex(D, [DigraphVertexLabel(D, u), DigraphVertexLabel(D, v)]); # add vertex w
 
-  w := Last(DigraphVertices(D)); # w is the new vertix identifier 
+  vertices := DigraphVertices(D);
+  w := vertices[Length(vertices)]; # w is the new vertix identifier
```
james-d-mitchell, line 411:
```suggestion
  w := vertices[Length(vertices)]; # w is the new vertex identifier 
```

gap/oper.gi
```diff
@@ -474,7 +473,8 @@ function(D, u, v)
   # add vertex w with combined labels from u and v
   NewDigraph := DigraphAddVertex(NewDigraph, [DigraphVertexLabel(D, u), DigraphVertexLabel(D, v)]); # add vertex w
 
-  w := Last(DigraphVertices(NewDigraph)); # w is the new vertix identifier 
+  vertices := DigraphVertices(D);
+  w := vertices[Length(vertices)]; # w is the new vertix identifier
```
james-d-mitchell, line 477:
```suggestion
  w := vertices[Length(vertices)]; # w is the new vertex identifier 
```

gap/oper.gi
```diff
@@ -530,6 +530,16 @@ function(D, u, v)
 
 end);
 
+InstallMethod(DigraphContractEdge,
+"for a digraph and a list",
+[IsDigraph, IsList],
+function(D, edge)
+  if Length(edge) <> 2 then
```
james-d-mitchell, line 537: Shouldn't this also check `IsDenseList`?  In case, the second entry of edge is bound but the first not.

james-d-mitchell:
@saffronmciver is this ready for review?

mtorpey (changes requested):
Overall thoughts:
- `DigraphContractEdge` should always return a digraph, even if it's using mutability and just returning a pointer to the original thing.
- We should avoid copying digraphs in the mutable case.
- I see some code repeated in the mutable and immutable methods, e.g. the checks at the start.  Feel free to make another function that these can both call.  You can do this with the simple syntax `DIGRAPHS_SomeFunctionName := function(...) ... end;`
- The immutable method has a pattern where you make a graph and then add and remove things one at a time with `AddEdge` and so on.  It'd be better to do this with an immutable graph, and then do `MakeImmutable` just before returning, to avoid copying graphs again and again.

gap/oper.gi
```diff
@@ -377,6 +377,237 @@ InstallMethod(DigraphClosure,
...
 
+InstallMethod(DigraphContractEdge,
+"for a mutable digraph and two positive integers",
+[IsMutableDigraph, IsPosInt, IsPosInt],
+function(D, u, v)
+  local w, neighbours, transformations_to_edge, transformations_to_old_edges,
+  old_edge, new_edge, neighbour, t, copy, vertices;
+
+  if not IsDigraphEdge(D, u, v) then  # Check if [u, v] is an edge in digraph D
+    ErrorNoReturn("u, v is not an edge of D");
```
james-d-mitchell, line 388:
```suggestion
    ErrorNoReturn("expected an edge between the 2nd and 3rd arguments (vertices) ", u, " and ", v, " but found none");
```

gap/oper.gi
```diff
@@ -377,6 +377,237 @@ InstallMethod(DigraphClosure,
...
+  fi;
+
+  if IsMultiDigraph(D) then  # multi digraphs cannot have their edges contracted
```
james-d-mitchell, line 391: These checks should be combined into a single `if A then Error; elif B then Error;...` if you see what I mean

gap/oper.gi
```diff
@@ -377,6 +377,237 @@ InstallMethod(DigraphClosure,
...
+    ErrorNoReturn("Cannot contract an edge for a MultiDigraph");
```
james-d-mitchell, line 392: Also the error messages should follow the same format as everywhere else in the package:

> the 1st argument (a digraph) must not have multiple edges

gap/oper.gi
```diff
@@ -377,6 +377,237 @@ InstallMethod(DigraphClosure,
...
+  fi;
+
+  if u = v then  # edge should not be contracted if u = v
+    ErrorNoReturn("Cannot contract an edge with the same source and range");
```
james-d-mitchell, line 396: Similarly this error message should be revised.

gap/oper.gi
```diff
@@ -377,6 +377,237 @@ InstallMethod(DigraphClosure,
...
+  fi;
+
+  edges_to_not_include := [];  # Incident edges to [u, v] that will be
+                               # transformed to be incident to w
+
+  existing_edges := [];  # existing edges that should be re added
+
+  NewDigraph := DigraphByEdges([]);  # contracted edge should not be added to
+                                     # the new digraph
```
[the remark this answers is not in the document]
james-d-mitchell: Or even `NullDigraph(0);`!

gap/oper.gi
```diff
@@ -377,6 +377,237 @@ InstallMethod(DigraphClosure,
...
+  fi;
+
+  if IsDigraphEdge(D, v, u) then  # if (v, u) is an edge,
+    DigraphRemoveEdge(D, v, u);   # disallow loops, so also remove (v, u)
+  fi;
+
+  # remove the contracted edge
+  DigraphRemoveEdge(D, u, v);
+
+  copy := DigraphImmutableCopy(D);  # create a copy of the original digraph,
```
james-d-mitchell, line 406: If `D` doesn't have edge or vertex labels, then this copy is unnecessary, we should just return `D` right away here without doing anything extra. 

Also: even if `D` does have edge and/or vertex labels, then we don't really require the copy of `D` but only of its edge and/or vertex labels, and even then probably only the labels of the vertices or edge incident to the edge being collapsed. I think this should be updated to copy only the very least amount of data from `D` required.

james-d-mitchell (changes requested):
This looks pretty good for a first PR, thanks! Handling the edge/node labels is much more complicated than I expected, looks like you've done a pretty thorough job. Please make the changes indicated by @mtorpey and me, and then we'll have another look!

james-d-mitchell:
I've just rebased this onto main which will hopefully fix the CI.

james-d-mitchell:
Closing and re-opening to trigger ci

mtorpey (approved):
This is a big improvement, and I think it's now ready for merging.  All my comments have been addressed, and it looks to me like all of @james-d-mitchell's too.

Recommend merging.

james-d-mitchell:
Thanks @saffronmciver !

PR #622, opened 2024-03-05
Use the new graphviz package

mpan322:
## Notes:
 - I tried to limit the number of changes I made to the base code as much as possible (no need to add bugs if possible).
 - This likely results in some redundant code (TODO).
 - Tests for the code have been updated, but they have been checked that they are still testing the same examples.
      - _Minor Note_ Previously graphs in the new code were laid out differently than graphs in the old code.  Following recent changes to graphviz (adding an ordering in output based on the order they were added to the object) this is no longer the case. They now match up directly. 
 - When writing these I added `GV_<old name>` methods for testing. Let me know if you would like these removed.
      - There should be a `<old name>` equivalent of `GV_<old name>` which do the exact same thing.

### DIFF
--- a/.github/workflows/gap.yml
+++ b/.github/workflows/gap.yml
@@ -28,6 +28,7 @@ jobs:
           - stable-4.12
         pkgs-to-clone:
           - NautyTracesInterface
+          - mpan322/graphviz
         ABI: ['']
 
         include:
--- a/PackageInfo.g
+++ b/PackageInfo.g
@@ -390,6 +390,7 @@ Dependencies := rec(
   GAP := ">=4.10.0",
   NeededOtherPackages := [["io", ">=4.5.1"],
                           ["orb", ">=4.8.2"],
+                          ["graphviz", "*"],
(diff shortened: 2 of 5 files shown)


james-d-mitchell:
Nice! @mpan322 the ci is failing because it doesn't know how to install the graphviz package. That should be easy to fix (in the morning!)

james-d-mitchell:
@mpan322 it looks like the ci now does (mostly) install the graphviz package, but there are a few issues:

1. The linting fails for the files you've modified in the Digraphs package
2. It seems that graphviz uses `Pluralize` which was introduced in GAP 4.12, so we should probably not use `Pluralize` in graphviz, and ensure that the graphviz package requires GAP 4.10 or higher (I'm not sure what it requires ATM but it must not require GAP 4.12 because it wouldn't have loaded in the CI jobs using earlier versions of GAP). 

There might be other issues with graphviz not being installed in some of the CI jobs, I can fix those, probably best to fix 1 and 2 above first though. Thanks again!

PR #627, opened 2024-03-14
Abort if malloc fails

DanielPointon:


james-d-mitchell:
Just to say that I've rebased this onto the current `main` which should fix the ci.

james-d-mitchell (approved):
I made a few minor changes to this, I don't think it's possible to test these functions at present (i.e. to test that an error is given if c/malloc fails) but I'm happy to merge this if the CI passes.

PR #635, opened 2024-04-01
cliques: some perf improvements

james-d-mitchell:
This commit includes some performance improvements for the clique finding code.  The example in the newly added extreme/cliques.tst file takes approx. 4s with the modified code, versus at least 3 minutes previously (I never ran it to completion previously because it was taking too long). If we do:

~~~
List(DigraphCliquesReps(D),
     x -> Orbit(AutomorphismGroup(D), x, OnSets));
~~~
for the same example, then this takes about 12s, doing the same with
~~~
List(DigraphCliquesReps(D),
     x -> Orb(AutomorphismGroup(D), x, OnSets));
~~~
instead runs out of memory repeatedly (and takes longer than 12s). Note that the automorphism group is of size 16, and so the orbits are relatively small in comparison to the number of cliques.


james-d-mitchell:
TODO: this means we don't need Orb for Digraphs any more.

reiniscirpons (dismissed):
PR looks good, thanks James! Can confirm that the extreme test runs out of memory (and takes way longer) without these changes, much better now.

gap/cliques.gi
```diff
@@ -519,16 +539,13 @@ function(arg)
...
-      out := [];
+      orbits := HashMap();
       for c in cliques do
-        if not ForAny(orbits, x -> c in x) then
-          orb := Orb(G, c, OnSets);
-          Enumerate(orb);
-          Add(orbits, orb);
-          Append(out, orb);
+        if not c in orbits then
+          MultiOrbit(G, c, orbits);
```
reiniscirpons, line 545: This line is a bit confusing, since `MultiOrbit` returns an orbit and its non-obvious that it changes the `orbits` hashmap in place (this is what its used here). Would it make sense to just inline the bit of the `MultiOrbit` code thats relevant to updating the `orbits` hashmap?

james-d-mitchell: I'm not sure it makes sense to inline the bit of code in `MultiOrbit`, it is used elsewhere, and then we'd have duplication of the method. Maybe I can rename the function to `AddOrbitToMap` or something?

PR #638, opened 2024-04-05
Html manual images

### LINKED ISSUE #637: Images in documentation
It would be nice to figure out how to include images in the documentation. This would be useful, for example, in displaying equations.

james-d-mitchell:
This PR resolves #637, I think.


doc/oper.xml
```diff
@@ -1352,8 +1352,8 @@ false
 
 <#GAPDoc Label="VerticesReachableFrom">
 <ManSection>
-  <Oper Name="VerticesReachableFrom" Arg="digraph, root"/>
-  <Oper Name="VerticesReachableFrom" Arg="digraph, list"/>
+  <Oper Name="VerticesReachableFrom" Arg="digraph, root" Label="for a vertex"/>
+  <Oper Name="VerticesReachableFrom" Arg="digraph, list" Label="for a list of vertices"/>
```
mtorpey, line 1356: Should the labels include "a digraph and"?

doc/prop.xml
```diff
@@ -1307,14 +1307,27 @@ true
...
+    <Alt Only="HTML">
+        <![CDATA[
+        <p>
+        <center>
+            <img height="200" src="m3.png"/>
+            &nbsp;&nbsp;&nbsp;&nbsp;
+            <img height="200" src="n5.png"/>
+            <figcaption>The lattices <e>M3</e> and <e>N5</e>.</figcaption>
+        </center>
+        </p>
```
mtorpey, line 1327: The HTML standard says that `figcaption` can only appear as the child of a `figure` element, so I guess this is invalid.  Should we use `<figure>` here instead of `<p><center>` ?  (It's more semantic anyway!)

mtorpey (commented):
The images appear nicely in HTML, so this is good!  Just a couple of comments.

doc/oper.xml
james-d-mitchell: Good catch, updating this now.

doc/prop.xml
james-d-mitchell: Great point! I've updated this to use figure, plus center, and the output looks good I think:
Screenshot 2024-04-17 at 09 12 59

james-d-mitchell:
@MTWhyte what says ye?

doc/prop.xml
mtorpey: Strictly this is still invalid HTML, since our `figcaption` is a (direct) child of a `center` element and not a `figure` element.  I can definitely see we want them centered though.

The "proper" solution would be to add
```css
figure {
    text-align: center;
}
```
to `manual.css`, but that's a matter for GAPDoc.

For now, your solution seems the best one.

mtorpey: One last thing though: do we want a `p` element inside `figcaption`?  I think it's standard just to put text directly into `figcaption`, unless there's some reason that's not working.

james-d-mitchell: Thanks @mtorpey. For the first comment, if I center then figure, this will resolve that issue, right?  Or I guess we can adjust the css in doc/ to include your proper solution? 

About the <p> in <figcaption> I inserted that because o/w the caption was very close to the figures, and with it, there is a reasonable vertical space. If there's a proper solution for that I'll be happy to learn it!

mtorpey: For both of these, the proper fix is definitely with css.  The figcaption spacing issue would be solved with
```css
figcaption {
    margin: 1em 0;
}
```
but from what I can tell GAPDoc produces the css file when we compile the doc – unless I'm missing something, we can't specify our own without changing GAPDoc itself.

If we can't change the CSS ourselves, then I like your `<center><figure>` solution, and `<figcaption><p>` seems a reasonable workaround – I believe it's at least legal.

james-d-mitchell: Good point! It seems that we can add our own style sheets, from the GAPDoc manual:

> The generated HTML files refer to the file manual.css which conforms to the W3C specification CSS 2.0, see https://www.w3.org/TR/REC-CSS2, and the javascript file manual.js (only in browsers which support CSS or javascript, respectively; but the HTML files are also readable without any of them). To add a style mystyle one or both of mystyle.css and mystyle.js must be provided; these can overwrite default settings and add new javascript functions. For more details see the comments in manual.js.

It's not completely clear to me how we do this, any thoughts @mtorpey ?

mtorpey: I've just been reading this myself.  It looks like we can maybe add another file called `mystyle.css` and the Javascript will import it automatically.  Might be a bit complicated though – maybe we should merge this as-is and make a separate issue?

MTWhyte (commented):
Looks good (I seem to have just missed the review request notification!)

PR #639, opened 2024-04-10
Converting the displaying code of Digraphs from the old version to the graphviz version

mpan322:
- Updated the old displaying code of the digraphs package
 - Added new equivalents which output graphviz objects
 - Updated existing test code so new version passes (validated maually that they are testing the same thing).


james-d-mitchell:
Superseded by #644

PR #643, opened 2024-05-03
Update for gaplint 1.2.0

james-d-mitchell:


james-d-mitchell:
Now the gap-4.10 job is failing because of changes to datastructures...

PR #650, opened 2024-05-22
Edge-weights #3: minimum spanning trees

mtorpey:


james-d-mitchell (changes requested):
Quick comments that occur to me on quick inspection:

* there seem to be some white space only changes in these changes, might be easier to review if those were reverted or separated into a separate pr
* there's an xml entity that should be included in the doc for attributes and properties that says something to the effect of "for mutable digraphs this is recomputed every time the function is called"
* I noticed that there's a union find implementation in the pr, but this already exists in the data structures package which I think we already require, be better to use that than roll our own again.

Otherwise looks good! On my phone so excuse the brevity!

mtorpey:
@james-d-mitchell: Thanks for the comments.  Changed as you suggested.

Failing Codespell due to an unrelated "satsifies" elsewhere in the doc.  Not sure why that's only appearing here, but I don't think it's related to this PR.

james-d-mitchell (approved):
Looks good

Issue #485, opened 2021-05-25
Read dreadnaut digraphs


wilfwilson:
This was raised in issue #15, but as far as I know never implemented.

_Originally posted by @mtorpey in https://github.com/digraphs/Digraphs/issues/15#issuecomment-261236109_

james-d-mitchell:
@wilfwilson any idea of where the dreadnaut format is defined?

james-d-mitchell:
@forktree

james-d-mitchell:
[nautest1.dre.txt](https://github.com/digraphs/Digraphs/files/7513698/nautest1.dre.txt)

mtorpey:
Help output from Drednaut on the command line, for reference:

```
Dreadnaut (Debian 2.6r10+ds-1ubuntu1) version 2.6 (64 bits).
> h
 Modes: An = dense, As = sparse, At = Traces; extra + to convert graph
 +- a : write automs     v,vv : write degrees    *=# : select invariant:
    b : write canong      w=# : set worksize (units of 2m)
 +- c : canonise            x : run nauty          -1 = user-defined
 +- d : digraph or loops  y=# : set tc_level        0 = none
    e : edit graph          z : write hashcode      1 = twopaths
 -f, f=#, f=[...] : set colours                     2 = adjtriang(K=0,1)
    g : read graph        $=# : set origin          3 = triples
  h,H : help               $$ : restore origin      4 = quadruples
    i : refine              ? : type options        5 = celltrips
    I : refine using invar  _ : compl  __ : conv    6 = cellquads
    j : relabel randomly    % : Mathon doubling     7 = cellquins
 k=# # : set invar levels   & : type colouring      8 = distances(K)
  K=# : set invar param    && : + quotient matrix   9 = indsets(K)
  l=# : set line length   >ff : write to file      10 = cliques(K)
 +- m : write markers    >>ff : append to file     11 = cellcliq(K)
  n=# : set order      ->/->> : close/flush output 12 = cellind(K)
    o : write orbits      <ff : read from file     13 = adjacencies
 +- p : set autom format    @ : save canong        14 = cellfano
    q : quit                # : canong = savedg?   15 = cellfano2
  r,R : relabel/subgraph   ## : + write mapping    16 = refinvar
  s=# : random g (p=1/#) sr=# : random reg   "..." : copy comment
 -G,G=# : schreier param  F=# : fix extra vertex  FF: fix target vertex
[...]

PR #651, opened 2024-05-26
Dreadnaut support

pramothragavan:
Broadly speaking, a dreadnaut file starts with "configuration" information about the graph, such as the number of vertices (denoted by 'n'), the start index for vertex numbering (denoted by '$') and whether or not a graph is a digraph (denoted by the presence of 'd').  The configuration section always ends with a 'g'. The rest of the file gives information concerning individual vertices in the form of adjacency lists. For example:

```
n=2
$=1
d
g
1: 1 2;
2: 2;
```

would represent a 1-indexed digraph with 2 vertices with edges {1,1}, {1,2}, {2,2}.

General overview:

**Decoder:**

- `DIGRAPHS_ParseDreadnautConfig` aims to get values for either '$' (which indicates the start index for vertex numbering) or 'n' (which indicates the number of vertices). Note that '$' defaults to 0 and that I chose to reindex all graphs such that vertex numbering starts at one (which I think is convention for the Digraphs package?)
-  `DIGRAPHS_LegalDreadnautEdge` aims to filter out illegal edges and throws an error if an edge is illegal. An example of an illegal edge might be a loop for an undirected graph or an edge containing a vertex that is not allowed within the constraints of the values of '$' and 'n'. (In the case of illegal edges, nauty throws a warning message and then ignores the edge so I was trying to replicate this behaviour).
- `DIGRAPHS_SplitDreadnautLines` effectively takes a line of dreadnaut (e.g. "1: 2 3 5; 4: 2 1 3; 2: 3;") and aims to split this into parts which are to be handled individually (in this case the parts would be ["1: 2 3 5;", "4: 2 1 3;", "2: 3;"]). The idea here is that although usually these parts would each be on their own line, it's techincally fine for some or all of them to share a line (with or without a semicolon) so I thought it made more sense to condense everything onto one line and then split into parts. There are various auxiliary commands that can be used within the dreadnaut format alongside the definition of the graph (more info [here](https://pallini.di.uniroma1.it/nug28.pdf#page=6)) which I mostly chose to neglect, with the exception of 'f' which defines a partition of vertices. Note that '$$' at the end of a file means reindex the graph to start counting at 0 (which I ignored).
- `DIGRAPHS_ParseDreadnautGraph` intends to parse the non-configuration part of the file, which has been split into parts after being fed through to `DIGRAPHS_SplitDreadnautLines`

These are all combined in `ReadDreadnautGraph`.

**Encoder:**
`WriteDreadnautGraph` takes a digraph and encodes into dreadnaut format.

I'm in the process of writing documentation!


gap/io.gi
```diff
@@ -1389,6 +1389,370 @@ function(str)
...
+    return edgeList;
+end);
+
+InstallMethod(ReadDreadnautGraph, "for a digraph", [IsString],
+function(name)
+  local file, config, graphData, line, edgeList, foundG, r, D, exclamPosition;
+  file := IO_CompressedFile(UserHomeExpand(name), "r");
+
+  if file = fail then
+    ErrorNoReturn("cannot open the file given as the 1st argument <name>,");
```
james-d-mitchell, line 1685: Please include the file name in the error message.

gap/io.gi
```diff
@@ -1459,6 +1823,68 @@ function(name, D)
...
 end);
 
+InstallMethod(WriteDreadnautGraph, "for a digraph", [IsString, IsDigraph],
+function(name, D)
+  local file, n, verts, nbs, labels, i, degs, filteredVerts,
+        out, positions, joinedPositions, dflabels;
+
+  file := IO_CompressedFile(UserHomeExpand(name), "w");
+  if file = fail then
+    ErrorNoReturn("cannot open the file given as the 1st argument <name>,");
```
james-d-mitchell, line 1833: Same again include the file name in the error message.

james-d-mitchell (changes requested):
Generally speaking this looks really good! I've added a few comments, mostly about the adding some details to the error messages if possible. Some general comments:

* it'd be great if the error messages could report in what line of the file the error occurs, I think we discussed this, but don't exactly remember the outcome of this. One approach would be to store the original file contents in a variable, and then search within that for the part that causes the error. I'm not sure if this would actually work or not, just a thought.
* in the PR description you mention:

> Note that '$' defaults to 0 and that I chose to reindex all graphs such that vertex numbering starts at one (which I think is convention for the Digraphs package?)

this sounds appropriate, and yes this is the convention (really more a requirement in Digraphs, i.e. at present it's only possible to have digraphs with vertices `[1 .. n]` for some `n`). It'd be best if the code at the very least issued a warning when you are renumber the vertices, to avoid violating the principal of least astonishment (i.e. try to read a graph with nodes not `[1 .. n]`, then silently getting a graph with nodes `[1 .. n]` would be surprising, so better issue a warning that this is happening). It'd also be useful to have the original vertices as labels in the newly constructed digraph, so if for example the graph has 0-indexed vertices, then the labels of the vertices in the output graph would be set using `DigraphSetVertexLabels(D, [0 .. n - 1]);` (if this is the correct mapping). 
* You mention in a couple of other places in the PR description that your code potentially ignores some other parts of dreadnaut files, if you detect parts that are ignored for whatever reason then, please issue a warning for each of these too, again to avoid surprising the user. 
* Have you checked how good the code coverage of your tests is? Given the number of lines of code in the implementation versus the number of lines of tests, I'm guessing that there's maybe some work to do there. I've sent you a python script by email that you can use to check the code coverage, just run `./code-coverage-test-gap.py tst/standard/io.tst` inside the digraphs directory.

gap/io.gi
```diff
@@ -1389,6 +1389,370 @@ function(str)
...
+        RemoveCharacters(vertex, " ");
+        vertex := Int(vertex);
+        adjacencyPart := subparts[2];
+      fi;
+
+      vertex := vertex - r.dollarValue + 1;
+
+      if vertex > r.nValue or vertex < 1 then
+        Error("illegal vertex ", vertex, " (original indexing)");
+        continue;
```
james-d-mitchell, line 1656:
```suggestion
```

james-d-mitchell:
@pramothragavan please let me know when you think this is ready again, and thanks !

pramothragavan:
> @pramothragavan please let me know when you think this is ready again, and thanks !

Will do!

mtorpey:
Hi @pramothragavan!  Looking forward to hopefully seeing you soon for the new VIP.

What state did this Dreadnaut project get to?  Would it be a good thing for you to get back into this semester if there's work still to do on it?

pramothragavan:
Hi — there are definitely some kinks to be dealt with, but I think it would be a good place to start!

pramothragavan:
This is a significant overhaul on previous versions -- `WriteDreadnautGraph` is untouched, but the decoder has been completely rewritten.

As @james-d-mitchell suggested, I've taken the parser used in the dreadnaut program and effectively rewritten it in GAP. The original C code uses a stream to parse character by character. GAP has a Stream object, but this lacks some of the functionality needed, so I created a record called `Stream` that aligns GAP's streams with how they're used in C. Other helper functions I've added:

- `DIGRAPHS_GETNWC` finds the next character in the stream that is not in `" ,\t"`
- `DIGRAPHS_GETNWL` finds the next character in the stream that is not in `" \n\t\r"`
- `DIGRAPHS_readinteger` reads integers from the stream (i.e. avoiding issues with reading `"10"` as opposed to `"1"` and `"0"` that might arise when parsing character by character)
- `DIGRAPHS_GetInt` also reads the next integer from the stream. There are some instances where dreadnaut allows for an optional `'='` character (e.g. `n=2` is the same as `n2`). This function ignores any `'='` characters and then calls `DIGRAPHS_readinteger`. 
- `DIGRAPHS_readgraph` parses the graph's adjacency data
- `DIGRAPHS_ParsePartition` is used to parse a partition, if given. The partition is stored using vertex labels.

Documentation for various commands is given [here](https://pallini.di.uniroma1.it/nug28.pdf#page=6)  (pages 6-12). Many of these are used to manipulate the graph and I have focused on supporting commands more closely tied to directly defining the graph. 

For now, I need to write (many) tests but I'm also interested if there are any commands that you'd like to see support for. I'm happy to implement anything really, but didn't want to waste time on things you didn't want. The commands that I am currently supporting are:

- All of those mentioned in section (A) of the above link. In dreadnaut, these would just define the mode which dreadnaut is using. This is important for subsequent use of nauty/traces, but is irrelevant for actually reading in the graph so this is just ignored.
- From section (B): `n=#`, `g` (and all subcommands), `_`, `__`
- From section (C): `f`
- From section (D): `$=#`, `$$`, `+`, `d, -d`
- From section (F): `"..."`, `!`, `q`

I think a couple of the unsupported commands from (B) might be worth looking into. Anything unsupported currently should raise an InfoWarning, with the exception of `<`, `>`, `e` (these three relate to reading in, outputting and editing graphs) which raise `ErrorNoReturn`.

gap/io.gi
```diff
@@ -1390,6 +1390,571 @@ function(str)
...
+        CloseStream(Stream.file);
+        ErrorNoReturn("Expected integer on line ", r.newline,
+                    " but was not found");
+    fi;
+
+    minus := char = '-';
+    if char = '-' or char = '+' then
+        res := 0;
+    else
+        res := Int([char]);
```
james-d-mitchell, line 1430:
```suggestion
        res := IntChar(char);
```

gap/io.gi
```diff
@@ -1390,6 +1390,571 @@ function(str)
...
+    fi;
+
+    char := Stream.GetChar(r);
+
+    while IsDigitChar(char) do
+        res := res * 10 + Int([char]);
```
james-d-mitchell, line 1436:
```suggestion
        res := res * 10 + IntChar(char);
```

gap/io.gi
james-d-mitchell: Ah ok my bad, happy with how it is

gap/io.gi
```diff
@@ -1390,6 +1390,600 @@ function(str)
...
+    local c, w, v, neg;
+
+    v := 1;
+    neg := false;
+    c := ' ';
+
+    if r.n = fail then
+      CloseStream(Stream.file);
+      ErrorNoReturn("Vertex number must be declared before `g' ",
+                  "on line ", r.newline);
```
james-d-mitchell, line 1472:
```suggestion
                   "on line ", r.newline);
```

gap/io.gi
```diff
@@ -1390,6 +1390,600 @@ function(str)
...
+        if IsDigitChar(c) then
+            Stream.UngetChar(r, c);
+            w := DIGRAPHS_readinteger(r, Stream);
+            w := w - r.labelorg + 1;
+
+            if neg then
+                neg := false;
+
+                if (w < 1) or (w > r.n) or (w = v and not r.digraph) then
+                    Info(InfoWarning, 1, "Ignoring illegal edge (",
```
james-d-mitchell, line 1488: You might consider using `StringFormatted` to make this easier to digest.

james-d-mitchell (commented):
Some comments delivered verbally, this looks really good, just some minor things to do.

pramothragavan:
#485 I think this is basically complete, but I don't really like some of the behaviour with `ReadDigraphs`. As @james-d-mitchell requested verbally, I've implemented the hashsets `WholeFileEncoders` and `WholeFileDecoders`, which currently only contain `WriteDreadnautGraph` and `ReadDreadnautGraph` respectively, but is a bit more futureproof than the current implementation (hopefully DIMACS can also feature in `ReadDigraphs`?). There are also corresponding functions `IsWholeFileEncoder` and `IsWholeFileDecoder`, but I'm not sure where the documentation for this is meant to go.

Initially, the idea was that we need to handle whole file decoders separately to single line decoders in `ReadDigraphs`. However, it transpires that if multiple graphs were given in a .dre file, dreadnaut would just read in the last one (effectively overwriting the previous graph each time). This behaviour is mirrored by `ReadDreadnautGraph` (with an InfoWarning issued) and by extension `ReadDigraphs`, but this outcome feels unexpected.

I also think there is ambiguity as to whether you should specify `DigraphFromDreadnautString` or `ReadDreadnautGraph` as the optional decoder argument in `ReadDigraphs`. I think having two encoders/decoders might be overcomplicating things, especially given that `ReadDigraphs` could always be used instead of `ReadDreadnautGraph`. Same for writes.

doc/io.xml
```diff
@@ -127,6 +127,19 @@ gap> for x in iter do od;
         Digraphs are pickled using the &IO; package. This is particularly good
         when the <Ref Attr="DigraphGroup"/> is non-trivial.
       </Item>
+      <Mark>dreadnaut (.dre)</Mark>
+      <Item>
+        A graph format designed for directed and undirected graphs.
+        It supports loops but multiple edges will be ignored.  The format consists of an
```
james-d-mitchell, line 133:
```suggestion
        The format supports loops but multiple edges will be ignored.  The format consists of an
```

gap/io.gi
```diff
@@ -1390,22 +1457,605 @@ function(str)
...
+        else
+          partition[v1] := 1;
+        fi;
+        r.partition := partition;
+        return;
+    elif c <> '[' then
+        ErrorNoReturn("Partitions should be specified",
+                    " using one of the following formats:\n",
+                      "  * A single number (e.g. 'f=3')\n",
+                      "  * A list of cells (e.g. 'f=[...]')",
```
james-d-mitchell, line 1692: Seems like the `\n` is missing here, but my comment is that I think error messages shouldn't have `\n` in them at all.

james-d-mitchell (changes requested):
Looks good, some minor changes suggested, and then I'll be happy to merge. One thing that is less major: can you please use 2 spaces for indentation, rather than 4?

doc/io.xml
```diff
@@ -127,6 +127,19 @@ gap> for x in iter do od;
...
+        initial section that defines the graph's structural properties, such as the number of
+        vertices, the index origin for vertices, and whether the graph is directed. This is followed
```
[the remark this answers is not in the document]
pramothragavan: Yeah it's exactly that. I'll change it to starting value for vertices

doc/io.xml
```diff
@@ -770,3 +809,82 @@ gap> OutNeighbours(gr);
...
+<ManSection>
+  <Oper Name="DigraphFromDreadnautString" Arg="s"/>
+  <Oper Name="DreadnautString" Arg="digraph"/>
+  <Description>
+    These operations read or write a single digraph to or from a string in the dreadnaut format, as appropriate. <P/>
+
+    <C>DigraphFromDreadnautString</C> expects the argument <A>s</A> to be a string, containing a single graph in dreadnaut format,
+    and returns a digraph.<P/>
+
+    If the vertices are not already 1-indexed, they will be reindexed accordingly.
```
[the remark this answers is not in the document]
pramothragavan: They're not! But I'll add that in

doc/io.xml
```diff
@@ -770,3 +809,82 @@ gap> OutNeighbours(gr);
...
+    If a partition is explicitly specified, each vertex will be assigned a label corresponding to its part in the partition.
+    Should the graph be undirected, the symmetric closure of the graph will be returned.
+    See <Ref Oper="DigraphVertexLabels"/> and <Ref Prop="DigraphSymmetricClosure"/>.<P/>
+
+    <C>DreadnautString</C> expects a digraph <A>digraph</A> and returns a string.<P/>
+
+    If <A>digraph</A> has multiple edges, then <C>DreadnautString</C> returns a digraph constructed from <A>digraph</A> by removing all multiple edges. 
+    If non-default integer vertex labels are provided, they will be used to define the partition structure.<P/>
```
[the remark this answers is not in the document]
pramothragavan: Ah so what I mean is that in GAP, if you create a digraph on n vertices, its vertex labels will be `[1 .. n]` (default integer vertex labels). If the given digraph has vertex labels different to `[1 .. n]` then (as long as the labels are integers), they will be used to define a partition. So, if the labels are `[1, 1, ..., 1]` in GAP then this would be converted into the partition `f = [1 2 .. n]` in dreadnaut (i.e. all vertices in the same part).

My intention was that if the labels in GAP are the default integer vertex labels, then the user wasn't trying to define a partition so no need to create one in dreadnaut (in dreadnaut the default partition is every vertex in the same part). But I'm now realising maybe the user might have intentionally set the vertex labels to `[1 .. n]` and wants this to inform a partition. I could raise a Y/N message about creating a partition in the case of default integer vertex labels? This might be annoying because it'd pop up most of the time, I imagine.

gap/io.gi
```diff
@@ -567,9 +605,19 @@ function(arg...)
       Info(InfoWarning, 1, "Writing to ", name);
     fi;
     file := DigraphFile(name, encoder, mode);
+    if NameFunction(file!.coder) in WholeFileEncoders and file!.mode <> "w" then
+      Info(InfoWarning, 1, NameFunction(file!.coder), " is a whole file ",
+          "encoder and so the argument <mode> is defaulting to \"w\". This ",
+          "will overwrite any current file contents. Continue (Y/N)?");
+      if not (NormalizedWhitespace(LowercaseString(ReadLine(InputTextUser())))
+              in ["y", "yes"]) then
```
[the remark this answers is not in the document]
pramothragavan: Think we're fine because we're doing `LowercaseString`?

doc/io.xml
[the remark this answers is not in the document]
james-d-mitchell: In some sense the less "magic" the better, be a good idea to simply add the partition as an argument to the function, then just it if present, and don't use it if not. Not completely sold on the y/n questions because this requires user intervention, so couldn't easily use it in code, if you see what I mean.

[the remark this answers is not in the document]
pramothragavan: Agreed, I think this is better. I suppose that if the user wanted to write to a file using `WriteDigraphs`, it wouldn't be possible to include a partition? I'm happy to let this slide and maybe just add a note in the doc saying they could use a combination of `DreadnautString` and `IO_Write` to achieve that result. If you think that sounds good, I'll implement it and hopefully wrap this up!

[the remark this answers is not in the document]
james-d-mitchell: > Agreed, I think this is better. I suppose that if the user wanted to write to a file using `WriteDigraphs`, it wouldn't be possible to include a partition? I'm happy to let this slide and maybe just add a note in the doc saying they could use a combination of `DreadnautString` and `IO_Write` to achieve that result.
> 
> ^ I've just implemented this, so hopefully we're all good @james-d-mitchell

Sounds good to me.

gap/io.gi
```diff
@@ -567,6 +605,16 @@ function(arg...)
       Info(InfoWarning, 1, "Writing to ", name);
     fi;
     file := DigraphFile(name, encoder, mode);
+    if NameFunction(file!.coder) in WholeFileEncoders and file!.mode <> "w" then
```
james-d-mitchell, line 608: One final comment, can you make this y/n choice optional please? For the same reason as elsewhere, it will be impossible to use this in code if user input is required, best to have a flag/option something that disables this behaviour. If that is too hard to achieve within the current framework, then maybe the best thing to do is just to issue the info warning, but remove the proceed check?

gap/io.gi
```diff
@@ -1118,59 +1175,60 @@ InstallMethod(DigraphFromPlainTextString, "for a string",
...
-    ErrorNoReturn("the format of the file given as the 2nd argument <name> ",
-                  "cannot be determined,");
-  end;
-
   # Helper function to read a string into a non-negative integer
   int_from_string := function(string)
     local int;
     int := Int(string);
     if int = fail or int < 0 then
-      malformed_file();
```
james-d-mitchell, line 1146: Sorry also: why did you make this change please? In particular, where is `file` now closed?

gap/io.gi
pramothragavan: Yeah I agree it's not ideal, but I think user could avoid this behaviour by specifying ``mode="w"`` when they're using a whole file encoder.  So, ``WriteDigraphs(filename, D, WriteDIMACSDigraph, "w")`` wouldn't run into this issue. In general ``mode`` defaults to ``"a"`` so I felt the need to warn before doing something destructive. Maybe it'd be better to change to infowarning to

```
      Info(InfoWarning, 1, NameFunction(file!.coder), " is a whole file ",
           "encoder and so the argument <mode> is defaulting to \"w\". This ",
           "will overwrite any current file contents. (This warning can be avoided by",
           "specifying the argument <mode> as \"w\".) Continue (Y/N)?");
```

james-d-mitchell: It's more the issue that user input is required that I'm concerned about. Could we maybe just refuse to write if "a" is passed as the mode? I.e. forbid "a" when using whole file decoders? The info warning can then be an error?

gap/io.gi
pramothragavan: So I think the idea is that you could either reach this point by using ``ReadDigraphs``, ``ReadDIMACSDigraph`` or ``DigraphFromDIMACSString``. In the third case, there wouldn't be a file to close so we wouldn't want any file closing logic in this function. For the first two cases, file closing should be handled by ``ReadDigraphs`` here:

https://github.com/digraphs/Digraphs/blob/54ea602693573b6c2f525be5d31265bd5208e2c5/gap/io.gi#L643-L645

gap/io.gi
pramothragavan: Yeah cool that makes sense

gap/io.gi
james-d-mitchell: Sounds good!

james-d-mitchell (approved):

gap/io.gi
pramothragavan: Sorry I've messed up here and then to make things worse I've managed to convince you that it was fine. You're right that if we're reading in malformed files for a whole file decoder then there was nothing in place to close these files. PR #757 addresses this. 

What's more, there's an underlying issue with `WriteDigraphs`. Regardless of file format, if the string encoder raises an error, there's nothing in place to close the file. I'm not really sure how to fix this, but there's more info in the PR. Sorry that this is coming after the merge

james-d-mitchell:
No worries @pramothragavan it happens! I could also have checked before merging. Looks like you have a fix in the works, which is great!

Issue #570, opened 2023-01-18
Rename DigraphDijkstra


mtorpey:
[`DigraphShortestPath`](https://digraphs.github.io/Digraphs/doc/chap5_mj.html#X80E9D645843973A6) and [`DigraphDijkstra`](https://digraphs.github.io/Digraphs/doc/chap5_mj.html#X79352A8286D1D8F6) are two closely related functions that both find the shortest path between two vertices, where "shortest" means "fewest edges".

The algorithm they're using appears to be basically the same: it's a breadth-first search starting at the source vertex and proceeding until the destination vertex is found.  Dijkstra's algorithm in general is smarter than this, since it uses edge lengths to decide where to explore next; but we don't have edge lengths, so this isn't relevant here.

The reason both operations are valuable is that they have different outputs.  They both return two lists, but don't be fooled!  DigraphShortestPath returns
> a pair of lists [v,a] as follows:
>   •    v is the list [v_1, v_2, ..., v_n].
>   •    a  is  the  list  of  positive  integers  [a_1,  a_2,  ..., a_n-1] where for each each i < n, a_i is the position of v_i+1 in OutNeighboursOfVertex(digraph,v_i) corresponding to the edge e_i.

whereas DigraphDijkstra returns
> two lists. Each element of the first list is the distance of the corresponding element from source. If a vertex was not visited in the process of calculating the shortest distance to target or if there is no path connecting that  vertex with source, then the corresponding distance is infinity. Each element of the second list gives the previous vertex in the shortest path from source to the corresponding vertex. For source and for any vertices that remained unvisited this will be -1.

The issue here is that "DigraphDijkstra" isn't a helpful name.  We usually name our operations after the thing they return rather than the author of the core algorithm we're using.  The output of DigraphShortestPath is consistent with other "paths" in the package, so this is reasonable.  Hence we need a name to replace "Dijkstra".

The main thing that distinguishes Dijkstra from ShortestPath is that it gives information on how to reach *multiple* vertices in the shortest distance: all vertices in the connected component, or else all the ones it can find while searching for `destination`.  Hence, a good option might be `DigraphShortestPaths`, which I don't think would be too confusing.

We could make this change easily, although we should make sure to preserve DigraphDijkstra as a synonym for backwards compatibility, till 2.0

Any thoughts?

mtorpey:
Added to the VIP board as an easy task, noting that it's important we keep `DigraphDijkstra` as a synonym to preserve backwards compatibility.

[...]

PR #656, opened 2024-06-07
Add edge-weighted shortest path algorithms

mtorpey:
This adds algorithms for finding shortest paths in edge-weighted digraphs, where "shortest" means "lowest total weight".

Three different "shortest path" problems are solved:
- All pairs: `DigraphShortestPaths(digraph)`
- Single source: `DigraphShortestPaths(digraph, source)`
- Source and destination: `DigraphShortestPath (digraph, source, dest)`

The "all pairs" problem has two algorithms:
- *Johnson*: better for sparse digraphs
- *Floyd–Warshall*: better for dense graphs

The "single source" problem has three algorithms:
- If "all pairs" is already known, extract information for the given source
- *Dijkstra*: faster, but cannot handle negative weights
- *Bellman–Ford*: slower, but handles negative weights

The "source and destination" problem calls the "single source" problem and
extracts information for the given destination.

Justification and benchmarks are in @RaiyanC's MSci thesis, Chapter 6.

### DIFF
--- a/doc/weights.xml
+++ b/doc/weights.xml
@@ -122,3 +122,116 @@ gap> EdgeWeights(T);
     </Description>
     </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="EdgeWeightedDigraphShortestPaths">
+<ManSection>
+  <Attr Name="EdgeWeightedDigraphShortestPaths" Label="for a digraph" Arg="digraph"/>
+  <Oper Name="EdgeWeightedDigraphShortestPaths" Label="for a digraph and a pos int" Arg="digraph, source"/>
+  <Returns>A record.</Returns>
+  <Description>
+    If <A>digraph</A> is an edge-weighted digraph, this attribute returns a
+    record describing the paths of lowest total weight (the <E>shortest
+    paths</E>) connecting each pair of vertices.  If the optional argument
+    <A>source</A> is specified and is a vertex of <A>digraph</A>, then the
+    output will only contain information on paths originating from that
+    vertex. <P/>
+
+    In the two-argument form, the value returned is a record containing three
+    components: <C>distances</C>, <C>parents</C> and <C>edges</C>.  Each of
+    these is a list of integers with one entry for each vertex in the
+    digraph. <P/>
+    <List>
+      <Item>
+        <C>distances[v]</C> is the total weight of the shortest path from
+        <A>source</A> to <C>v</C>.
+      </Item>
+      <Item>
+        <C>parents[v]</C> is the final vertex before <C>v</C> on the shortest
+        path from <A>source</A> to <C>v</C>.
+      </Item>
+      <Item>
+        <C>edges[v]</C> is the index of the edge of lowest weight going from
+        <C>parents[v]</C> to <C>v</C>.
+      </Item>
+    </List>
+    Using both these components together, you can find the shortest edge
+    weighted path to all other vertices from a starting vertex. <P/>
+
+    If no path exists from <A>source</A> to <C>v</C>, then <C>parents[v]</C> and
+    <C>edges[v]</C> will both be <K>fail</K>.  The distance from <A>source</A>
+    to itself is considered to be 0, and so both <C>parents[<A>source</A>]</C> and
+    <C>edges[<A>source</A>]</C> are <K>fail</K>. <P/>
+
+    In the one-argument form, the value returned is also a record containing
+    components <C>distances</C>, <C>parents</C> and <C>edges</C>, but each of
+    these will instead be a list of lists in which the <C>i</C>th entry is the
+    list that corresponds to paths starting at <C>i</C>.  In other words, the
+    following equalities apply.
+    <List>
+      <Item>
+        <C>EdgeWeightedDigraphShortestPaths(<A>digraph</A>).distances[<A>source</A>]
+        = EdgeWeightedDigraphShortestPaths(<A>digraph</A>, <A>source</A>).distances</C>
+      </Item>
+      <Item>
+        <C>EdgeWeightedDigraphShortestPaths(<A>digraph</A>).parents[<A>source</A>]
+        = EdgeWeightedDigraphShortestPaths(<A>digraph</A>, <A>source</A>).parents</C>
+      </Item>
+      <Item>
+        <C>EdgeWeightedDigraphShortestPaths(<A>digraph</A>).edges[<A>source</A>]
+        = EdgeWeightedDigraphShortestPaths(<A>digraph</A>, <A>source</A>).edges</C>
+      </Item>
+    </List>
+
+    Edge weights can have negative values, but this operation will fail with an
+    error if a negative-weighted cycle exists. <P/>
+
+    For a simple way of finding the shortest path between two specific vertices,
+    see <Ref Oper="EdgeWeightedDigraphShortestPath"/>. See also the non-weighted
+    operation <Ref Oper="DigraphShortestPath"/>. <P/>
+
+    <Example><![CDATA[
+gap> g := EdgeWeightedDigraph([[2, 3], [4], [4], []], [[5, 1], [6], [11], []]);
+<immutable digraph with 4 vertices, 4 edges>
+gap> EdgeWeightedDigraphShortestPath(g, 1);
+rec( distances := [ 0, 5, 1, 11 ], edges := [ fail, 1, 2, 1 ],
+  parents := [ fail, 1, 1, 2 ] )
+gap> g := EdgeWeightedDigraph([[2], [3], [1]], [[-1], [-2], [-3]]);
+<immutable digraph with 3 vertices, 3 edges>
+gap> EdgeWeightedDigraphShortestPath(g, 1);
+Error, negative cycle exists,
+gap> g := EdgeWeightedDigraph([[2], [3], [1]], [[1], [2], [3]]);
+<immutable digraph with 3 vertices, 3 edges>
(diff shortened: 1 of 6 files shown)


james-d-mitchell (commented):
Another quick partial review: it looks like there's another implementation of Floyd-Warshall in this PR, is there a reason not to use the existing implementation? Iirc the existing implementation is pretty flexible, it should be possible to use it for this too, but I might be mistaken

mtorpey:
> Is there a reason not to use the existing implementation [of Floyd–Warshall]? Iirc the existing implementation is pretty flexible.

I've taken a look at FLOYD_WARSHALL in digraphs.c, and it would need quite a lot of reworking to use here, for 2 reasons:
- It always creates a distance matrix consisting of two values (edge and non-edge) rather than multiple values
- It only stores distances, and doesn't remember info about the actual paths.

Obviously we could refactor the C code, but perhaps that's a future issue?

gap/weights.gi
```diff
@@ -166,3 +166,453 @@ function(digraph)
...
+#
+# The "single source" problem has three algorithms:
+# - If "all pairs" is already known, extract information for the given source
+# - Dijkstra: faster, but cannot handle negative weights
+# - Bellman-Ford: slower, but handles negative weights
+#
+# The "source and destination" problem calls the "single source" problem and
+# extracts information for the given destination.
+#
+# Justification and benchmarks are in Raiyan's MSci thesis, Chapter 6.
```
james-d-mitchell, line 191: Any chance of adding a link to this thesis?

gap/weights.gi
```diff
@@ -166,3 +166,453 @@ function(digraph)
...
+end);
+
+InstallMethod(EdgeWeightedDigraphShortestPaths,
+"for a digraph with edge weights and known shortest paths and a pos int",
+[IsDigraph and HasEdgeWeights and HasEdgeWeightedDigraphShortestPaths, IsPosInt],
+function(digraph, source)
+  local all_paths;
+  if not source in DigraphVertices(digraph) then
+    ErrorNoReturn("the 2nd argument <source> must be a vertex of the ",
+                  "digraph <digraph> that is the 1st argument,");
```
james-d-mitchell, line 228:
```suggestion
                  "of the 1st argument <digraph> (a digraph),");
```
or 
```suggestion
                  "of the 1st argument <digraph>,");
```
be a bit shorter?

gap/weights.gi
```diff
@@ -166,3 +166,453 @@ function(digraph)
...
+             parents   := all_paths.parents[source]);
+end);
+
+InstallMethod(EdgeWeightedDigraphShortestPaths,
+"for a digraph with edge weights and a pos int",
+[IsDigraph and HasEdgeWeights, IsPosInt],
+function(digraph, source)
+  if not source in DigraphVertices(digraph) then
+    ErrorNoReturn("the 2nd argument <source> must be a vertex of the ",
+                  "digraph <digraph> that is the 1st argument,");
```
james-d-mitchell, line 243: Same comment as above?

gap/weights.gi
```diff
@@ -166,3 +166,453 @@ function(digraph)
...
+end);
+
+InstallMethod(EdgeWeightedDigraphShortestPath,
+"for a digraph with edge weights and two pos ints",
+[IsDigraph and HasEdgeWeights, IsPosInt, IsPosInt],
+function(digraph, source, dest)
+  local paths, v, a, current, edge_index;
+  if not source in DigraphVertices(digraph) then
+    ErrorNoReturn("the 2nd argument <source> must be a vertex of the ",
+                  "digraph <digraph> that is the 1st argument,");
```
james-d-mitchell, line 260: Same again

gap/weights.gi
```diff
@@ -166,3 +166,453 @@ function(digraph)
...
+  elif not dest in DigraphVertices(digraph) then
+    ErrorNoReturn("the 3rd argument <dest> must be a vertex of the ",
+                  "digraph <digraph> that is the 1st argument,");
```
james-d-mitchell, line 263: same again

gap/weights.gi
```diff
@@ -166,3 +166,453 @@ function(digraph)
...
+        idx := adjMatrix[u][v][2];
+
+        distances[u][v] := w;
+        parents[u][v]   := u;
+        edges[u][v]     := idx;
+      fi;
+    od;
+  od;
+
+  # try every triple: distance from u to v via k
```
james-d-mitchell, line 417: Can you please add a comment about why we aren't using the C Floyd-Warshall here? Just to understand again in the future, doesn't have to be very elaborate comment.

gap/weights.gi
```diff
@@ -166,3 +166,453 @@ function(digraph)
...
+function(digraph)
+  local maxNodes, threshold, digraphVertices, nrVertices, nrEdges;
+
+  digraphVertices := DigraphVertices(digraph);
+  nrVertices := Size(digraphVertices);
+  nrEdges := DigraphNrEdges(digraph);
+
+  maxNodes := nrVertices * (nrVertices - 1);
+
+  # the boundary for performance is edge weight 0.125
```
james-d-mitchell, line 206: Can you possibly reflow this comment, so that it uses more of the width, and possibly add some punctuation if applicable? Thanks!

gap/weights.gi
```diff
@@ -166,3 +166,453 @@ function(digraph)
...
+          fi;
+        od;
+      fi;
+    od;
+  od;
+
+  # detect negative cycles
+  for i in vertices do
+    if distances[i][i] < 0 then
+      ErrorNoReturn("1st arg <digraph> contains a negative-weighted cycle,");
```
james-d-mitchell, line 437: This error message could be more helpful, possibly explain why a negative-weighted cycle is not ok, and how to compute the correct answer in this case?

gap/weights.gi
```diff
@@ -166,3 +166,453 @@ function(digraph)
...
+  # than total number of edges in a connected
+  # graph we use johnson's algorithm
+  # which performs better on sparse graphs, otherwise
+  # we use floyd warshall algorithm.
+  # This information is gathered from benchmarking tests.
+  threshold := Int(maxNodes / 8);
+  if nrEdges <= threshold then
+    return DIGRAPHS_Edge_Weighted_Johnson(digraph);
+  else
+    return DIGRAPHS_Edge_Weighted_FloydWarshall(digraph);
```
james-d-mitchell, line 217: Isn't there an issue here if the graph contains a negative cycle? In particular, if it does contain a negative cycle, then shouldn't we use the other algorithm? So, wouldn't it be better to return `fail` from `DIGRAPHS_Edge_Weighted_FloydWarshall` in the case `digraph` contains a negative cycle, and then just run `DIGRAPHS_Edge_Weighted_Johnson` anyway?

doc/weights.xml
```diff
@@ -122,3 +122,98 @@ gap> EdgeWeights(T);
...
+    </List>
+    Using these three components together, you can find the shortest edge
+    weighted path to all other vertices from a starting vertex. <P/>
+
+    If no path exists from <A>source</A> to <C>v</C>, then <C>parents[v]</C> and
+    <C>edges[v]</C> will both be <K>fail</K>.  The distance from <A>source</A>
+    to itself is considered to be 0, and so both <C>parents[<A>source</A>]</C> and
+    <C>edges[<A>source</A>]</C> are <K>fail</K>.
+    Edge weights can have negative values, but this operation will fail with an
+    error if a negative-weighted cycle exists. <P/>
```
james-d-mitchell, line 165: Not sure that this is accurate, isn't it only the case that it'll fail when the input digraph is "dense", and Floyd-Warshall is applied?

james-d-mitchell (changes requested):
Assuming that the algorithms work as intended (I haven't really checked this, but I'm hoping @mtorpey and @RaiyanC did), this looks good, there are a number of minor things that should be changed, and one other thing too:

* can you please add to the doc something about the complexity of each of the functions? 

We have tried to do this elsewhere in the documentation, and it's often really helpful.

markuspf:
Small nitpick about the "source-destination" solution: 

At the moment (as documented) this calls the single-source solution and extracts the information from that; this is in many cases way less efficient than exiting early once the destination is found; 

This is easy to fix and shouldn't be a blocker for this PR. Just leaving this here as it will pop back up when #570 as I purposefully implemented Dijkstra with source and destination and single source for exactly that reason.

gap/weights.gi
mtorpey: Updated as you suggested.

gap/weights.gi
mtorpey: Replaced with `return fail;` so that we can fall back to Johnson, as JDM suggested.

gap/weights.gi
mtorpey: Good idea. I'm using the second suggestion in all of these.

gap/weights.gi
mtorpey: Done, at the top of the function.

gap/weights.gi
mtorpey: Rewrote to be cleaner and more descriptive.

gap/weights.gi
mtorpey: Replaced with `return fail`, and the outer method will call `TryNextMethod`.  There are now documentation and info warnings explaining that we have no applicable method if there are negative cycles.

doc/weights.xml
mtorpey: Updated.

PR #657, opened 2024-06-12
Refactor Floyd–Warshall C implementation

mtorpey:
In digraphs.c we have a `FLOYD_WARSHALL` function that does a generalised version of the Floyd–Warshall algorithm with various options that customise the output.  As noted in comments, some of this was rather hacky, with flags for special cases that return a totally different type of object early in the function and so on.

This PR adds a `post` parameter to FLOYD_WARSHALL, which takes a post-processing function that inspects the state of the distance matrix and produces a final output.  So far we have three `post` functions:
- `FW_POST_DIAMETER`, used by `DIGRAPH_DIAMETER` (returns the highest number in the matrix, or fail)
- `FW_POST_HAS_CHANGED` used by `IS_TRANSITIVE_DIGRAPH` (returns a bool saying whether the matrix was changed in the main loop)
- `FW_POST_DISTANCES`, used by the other functions (returns the distance matrix as a GAP object)

and more could be added.  This replaces the `diameter` and `copy` parameters which previously hard-coded some of this into the function.

There are also some readability improvements, basically changing variable names to describe what they do:
- `dist1` and `dist2` are now `nonEdgeDist` and `edgeDist`
- `shortest` is now `zeroEdgeDist`
- `copy` is now `keepCopy`, and is now used for nothing but creating a copy of `dist`.
- `adj` is now `distCopy` (it's just a copy of `dist`)
- `out` was previously used for two things: out neighbours and function output.  This is now split into `neighbours` and `out`.

This will be a useful first step towards making the function more configurable, which might be useful for edge-weighted algorithms later.

### DIFF
--- a/src/digraphs.c
+++ b/src/digraphs.c
@@ -1115,78 +1115,66 @@ static Obj FuncIS_MULTI_DIGRAPH(Obj self, Obj digraph) {
  *       - 3 integers i, j, k
  *       - An integer n (the number of vertices of digraph)
  *      and modifies the matrix dist according to the values of i, j, k.
- *   3. Int val1 s.t initially dist[i][j] = val1 if [ i, j ] isn't an edge.
- *   4. Int val2 s.t initially dist[i][j] = val2 if [ i, j ] is an edge.
- *   5. bool copy:
- *      - If true, FLOYD_WARSHALL stores the initialised dist, and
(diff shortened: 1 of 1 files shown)


james-d-mitchell (approved):
Looks good to me, thanks @mtorpey

PR #662, opened 2024-06-20
ci: remove the duplicate Ubuntu from ubuntu.yml

james-d-mitchell:


james-d-mitchell:
Maybe there's a good reason for not doing this but I can't think what it is...

Joseph-Edwards (approved):
I think this is an artefact from when the Mac and Cygwin tests were part of the same workflow as Ubuntu. Now that they're separate, it seems sensible to remove it

PR #665, opened 2024-06-26
New workflow cancels old one

### LINKED ISSUE #661: The CI should cancel existing jobs (if any)
Just to avoid the situation where we push multiple times in quick succession to the same PR, and then have lots of CI jobs running, when we only care about the most recent ones.

Joseph-Edwards:
Fixes #661.


Joseph-Edwards:
I tested this on my fork by pushing some nonsense commit, and it seems to work.

Issue #660, opened 2024-06-20
The CI doesn't fail if there are undeclared local variables but should


james-d-mitchell:
Just noticed that in PR #655 there were undeclared local variables which resulted in:

~~~
Syntax warning: Unbound global variable
~~~

warnings at the start of the GAP session, but the CI all passed. We should fix this!

james-d-mitchell:
I've assigned you to this issue @Joseph-Edwards, but this isn't urgent at all. If you think it'd be easier for me to do this, then feel free to reassign it!

james-d-mitchell:
Resolved by #666

PR #666, opened 2024-06-26
Warnings when loading a package causes CI to fail

Joseph-Edwards:
As pointed out in #660, CI jobs were not failing when loading digraphs threw warnings. This PR addresses that by capturing the output from `LoadPackage` and checking for any occurrence of `"warning"`. If only syntax warnings are to be captured, this can be easily changed.

This has been implemented by editing the action that was created in #641. Hopefully, in the future, this functionality will be added to the official GAP action.


james-d-mitchell:
Thanks @Joseph-Edwards

Issue #671, opened 2024-08-27
Implement a hash function for digraphs


james-d-mitchell:
So that they can be inserted into:

* datastructures hash maps and sets
* dictionaries (for use with `Orbit` in the GAP library.

This is related to the proposed solution to #670.

james-d-mitchell:
Resolved in v1.9.0

PR #675, opened 2024-08-27
Add hash function for digraphs

reiniscirpons:
This PR adds a hash function for digraphs, resolving #671 .


src/digraphs.c
```diff
@@ -1384,6 +1384,80 @@ static bool EqJumbledPlists(Obj l, Obj r, Int nr, Int* buf) {
...
+/* Minor modifications to get it to compile in C rather than C++ and
+integrate with GAP  SL*/
+
+/* Digraphs takes parts of this source from GAP */
+
+#define BIG_CONSTANT(x) (x##LLU)
+
+#ifndef SYS_IS_64_BIT
+
+static inline uint32_t fmix4(uint32_t h) {
```
fingolfin, line 1400:
```suggestion
static inline UInt fmix(UInt h) {
```

src/digraphs.c
```diff
@@ -1384,6 +1384,80 @@ static bool EqJumbledPlists(Obj l, Obj r, Int nr, Int* buf) {
...
+
+  n   = DigraphNrVertices(digraph);
+  out = FuncOutNeighbours(self, digraph);
+
+#ifndef SYS_IS_64_BIT
+  uint32_t* buf;
+  buf = safe_calloc(n, sizeof(uint32_t));
+#else
+  uint64_t* buf;
+  buf = safe_calloc(n, sizeof(uint64_t));
```
fingolfin, line 1437: Hash functions are supposed to be as fast and low overhead as possible. Doing allocations in them is therefore suboptimal.

In this case, you feed this buffer into `HASHKEY_MEM_NC` in the end, which just has to read everything in that buffer out again and mix it back together.

So you could avoid using a buffer if just mix everything together it, using a "good enough" mixing function -- here by mixing function I mean one which takes two `UInt` and combines them into a single UInt in a way that's "good foro hasing. E.g. it should be far from commutative and associative.

(I think this also referred to as "hash combiner" and possibly other names).

A very simple hash combiner better than `+` would be
```
static UInt hash_combine(UInt a, UInt b) {
  return fmix(a + fmix(b))
}
```
but it is inefficient because it invokes `fmix` twice. I'll see if I can find a web page with better suggestions.

src/digraphs.c
```diff
@@ -1384,6 +1384,80 @@ static bool EqJumbledPlists(Obj l, Obj r, Int nr, Int* buf) {
...
+  for (i = 1; i <= n; i++) {
+    a = ELM_PLIST(out, i);
+    PLAIN_LIST(a);
+    nr = LEN_PLIST(a);
+    for (j = 1; j <= nr; j++) {
+#ifndef SYS_IS_64_BIT
+      buf[i - 1] += fmix4((uint32_t) INT_INTOBJ(ELM_PLIST(a, j)));
+#else
+      buf[i - 1] += fmix8((uint64_t) INT_INTOBJ(ELM_PLIST(a, j)));
+#endif
```
fingolfin, line 1449: With my suggestions from above, you could replace `buf` by a single var `UInt h = 0;` and then do
```suggestion
      h = hash_combine(h, (UInt) INT_INTOBJ(ELM_PLIST(a, j)));
```

src/digraphs.c
fingolfin: E.g. https://www.boost.org/doc/libs/1_55_0/doc/html/hash/reference.html#boost.hash_combine suggests
```C
seed ^= hash_value(v) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
```

src/digraphs.c
reiniscirpons: Thanks for the suggestion, getting rid of `buf` is very nice. I've not implemented this exact logic since we actually want the hash to be commutative on each list of out neighbors so that `Digraph([[1, 2], []])` and `Digraph([[2, 1], []])` get the same hash. I elected to do
```c
    for (j = 1; j <= nr; j++)
      h += fmix(INT_INTOBJ(ELM_PLIST(a, j)));
    h = fmix(h + 0x9e3779b9);
```
This way the order of elements within each out neighbour list does not matter, but when accumulating we introduce some non-commutativity so `Digraph([[1, 2], []])` and `Digraph([[], [1, 2]])`  should get different hashes.

fingolfin: Oh indeed, if the representation is not unique then commutativity may be useful and even necessary. My apologies, I should have considered that. Thank you for doing it properly despite my misleading comment

PR #679, opened 2024-08-28
Add DigraphAllChordlessCycles

MeikeWeiss:
This pull request adds a method for computing all chordless cycles. A cycle C is chordless if two vertices in C are not connected by an edge not included in C.


doc/attr.xml
```diff
@@ -1374,6 +1374,45 @@ gap> DigraphLongestSimpleCircuit(D);
...
+    If <A>digraph</A> is a digraph, then <C>DigraphAllChordlessCycles</C> 
+    returns a list of the <E>chordless cycles</E> in <A>digraph</A>. <P/>
+
+    A chordless cycle <M>C</M> is a undirected simple circuit (see Section <Ref Subsect="Definitions" Style="Number" /> ) 
+    where each pair of vertices in <M>C</M> are not connected by an edge not in <M>C</M>. 
+    Here, cycles of length two are ignored.<P/>
+
+    For a digraph without multiple edges, a simple circuit is uniquely
+    determined by its subsequence of vertices. However this is not the case for
+    a multidigraph.  The attribute <C>DigraphAllSimpleCircuits</C> ignores
```
reiniscirpons, line 1391:
```suggestion
    a multidigraph.  The attribute <C>DigraphAllChordlessCycles</C> ignores
```

tst/standard/attr.tst
```diff
@@ -948,6 +948,25 @@ gap> gr := Digraph([[3, 6, 7], [3, 6, 8], [1, 2, 3, 6, 7, 8],
...
+gap> DigraphAllChordlessCycles(D);
+[ [ 2, 1, 3 ] ]
+gap> D := CompleteDigraph(4);;
+gap> DigraphAllChordlessCycles(D);
+[ [ 2, 1, 3 ], [ 2, 1, 4 ], [ 3, 1, 4 ], [ 3, 2, 4 ] ]
+gap> D := Digraph([[2, 4, 5], [3, 6], [4, 7], [8], [6, 8], [7], [8], []]);;
+gap> DigraphAllChordlessCycles(D);
+[ [ 6, 5, 8, 7 ], [ 4, 3, 7, 8 ], [ 3, 2, 6, 5, 8, 4 ], [ 3, 2, 6, 7 ], 
+  [ 4, 1, 5, 6, 7, 3 ], [ 4, 1, 5, 8 ], [ 2, 1, 5, 6 ], [ 2, 1, 5, 8, 7, 3 ], 
+  [ 2, 1, 4, 3 ], [ 2, 1, 4, 8, 7, 6 ] ]
```
reiniscirpons, line 968: It might be worth standardizing the output in tests to future-proof them. Currently, if someone made a change to the `DigraphAllChordlessCycles` function in a way which changed e.g. the order of cycles or order of points in a cycle, then this would cause the test to break.

Maybe doing something like 
```
List(DigraphAllChordlessCycles(D), x->SortedList(List(x, SortedList)));
```
to ensure the test output does not depend on the ordering of the cycles and points. Alternatively this could be done inside the `DigraphAllChordlessCycles` function directly.

reiniscirpons (changes requested):
Thanks for the PR! A couple of suggestions but otherwise looks very good!

reiniscirpons (approved):

reiniscirpons:
Ill wait for the CI to pass and then merge. Thanks again!

PR #682, opened 2024-08-28
kernel: use GAP's GVAR_FUNC macro

fingolfin:
... to get rid of some repetition

This macro has been available since GAP 4.9.

Since GAp 4.12 we also have `GVAR_FUNC_0ARGS`, `GVAR_FUNC_1ARGS`, `GVAR_FUNC_2ARGS`, etc. which also help catch bugs (e.g. mismatch between number of arguments of the given C function and what GAP is being promised here).  But since this package still tries to be compatible with GAP 4.10 I didn't make use of them.

### DIFF
--- a/PackageInfo.g
+++ b/PackageInfo.g
@@ -274,7 +274,7 @@ Persons := [
     IsAuthor      := false,
     IsMaintainer  := false,
     Email         := "markus.pfeiffer@morphism.de",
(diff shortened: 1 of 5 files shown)


james-d-mitchell:
Thanks for the contribution @fingolfin! Any chance you could please run clang-format on `digraphs.c`, and rebase onto the most recent `stable-1.8` ?

fingolfin:
Done

PR #689, opened 2024-08-30
Added method AllUndirectedSimpleCircuits

MeikeWeiss:
I added a method that computes all undirected simple circuits by using the method AllSimpleCircuits.

### DIFF
--- a/gap/attr.gi
+++ b/gap/attr.gi
@@ -1526,22 +1526,22 @@ end);
 
 # Compute all undirected simple circuits by filtering the output
 # of DigraphAllSimpleCircuits
-InstallMethod(DigraphAllUndirectedSimpleCircuits, "for a digraph",
-                     [IsDigraph], function(D)
-    local digraph, cycles, keep, cycle, cycleRev;
-    digraph := DigraphSymmetricClosure(
-        DigraphRemoveAllMultipleEdges(DigraphMutableCopyIfMutable(D)));
-    cycles := Filtered(DigraphAllSimpleCircuits(digraph),
-        c -> Length(c) > 2 or Length(c) = 1);
-    keep := HashSet();
-    for cycle in cycles do
-        cycleRev := [cycle[1]];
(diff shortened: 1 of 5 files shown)


gap/attr.gi
```diff
@@ -1524,6 +1524,26 @@ function(D)
...
+    cycles := Filtered(DigraphAllSimpleCircuits(digraph),
+        c -> Length(c) > 2 or Length(c) = 1);
+    remove := [];
+    for cycle in cycles do
+        if not cycle in remove and Length(cycle) <> 1 then
+            cycleRev := [cycle[1]];
+            Append(cycleRev, Reversed(cycle{[2 .. Length(cycle)]}));
+            Add(remove, cycleRev);
+        fi;
+    od;
```
reiniscirpons, line 1543: Having `remove` be implemented using [`HashSet`](https://gap-packages.github.io/datastructures/doc/chap8_mj.html#X789A593B7C227BE5) from the datastructures package instead of a list would be more efficient since the function is doing repeated lookups.

gap/attr.gi
```diff
@@ -1524,6 +1524,26 @@ function(D)
...
+    return Difference(cycles, remove);
```
reiniscirpons, line 1544: Perhaps you could try flipping the logic so that instead of storing cycles to remove, the function stores cycles to keep, this is a small optimization and would avoid having a call to `Difference` at the end. So something like
```
keep := []
for cycle in cycles do
  cycleRev := [cycle[1]];
  Append(cycleRev, Reversed(cycle{[2 .. Length(cycle)]}));
  if (not cycle in keep and not cycleRev in keep) or Length(cycle) <> 1 then
    Add(keep, cycle);
  fi;
od;
return keep;
```
might work.

reiniscirpons (changes requested):
Thanks for the PR. Some small efficiency improvements and its good to go.

reiniscirpons (approved):

reiniscirpons:
Ill wait for the PR to pass then merge, thanks again!

PR #690, opened 2024-08-30
Add `SubdigraphsMonomorphisms`

### LINKED ISSUE #670: Compute subgraphs
It would be nice to have a method that computes, for a given digraph G and a digraph H, all subgraphs of G that are isomorphic to H. These subgraphs do not have to be induced subgraphs. It can be computed theoretically by MonomorphismsDigraphs but the output is here is a transformation and not a digraph. Moreover, the length of the output of this method for G isomorphic to H is the order of the automorphism group but G has only one subgraph. So using this method is maybe not the best way.

I have some code that computes all subgraphs with the output of MonomorphismsDigraphs, but it only works for certain subgraphs so far.  I will see if it can be generalised. But maybe an algorithm not based on MonomorphismsDigraphs would be better.

james-d-mitchell:
Resolves #670

### DIFF
--- a/doc/grahom.xml
+++ b/doc/grahom.xml
@@ -87,7 +87,7 @@
 
         For backwards compatibility, <A>injective</A> can also be <K>false</K>
         or <K>true</K> which correspond to the values <C>0</C> and <C>1</C>
-        described in the previous paragraph, respectively. 
+        described in the previous paragraph, respectively.
       </Item>
 
       <Mark><A>image</A></Mark>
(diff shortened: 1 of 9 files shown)


james-d-mitchell:
@MeikeWeiss could you please review this PR?

doc/grahom.xml
```diff
@@ -274,14 +274,35 @@ Transformation( [ 3, 1, 5, 4, 5 ] )
...
     <E>injective</E> homomorphisms.
     <Example><![CDATA[
-gap> gr1 := ChainDigraph(3);;
-gap> gr2 := Digraph([[3, 5], [2], [3, 1], [], [4]]);
-<immutable digraph with 5 vertices, 6 edges>
-gap> MonomorphismsDigraphs(gr1, gr2);
-[ Transformation( [ 1, 5, 4, 4, 5 ] ), 
-  Transformation( [ 3, 1, 5, 4, 5 ] ) ]
-gap> MonomorphismsDigraphsRepresentatives(gr1, CompleteDigraph(3));
-[ Transformation( [ 2, 1 ] ) ]
```
MeikeWeiss, line 284: I think you accidentally replaced the example of MonomorphismsDigraphs with examples of SubdigraphsMonomorphisms.

doc/grahom.xml
```diff
@@ -274,14 +274,35 @@ Transformation( [ 3, 1, 5, 4, 5 ] )
...
+    monomorphisms are all of the subdigraphs of <A>digraph2</A> that are
+    isomorphic to <A>digraph1</A>. Note that the subdigraphs of the previous
+    sentence are those obtained by applying the corresponding monomorphism to
+    the vertices and the edges of <A>digraph1</A>, and are therefore possibly
+    strictly contained in the induced subdigraph on the same vertex set.
+    <Example><![CDATA[
+gap> SubdigraphsMonomorphisms(CompleteBipartiteDigraph(2, 2),
+> CompleteDigraph(4));
+[ Transformation( [ 1, 3, 2 ] ), Transformation( [ 2, 3, 1 ] ),
+  Transformation( [ 3, 4, 2, 1 ] ) ]
```
MeikeWeiss, line 305: Maybe it would be nice to have also an example for SubdigraphsMonomorphismsRepresentatives, such that the difference between the methods is clear.

MeikeWeiss (changes requested):
Thanks, a lot and it looks good! I just have two remarks.

doc/grahom.xml
james-d-mitchell: Good point, doing it now

doc/grahom.xml
james-d-mitchell: Fixed!

MeikeWeiss (approved):
Thanks!

PR #678, opened 2024-08-28
Resolve issue #676

james-d-mitchell:
Properly transfer the original vertices as labels of the vertices in the reduced graph in `DigraphAllSimpleCircuits`.

Issue #676, opened 2024-08-28
`DigraphAllSimpleCircuits` failure when using a graph with "named" vertices


fingolfin:
Inspired by @mtorpey's talk just now, I wanted to look at the dependency graph of GAP packages. Here is the code snippet I used to create it:
```gap
vert := Set(RecNames(GAPInfo.PackagesInfo));;
f := {a,b} -> ForAny(PackageInfo(a)[1].Dependencies.NeededOtherPackages,x->x[1]=b);
gr := Digraph(vert, f);
```
Various operations with this work fine, but then I got this error:
```
gap> DigraphAllSimpleCircuits(gr);
Error, the 2nd argument <names> must be a list with length equal to the number of vertices of the digraph <D> that is the 1st argument, at /Users/mhorn/Projekte/GAP/repos/pkg/others/Digraphs/gap/labels.gi:62 called from
SetDigraphVertexLabels( C, Filtered( DigraphVertices( D ), function ( x )
        return OutDegrees( D ) <> 0;
    end ) ); at /Users/mhorn/Projekte/GAP/repos/pkg/others/Digraphs/gap/attr.gi:1460 called from
<function "DigraphAllSimpleCircuits for a digraph by out-neighbours">( <arguments> )
 called from read-eval loop at *stdin*:27
type 'quit;' to quit to outer loop
brk>
```

It seems that it gest confused by the use of strings as vertices instead of integers. Indeed if I use integers instead it works:
```gap
gap> gr2 := Digraph([1..Length(vert)], {i,j} -> f(vert[i], vert[j]));
<immutable digraph with 303 vertices, 134 edges>
gap> DigraphAllSimpleCircuits(gr2);
[ [ 211, 13 ] ]
```

james-d-mitchell:
Thanks for the report @fingolfin, will attempt to fix today!

james-d-mitchell:
Resolved in v1.9.0

PR #695, opened 2024-08-31
Add a test for issue #676

wilfwilson:
I've added a test for a version of the bug that @fingolfin reported in #676; see #678 which gave the fix.


james-d-mitchell:
Rebased onto `main` (the next release will be 1.9.0) to fix the gaplint job. Thanks @wilfwilson !

james-d-mitchell (approved):
Looks good!

Issue #669, opened 2024-08-26
Output of `PlanarEmbedding`


MeikeWeiss:
The planar embedding method returns only a minimal rotation system, i.e. for each vertex the method returns only a list with entries of smaller vertices. Such a list is not easy to work with. It must be possible to extend this list in such a way that we get a list containing for each vertex all neighbours in such a order that it defines a planar embedding.

Here are some examples of how the actual output and how it would be easier to work with the computed planar embedding:

~~~
PlanarEmbedding(CompleteDigraph(4));
[ [ 2, 4, 3 ], [ 3, 4 ], [ 4 ], [  ] ] 
 ------> [ [ 2, 4, 3 ], [ 3, 4, 1 ], [ 4, 2, 1], [ 1, 2, 3 ] ] 
~~~

Cubical Graph:

~~~
g:=DigraphByEdges([ [ 1, 3 ], [ 3, 1 ], [ 1, 7 ], [ 7, 1 ], [ 5, 7 ], [ 7, 5 ], [ 3, 5 ], [ 5, 3 ], [ 1, 4 ], [ 4, 1 ], [ 2, 3 ], [ 3, 2 ], [ 2, 4 ],[ 4, 2 ], [ 6, 7 ], [ 7, 6 ], [ 4, 6 ], [ 6, 4 ], [ 5, 8 ], [ 8, 5 ], [ 6, 8 ], [ 8, 6 ], [ 2, 8 ], [ 8, 2 ] ]);
PlanarEmbedding(g);
[ [ 3, 7, 4 ], [ 4, 8, 3 ], [ 5 ], [ 6 ], [ 8, 7 ], [ 7, 8 ], [  ], [  ] ]
------> [ [ 3, 7, 4 ], [ 4, 8, 3 ], [ 5, 1, 2 ], [ 6, 2, 1 ], [ 8, 7, 3 ], [ 7, 8, 4], [ 1, 5, 6], [ 2, 6, 5] ]
~~~

Note that the rotation system for a planar graph is not unique so the output could differ.

james-d-mitchell:
Thanks @MeikeWeiss I'll think about this, and see if we can't offer what you suggest?

MeikeWeiss:
I wanted to test the method with some examples to confirm this, but I pulled the new version and now I get the error message:
~~~
Error, Variable: 'DIGRAPH_NRADJACENCIES' must have a value
~~~
I'm not sure what the problem is.

[...]

PR #696, opened 2024-08-31
Fix planarity

Joseph-Edwards:
This PR addresses https://github.com/digraphs/Digraphs/issues/669.

For planarity checking, the `edge-addition-planarity-suite` code needs its graphs to be antisymmetric. Previously, the antisymmetric digraph was constructed in GAP, and then the desired output was constructed in C. This meant that the C code couldn't reason about the user's input digraph, and hence only a minimal rotation system was returned.

Now, the antisymmetric digraph is constructed in C, and the return list in constructed by checking the existence of edges in the user's original input graph.


Joseph-Edwards:
For some of the `SubdigraphHomeomorphicToK[...]` functions, I'm not 100% sure what the best output/documentation should be. At the moment, on input `D`, these functions return subdigraphs `G` such that, when directions are ignored, it is Homemorphic to K[...]. I don't know if this is desirable (or even the correct use of homeomorphic). For example:

``` GAP
gap> D := CompleteDigraph(4);
<immutable complete digraph with 4 vertices>
gap> SubdigraphHomeomorphicToK4(D);
[ [ 2, 4, 3 ], [ 3, 4, 1 ], [ 4, 2, 1 ], [ 1, 2, 3 ]
```

An alternative output could be:
``` GAP
[ [ 2, 4, 3 ], [ 3, 4 ], [ 4 ], [ ] ]
```

It is a question of whether the output should be thought of as a graph (because direction doesn't matter in the case of planarity) or a digraph (because we are in the digraphs package).

I'm happy to change the behaviour to whatever seems the most appropriate, or to provide some more examples.

gap/planar.gi
```diff
@@ -32,69 +32,90 @@
...
-  if IsEmptyDigraph(D) or DigraphNrVertices(D) < 3 then
-    return [];
-  elif HasIsPlanarDigraph(D) and not IsPlanarDigraph(D) then
+  local C;
+  C := DigraphMutableCopy(D);
+  DigraphRemoveAllMultipleEdges(C);
+  DigraphRemoveLoops(C);
+  if IsEmptyDigraph(C) or DigraphNrVertices(C) < 3 then
+    return OutNeighbors(C);
+  elif HasIsPlanarDigraph(C) and not IsPlanarDigraph(C) then
```
james-d-mitchell, line 41: This will always be false because C is mutable and hence Has no attributes. These conditions should probably be applied before C is created

Joseph-Edwards: Thanks for pointing this out; I've fixed this now.

gap/planar.gi
james-d-mitchell, line 1: As discussed, it seems that in `IsPlanarDigraph` we should do something like:

~~~
if IsDigraphWithNoLoops(D) and IsDigraphWithNoMultiEdges(D) and IsAsymmetricDigraph(D) then
   C := D;
else
   C := MutableCopyDigraph(D);
   DigraphRemoveLoops(C);
   DigraphRemoveAllMultiEdges(C);
   MaximalAntiSymmetricDigraph(C);
fi;

gap/attr.gi
```diff
@@ -701,6 +701,13 @@ function(D)
   return m;
 end);
 
+InstallMethod(DigraphNrAdjacencies, "for a digraph", [IsDigraphByOutNeighboursRep],
+function(D)
```
james-d-mitchell, line 705: You can replace these 5 lines by DIGRAPH_NRADJACENCIES

james-d-mitchell: Doc?

gap/attr.gi
```diff
@@ -701,6 +701,13 @@ function(D)
...
+  local m;
+  m := DIGRAPH_NRADJACENCIES(D);
+  return m;
+end);
```
james-d-mitchell, line 709:
```suggestion
DIGRAPH_NRADJACENCIES);
```

Joseph-Edwards:
Provided the tests pass, I think this is good to go @james-d-mitchell

Joseph-Edwards:
I've just noticed this:
```
gap> IsPlanarDigraph(HanoiGraph(6));
true
gap> IsPlanarDigraph(DigraphMutableCopy(HanoiGraph(6)));
false
gap> 
```

Not sure what's going on, but this probably shouldn't happen. This behaviour exists in the main branch as well. Trying to fix now.

Joseph-Edwards:
> I've just noticed this:
> 
> ```gap
> gap> IsPlanarDigraph(HanoiGraph(6));
> true
> gap> IsPlanarDigraph(DigraphMutableCopy(HanoiGraph(6)));
> false
> ```
> 
> Not sure what's going on, but this probably shouldn't happen. This behaviour exists in the main branch as well. Trying to fix now.

I think this is a problem with `HanoiDigraph` rather than the planarity stuff (see #698 ).

james-d-mitchell:
Hmm that seems fishy, the return value shouldn't depend on whether or not the graph is mutable.

Joseph-Edwards:
@james-d-mitchell `HanoiGraph` sets the property `IsPlanarDigraph` to `true`. My guess is that this value gets returned when dealing with the immutable one, but the planarity actually gets computed (and returns `false`) in the immutable case.

tst/standard/attr.tst
```diff
@@ -192,6 +192,12 @@ gap> AdjacencyMatrix(Digraph(rec(DigraphNrVertices := 0,
 >                                DigraphRange      := [])));
 [  ]
 
+#  DigraphNrAdjacencies
+gap> G := RandomDigraph(50);;
```
james-d-mitchell, line 196: Better not use random graphs in the tests, if this fails for some reason we won't have any idea what the input graph was

james-d-mitchell (approved):
One minor tweak and this is good to go

tst/standard/attr.tst
Joseph-Edwards: Is there a standard replacement for this, or should I just hardcode my own graph?

james-d-mitchell: Best to hard-code a return value of RandomDigraph(50) that you've found

Joseph-Edwards: I opted for `RandomDigraph(30)` so the input wasn't so unruly, but otherwise that's done now.

PR #702, opened 2024-09-10
Added copy in DigraphAllChordlessCycles

MeikeWeiss:
I had the problem that after calling `DigraphAllChordlessCycles` for a graph the output of `DigraphAllSimpleCircuits` for this graph was wrong. The output included a list of nodes that did not define a valid cycle. In my opinion the problem was that `SetDigraphVertexLabels(digraph, Reversed(DigraphDegeneracyOrdering(digraph)))` changed the given digraph. So now I added a call so that we only consider a copy of the input graph. 
This solved my problem, but I'm not sure this is the best way.

The graph I was having problems with was the following:
~~~
 g:=Digraph([ [ 2, 3, 7 ], [ 1, 4, 8 ], [ 1, 4, 9 ], [ 2, 3, 10 ], [ 6, 7, 9 ], [ 5, 8, 10 ], [ 8, 5, 1 ], [ 7, 6, 2 ], [ 5, 10, 3 ], [ 6, 9, 4 ] ]);
~~~


james-d-mitchell (changes requested):
Thanks for the PR @MeikeWeiss, please consider my comment, and let me know if that also works. If it does, then please add your previously failing example as a test case too.

MeikeWeiss:
Thank you for your comment! It seems to work.

james-d-mitchell (approved):

Issue #522, opened 2021-12-02
Implement `IsOrderIdeal` for a subset of a partial order digraph


james-d-mitchell:
This property would return `true` if the specified subset was "downwards" closed, i.e. contains every node less than the given nodes. This could more or less easily be done using `VerticesReachableFrom` if `VerticesReachableFrom` could take a list of nodes.

DanielPointon:
I'll be working on this following discussion in weekly meeting.

james-d-mitchell:
We could also add `IsOrderFilter` by replacing `downwards` with `upwards` in the original comment.

DanielPointon:
https://github.com/digraphs/Digraphs/pull/609

james-d-mitchell:
@ zljlzljlz Do you think you could try to implement a method for `IsOrderFilter` which would be almost identical to the method that Daniel did for `IsOrderIdeal`?

PR #707, opened 2024-10-09
Added IsOrderFilter

zljlzljlz:
#522 - returns true if the specified subset is upwards closed, false if not, based on IsOrderIdeal.


mtorpey:
I took a look at this and offered @zljlzljlz some pointers.  She made a lot of progress, which should hopefully be pushed soon.

james-d-mitchell:
Thanks both, I look forward to seeing this when pushed 👍

mtorpey:
It looks like `doc/title.xml` was added to the repo accidentally.  Could you remove this?

Otherwise, this is looking good – I'd recommend merging after that change.

james-d-mitchell:
Agreed I'll be happy to merge when the `title.xml` file is removed. Thanks @zljlzljlz !

mtorpey (approved):
Great

Issue #704, opened 2024-09-18
Wrong output `SubdigraphsMonomorphisms`


MeikeWeiss:
I have the graph
~~~
 d:=Digraph([ [ 2, 3, 4, 5 ], [ 1, 3, 4 ], [ 1, 2, 4, 5 ], [ 1, 2, 3, 5 ], [ 1, 3, 4 ] ]);
~~~
and I would like to compute all subgraphs isomorphic to K_{2,3} with `SubdigraphsMonomorphisms`. 
In my opinion the output here is wrong because the subgraph/transformation defined by `Transformation( [ 2, 5, 1, 3, 4 ] )` which is contained in the output of `MonomorphismsDigraphsRepresentatives(CompleteMultipartiteDigraph([2,3]),d)` is missing. This subgraph has a different edge set than the subgraphs defined by the returned transformations of ` SubdigraphsMonomorphisms(CompleteMultipartiteDigraph([2,3]),d)`.
~~~
gap> SubdigraphsMonomorphisms(CompleteMultipartiteDigraph([2,3]),d);
[ Transformation( [ 1, 3, 2 ] ), Transformation( [ 1, 4, 2, 3 ] ), Transformation( [ 3, 4, 2, 1 ] ) ]
gap> MonomorphismsDigraphsRepresentatives(CompleteMultipartiteDigraph([2,3]),d);
[ Transformation( [ 1, 3, 2 ] ), Transformation( [ 1, 3, 2, 5, 4 ] ), Transformation( [ 1, 4, 3, 2 ] ), Transformation( [ 2, 5, 1, 3, 4 ] ) ]
~~~

MeikeWeiss:
@james-d-mitchell Do you have any idea how to solve it?

james-d-mitchell:
Sorry @MeikeWeiss didnt have time to investigate yet, I'll try to look this afternoon

james-d-mitchell:
@MeikeWeiss I spent some time yesterday trying to figure out what's going on here, I can confirm that I can reproduce the bug. I'm not yet sure what is causing it, I'll investigate further today and hopefully find a fix.

james-d-mitchell:
Resolved in v1.10.0

PR #709, opened 2024-10-10
Fix issue 704

james-d-mitchell:
This PR fixes issue #704. It seems that the previously installed method for `SubdigraphsMonomorphisms` was just wrong (right in all the examples we checked but not all). This PR re-implements the method, this time hopefully correct. The performance is unfortunately worse than the previous incorrect method, I might be able to do something about that but it might be more or less complicated, so I thought I'd post this. 

@MeikeWeiss can you please take a look and confirm that this resolves your issue?


MeikeWeiss:
Thank you very much! This solves my problem and I have also tested several examples. Sorry it took so long, but while testing I found a bug in my code :)

MeikeWeiss (approved):

PR #716, opened 2024-12-06
Remove references to non-existing gap/doc.g

fingolfin:
Unfortunately this causes a warning when loading Digraphs in GAP 4.14.0

PR #717, opened 2024-12-13
Don't try to read doc.g, file was previously removed

stertooy:
The file `gap/doc.g` was removed in 32ed64a, but GAP still tries (and fails) to read it when Digraphs is loaded.

### DIFF
--- a/read.g
+++ b/read.g
@@ -22,8 +22,6 @@ BindGlobal("DIGRAPHS_NautyAvailable",
 
 Unbind(_NautyTracesInterfaceVersion);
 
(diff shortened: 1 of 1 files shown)


james-d-mitchell:
Thanks for the PR @stertooy, and sorry for the delay in responding, I've been unwell. This is a duplicate of #716, I'm afraid, so I'm not going to merge it, sorry! I do need to merge `stable-1.9` (where #716 was merged) into `main` though, so thanks for pointing this out.

stertooy:
Ah, I didn't notice it was fixed in another branch. I'll close this PR then :)

PR #723, opened 2025-02-05
Behaviour of DigraphRemoveEdge/DigraphRemoveEdges is inconsistent #260

saffronmciver:
- Edited functions for `DigraphRemoveEdge`, `DigraphRemoveEdges`, `DigraphReverseEdge`, `DigraphReverseEdges`, `DigraphRemoveVertices` and `QuotientGraph`.
- Checks added to see if any work needs to be done before making a copy.


saffronmciver:
`DigraphsTestsStandard` fails due to a test expecting output where the attributes of a graph are retained after the immutable call of `DigraphRemoveVertices`.

james-d-mitchell:
@saffronmciver if you pull from main and then rebase your branch then hopefully the lint job will pass

gap/oper.gi
```diff
@@ -166,7 +166,13 @@ end);
 
 InstallMethod(DigraphRemoveVertices, "for an immutable digraph and a list",
 [IsImmutableDigraph, IsList],
-{D, list} -> MakeImmutable(DigraphRemoveVertices(DigraphMutableCopy(D), list)));
+function(D, list)
+  if ForAll(list, IsPosInt) and
+      not (ForAny(list, x -> x <= DigraphNrVertices(D))) then
```
mtorpey, line 171: If the list has negative integers, then presumably the input digraph should still not be copied?

Instead of testing `IsPosInt` and then checking `<=`, would we be better to check `ForAny(list, x -> not x in DigraphVertices(D)`?  Then it's a single clean statement that catches everything.

gap/oper.gi
```diff
@@ -249,8 +259,19 @@ end);
 InstallMethod(DigraphRemoveEdge,
 "for a immutable digraph and two positive integers",
 [IsImmutableDigraph, IsPosInt, IsPosInt],
-{D, src, ran}
--> MakeImmutable(DigraphRemoveEdge(DigraphMutableCopy(D), src, ran)));
+function(D, src, ran)
+  local pos, DMutable;
+  DMutable := DigraphMutableCopy(D);
+  DIGRAPHS_CheckDigraphRemoveEdge(DMutable, src, ran);
```
mtorpey, line 265: Here we create a copy and then check it is valid.  Why not check the original graph *before* copying, so that extra work is not done?

gap/oper.gi
```diff
@@ -275,7 +302,32 @@ end);
 
 InstallMethod(DigraphRemoveEdges, "for an immutable digraph and a list",
 [IsImmutableDigraph, IsList],
-{D, edges} -> MakeImmutable(DigraphRemoveEdges(DigraphMutableCopy(D), edges)));
+function(D, edges)
+   local DMutableRemoved;
+   DMutableRemoved := DigraphRemoveEdges(DigraphMutableCopy(D), edges);
+
+   # If any edges have been removed
+   if DigraphNrEdges(D) <> DigraphNrEdges(DMutableRemoved) then
```
mtorpey, line 310: Again, could we somehow check this before copying, to avoid the extra work if it's not necessary?

I think you could check all the inputs with `IsDigraphEdge` rather than comparing numbers after a copy.

gap/oper.gi
```diff
@@ -334,7 +396,23 @@ end);
...
+  DMutable := DigraphMutableCopy(D);
+  for edge in E do
+    DigraphReverseEdge(DMutable, edge);
+    if edge[1] <> edge[2] then  # The edge could be reversed
+      changed := true;
+    fi;
+  od;
+
+  if changed = true then
+    return MakeImmutable(DMutable);
```
mtorpey, line 411: Again, could we figure out whether the condition is true *without* creating a copy first?

james-d-mitchell:
I'm not sure because I haven't looked at the or in detail but isn't this a backwards incompatible change? I.e it's not impossible that there's code in existence that relies on a copy being returned even if no changes are made? If so then we should probably not merge this into main just now, but maybe include stable-2.0 branch instead?

mtorpey:
@james-d-mitchell: that's correct.  This PR even involves changing the documentation to reflect the new behaviour, so it's a breaking change.

Do we have a branch for 2.0 yet?

mtorpey (commented):
@saffronmciver, you've fixed everything I asked for and this is looking great, except for one minor issue with an error message.  After it's changed, we should merge this (into something).

PR #724, opened 2025-02-05
Fix line lengths for gaplint v1.4

james-d-mitchell:
Apparently we had some lines of length 81 that weren't previously picked up because of an off by one error in gaplint.


wilfwilson (approved):
Shocking. An urgently needed PR, thank you.

Issue #711, opened 2024-10-20
Fix HTML entities on the Digraphs homepage


wilfwilson:
On the Digraphs homepage, <https://digraphs.github.io/Digraphs>, the entities `&Digraphs;` and `&GAP;` are being rendered as-is, not as we want them to be rendered:

<img width="629" alt="Screenshot 2024-10-20 at 11 56 15" src="https://github.com/user-attachments/assets/e9da9cc9-9d53-4667-8a63-74bd441d491e">

This should be fixed, either by just having them as plain text, or by doing something else to make them special.

forktree:
Apologies, I'm still working on this issue. My initial attempt borked the documentation page so I'm trying again, expanding the macro call.

Discussed with James at today's GAP meeting.

wilfwilson:
Actually I came up with a better solution. See #809.

PR #726, opened 2025-02-05
Expand named entity GAPDoc calls in Abstract(HTML)

forktree:
This is my second attempt at resolving [Issue 711](https://github.com/digraphs/Digraphs/issues/711).

Derived from XML, GAPDoc has macro substitution in the form of named entities. Calling a macro, e.g. &Digraphs;, will expand the html defined for the given named entity.
The intended behaviour of the calls is exhibited in Chapter 0 of the project manual [1].
More information about named entities can be found in Chapter 2.1-9 of the GAPDoc documentation manual [2].

AutoDoc doesn't seem to expand these macros. This results in the raw macro call being displayed in the places controlled by AutoDoc, namely the project homepage [3] and Digraphs' entry on the list of GAP packages [4].

Hence to prevent unnecessarily defining the Abstract twice, with different styling for GAPDoc and AutoDoc,
instead expand the macro in-place.

Note this means any future changes to the macro (named entity) will not propagate here,
e.g. if `<strong>` html tag is replaced with the italic `<i>` tag.
But changes to the `pkg` html class should propagate.

[1] -- https://digraphs.github.io/Digraphs/doc/chap0_mj.html
[2] -- https://docs.gap-system.org/pkg/gapdoc/doc/chap2.html
[3] -- https://digraphs.github.io/Digraphs/
[4] -- https://www.gap-system.org/packages/

### DIFF
--- a/PackageInfo.g
+++ b/PackageInfo.g
@@ -28,16 +28,21 @@ end);
 
 BindGlobal("_MakeAbstractFunction",
 function(OutputFormat)
-  local ValidOutputFormats, ErrorMsg, OutputType, AbstractRec, Abstract, line;
+  local ValidOutputFormats, ErrorMsg, OutputType, AbstractRec, AbstractParts,
+        line, Abstract;
+
+  # Validate the specified OutputFormat.
   ValidOutputFormats := ["HTML", "XML"];
   if not OutputFormat in ValidOutputFormats then
     ErrorMsg := Concatenation(["The given `OutputFormat` argument is invalid. ",
                                "Valid options are:\n"]);
     for OutputType in ValidOutputFormats do
-        ErrorMsg := Concatenation(ErrorMsg, "- ", OutputType, "\n");
+      ErrorMsg := Concatenation(ErrorMsg, "- ", OutputType, "\n");
     od;
     ErrorNoReturn(ErrorMsg);
   fi;
+
+  # Define the Abstract here.
   AbstractRec := [rec(str := "The", toReplace := false),
                   rec(str := "Digraphs", toReplace := true),
                   rec(str := "package is a", toReplace := false),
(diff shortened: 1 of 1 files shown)


forktree:
Upon testing, it seems the macros cannot be manually expanded in `AutoDoc.TitlePage.Abstract`. This yields the following XML parse error (see commit `a7c6741`): `Wrong element in Abstract: strong`.

I think this may be due to the fact that `<` is an illegal character in XML. One approach mentioned in the GAPDoc manual [1] is to wrap html tags within a CDATA block so it is ignored by the XML parser. I'm not sure if this would help us keep only a single representation, or if it would display the correct thing on each page -- will try this later.

Maybe storing a record with the separate XML and html representations, then loading it per string, would be a better solution?

The `GAPDoc2HTML` function seems like it would be ideal, calling it on the Abstract to generate AbstractHTML. But I can't find a way of extracting only the html markup without writing to external files.

Any thoughts would be appreciated.

[1] -- Chapter 2.1-12 https://docs.gap-system.org/pkg/gapdoc/doc/chap2.html

forktree:
I've implemented a solution that handles the respective markup for HTML and XML before the strings are passed to each of `AutoDoc.TitlePage.Abstract`, which expects XML/GAPDoc format, and `AbstractHTML`, which expects HTML.

If this is too much code, in commit `0688fa0` I manually fixed the markup in each instance.

james-d-mitchell:
Thanks for the PR @forktree, I like the idea of using a function to return the string. The actual function you've written is a bit over-complicated though, and will need to be reworked before we can merge it. I will add some more specific comments to the code itself.

forktree:
Thanks very much @james-d-mitchell, I am happy to make changes as you suggest. Would your suggestions come up in this window or elsewhere?

james-d-mitchell:
Let's try to resolve the issue that this PR solves at gap days.

james-d-mitchell:
I accidentally merged this PR, then dropped it again. Sorry!

PR #730, opened 2025-02-12
Renamed DigraphDijkstra to DigraphShortestPaths

devansh2605:
https://github.com/digraphs/Digraphs/issues/570

### DIFF
--- a/gap/oper.gi
+++ b/gap/oper.gi
@@ -1780,7 +1780,7 @@ function(digraph, source, target)
   while not IsEmpty(queue) do
     u := Pop(queue);
     u := u[2];
-    # TODO: this has a small performance impact for DigraphDijkstraS,
+    # TODO: this has a small performance impact for DigraphShortestPathsS,
     #       but do we care?
     if u = target then
       return [dist, prev];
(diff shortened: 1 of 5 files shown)


mtorpey:
This is looking good!  The only problem now (which I didn't mention before) is that we want to preserve backwards compatibility.  So, now that `DigraphDijkstra` is gone, we want to introduce `DigraphDijkstra` as a **synonym** for `DigraphShortestPaths`.  Look at how `DeclareSynonym` is used elsewhere in the package, and add new commits to this branch to add them to this PR.

james-d-mitchell:
Might even make sense to do something like we do is Semigroups and mark `DigraphDijkstra` as obsolete:

https://github.com/semigroups/Semigroups/blob/7cc3070a8fda8d9b866b5864357dff491c133eb9/gap/obsolete.gi#L11

devansh2605:
Currently, I have renamed `DigraphDijkstra` to `DigraphShortestPaths` and also declared it as a synonym using `DeclareSynonym`. Please let me know if I should proceed with https://github.com/digraphs/Digraphs/issues/735.

james-d-mitchell:
This is dependent on #742, and should be updated before being merged.

gap/oper.gi
```diff
@@ -1780,7 +1780,7 @@ function(digraph, source, target)
   while not IsEmpty(queue) do
     u := Pop(queue);
     u := u[2];
-    # TODO: this has a small performance impact for DigraphDijkstraS,
+    # TODO: this has a small performance impact for DigraphShortestPathsS,
```
mtorpey, line 1783: I think this might be a typo that you've preserved with find-and-replace!

PR #737, opened 2025-03-04
Attempt to add ExecuteDFS to main (Generic depth first search)

saffronmciver:
This pull request will build on #505 and #499 and attempt to add `ExecuteDFS` to the main branch again. 

My aim is to add back the fixed versions of functions using the generic DFS, adding more tests to ensure they are correct, and to look at improving the performance of these functions. 

I will also look at improving the performance of `ExecuteDFS` itself.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -26,6 +26,7 @@ KEXT_SOURCES += src/perms.c
 KEXT_SOURCES += src/planar.c
 KEXT_SOURCES += src/schreier-sims.c
 KEXT_SOURCES += src/safemalloc.c
+KEXT_SOURCES += src/dfs.c
 
 ifdef WITH_INCLUDED_BLISS
   KEXT_SOURCES += extern/bliss-0.73/defs.cc
--- a/gap/attr.gi
+++ b/gap/attr.gi
@@ -882,17 +882,17 @@ function(D)
   record := NewDFSRecord(D);
   count := 0;
   out := [];
-  PostOrderFunc := function(record, data)
+  PostOrderFunc := function(record, _)
     count := count + 1;
     out[count] := record.child;
   end;
-  AncestorFunc := function(record, data)
+  AncestorFunc := function(record, _)
     if record.current <> record.child then
       record.stop := true;
     fi;
   end;
   for i in DigraphVertices(D) do
-    if (record.preorder[i] <> -1) then
+    if (IsBound(record.preorder[i])) then
       continue;
     fi;
     ExecuteDFS(record,
--- a/gap/attr.gi
+++ b/gap/attr.gi
@@ -911,45 +911,6 @@ end);
 
 # InstallMethod(DigraphTopologicalSort, "for a digraph by out-neighbours",
 # [IsDigraphByOutNeighboursRep],
-# function(D)
-#   local N, record, count, out, PostOrderFunc, AncestorFunc, i;
-
-#   N := DigraphNrVertices(D);
-#   if N = 0 then
-#     return [];
-#   fi;
-#   record := NewDFSRecord(D);
-#   count := 0;
-#   out := [];
-#   PostOrderFunc := function(record, data)
-#     count := count + 1;
-#     out[count] := record.child;
-#   end;
-#   AncestorFunc := function(record, data)
-#     if record.current <> record.child then
-#       record.stop := true;
-#     fi;
-#   end;
-#   for i in DigraphVertices(D) do
-#     if (record.preorder[i] <> -1) then
-#       continue;
-#     fi;
-#     ExecuteDFS(record,
-#                fail,
-#                i,
-#                DFSDefault,
-#                PostOrderFunc,
-#                AncestorFunc,
-#                DFSDefault);
-#     if record.stop then
-#       return fail;
-#     fi;
-#   od;
-#   return out;
-# end);
-
-# InstallMethod(DigraphTopologicalSort, "for a digraph by out-neighbours",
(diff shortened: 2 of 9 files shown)


saffronmciver:
So far I have refactored the iterative solution of `ExecuteDFS` to a recursive one. The iterative stack can grow very large, since all successors are pushed to the stack if they have not been visited. if `current` is the node being visited and `successors := [1, 2]` is a list of successors to `current`, and both are unvisited, vertex `2` is pushed to the stack even if it would be visited in the branch where `1` is visited. The iterative solution checks if the `current` node has been visited and continues the loop if it has to deal with this. 

However, this has implications for `AncestorFunc` and `CrossFunc`. If `(current, 2)` is a back edge / cross edge, the relevant function is called before any calls to `PreorderFunc`, `PostorderFunc`, etc. are called by visiting the branch of `1`. With recursion, `AncestorFunc` or `CrossFunc` can easily be called after the recursive call exploring `1` is complete. 

I believe removing this behavior and using recursion would also allow forward edges to be recognised during the DFS if this is useful at all.

saffronmciver:
Below is an overview of the current state of generic DFS performance for selected examples. All benchmarking was done as an average of 100 runs. 

# General ExecuteDFS Performance

## Performance of `ExecuteDFS` on `CompleteDigraph`s

AverageExecuteDFS

The above graph shows the performance of `ExecuteDFS` on `CompleteDigraph`s of increasing size. Benchmarks include the creation of the `DFSRecord`, since this is always necessary in practice (and is what speeds up mostly when removing fields from the record).

An "all fields" lines refer to use of a `DFSRecord` with all of `use_preorder`, `use_postorder`, `use_edge` and `use_parents` being `true`, and "minimum fields" refers to the least of these being enabled as possible (for iterative, none, for recursive this is only `use_parents` and `use_edge`). Iterative with all fields enabled clearly performs worse, and there seems to be an increase in how much worse the "all fields" configurations perform for larger graphs. It is not completely clear whether iterative or recursive is better for "min fields". The recursive procedure is limited by needing to store `edge` and `parents` fields. Although, recursive DFS should have a lesser memory footprint (the stack for iterative DFS is unbounded).

# Examples for `CompleteGraphs`

A `CompleteGraph` is a particularly bad case for both procedures, since all successors that have not been visited are pushed to the stack in iterative DFS, and recursive DFS calls it's function for each successor. `DominatorTree` and `VerticesReachableFrom` appear to show consistent slowdown on size increase of the graph (more runs may be needed to investigate `DominatorTree` on larger graphs due to the increase when the size is 1500). The slowdown of `DominatorTree` and `VerticesReachableFrom` as compared to the currently implemented non generic DFS algorithms is shown below.

## Slowdown of `DominatorTree` and `VerticesReachableFrom` for `CompleteDigraph`s

DominatorTreeVerticesReachableFrom

`UndirectedSpanningForest` has a high slowdown factor for `CompleteDigraph`s. I am not sure whether replacing this algorithm with generic DFS is a good idea (but would like to further see why this occurs, since this was not the case for `DominatorTree` or `VerticesReachableFrom` when using inputs of `CompleteGraph`, although I should test with higher sizes of `CompleteGraph`s, but the runtimes were becoming unpractically large). 

## Slowdown of `UndirectedSpanningForest` for `CompleteDigraph`s

UndirectedSpanningForestComplete

## Slowdown of `UndirectedSpanningForest` for `BinaryTree`s

BinaryTree

The slowdown is much less significant for `BinaryTree`s which has less of the behaviour where a very large amount of calls 
are made for trying to visit vertices, or for iterative (currently being used for this procedure), there are less occurences of large quantities of repeated nodes being pushed to the stack. 

# Examples for `BinaryTrees`

The slowdown factors for all of `DigraphTopologicalSort`, `IsAntiSymmetricGraph` and `IsAcyclicGraph` is not significant as shown below (when called with `BinaryTree`). 

## Slowdown of `UndirectedSpanningForest`, `DigraphTopologicalSort`, `IsAntiSymmetricGraph` and `IsAcyclicGraph` for `BinaryTree`s

BinaryTree

# Examples for Modified `ChainGraph`s

For the following examples, I modified a `ChainDigraph` to concatenate 10 of itself into a single graph (so each graph has 10 `ChainDigraph`s with a root node of 1 that are all connected). This was to allow for an acyclic example that was not completely trivial. 

## Slowdown of `DigraphPath` and `LongestDistanceFromVertex` for modified `ChainDigraphs`

PathLongestDistance

The slowdown seems relatively consistent when increasing the size of this particular graph type for `DigraphPath` and `LongestDistanceFromVertex`

saffronmciver:
**Some other notes on the state of this procedure**

- Further investigation on whether using the recursive or iterative procedure is better might be useful (I think this can just depend on the type of graph, but I would like to understand why `CompleteDigraphs` are much worse for `UndirectedSpanningForest`). 
- I would like to add more tests to some implementations, I have checked through each one and tried to think of any ways they would not be correct, but I think some even more tests would be beneficial. 
- More tests should be added to ensure behaviour of `DFSRecord` configurations are correct (epecially combinations of them, and to check `ExecuteDFS` function behaviour is consistent across them when it should be). These were recently added so I have not got round to adding these tests.
- Making sure there are no other procedures existing that are currently using a non generic DFS procedure that would not perform too much worse with generic DFS (and that would be able to be implemented using it)
- Pushing backtrack nodes to the iterative stack is not necessary if no `PostOrderFunc` exists, and if `record.config.use_postorder` is false.
- If there is any way that the code can be made tidier while keeping the configuration options, this would also be good to change so it is more maintainable.

PR #738, opened 2025-03-26
Make random digraphs more random

flsmith:
The previous implementation of RandomDigraph was heavily biased towards empty and complete digraphs. This PR improves that (though there is still a bit of a bias for these, for reasons I haven't looked into yet).

When not specified, the edge-probability is now chosen as `Float(Random(0, n^2) / n^2)`. Maybe it would be a good idea to take a slightly higher resolution to avoid any strange resolution effects, but I don't know what these effects might be. This probability done for all types of (non-multi) digraphs, even when `n^2` isn't the correct maximum number of edges, since it shouldn't have much effect.

I haven't changed RandomMultiDigraph as its behaviour is different - it takes as second parameter a desired number of edges instead of an edge-probability (I guess it isn't clear what that would mean).

### DIFF
--- a/gap/digraph.gi
+++ b/gap/digraph.gi
@@ -1361,37 +1361,37 @@ end);
 InstallMethod(RandomDigraphCons, "for IsMutableDigraph and an integer",
 [IsMutableDigraph, IsInt],
 {_, n}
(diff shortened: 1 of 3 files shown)


james-d-mitchell:
Nice thanks @flsmith

PR #739, opened 2025-03-26
Is2EdgeTransitive overhall

frankiegillis:
Rewritten Is2EdgeTransitive to be much more efficient. Now uses Orbit-Stabiliser Theorem.


james-d-mitchell:
I think you're still working on this, is that right @frankiegillis ?

frankiegillis:
Yes, currently benchmarking the two methods for computing the orbit of a 2-edge: Method 1 (Compute the orbit directly) against Method 2 (Compute the stabiliser).

Here are some plots. To benchmark, I generated random digraphs using RandomDigraph using the edge probability as a way of controlling the average orbit length.
Image5
Image4
Image3
Image1 png

It turns out that Method 1 is almost always faster, apart from when tested on the list of named digraphs when it is much slower. I don't think this is a case of Method 2 being fast, but Method 1 being unusually slow. I'm looking into this.

frankiegillis:
A _2-edge_  in a digraph D is a triple (u, v, w) of distinct vertices such that (u, v) and (v, w) are (directed) edges. In what follows, I call the vertex v the _center_ of the 2-edge (u, v, w).

The idea is that if D is 2-edge transitive, then it must be transitive on 2-edge centers (since each 2-edge has a unique center). Therefore, every center must have the same in-degree and the same out-degree. The new algorithm first iterates through the vertices of D to form the list of 2-edge centers, returning false if any 2-edge centers have a different in-degree or out-degree. The number of 2-edges can then be calculated from knowing the in and out degree of a single vertex (minus loops, etc) and multiplying this by the number of 2-edge centers.

The orbit length of a single 2-edge is the calculated using the stabilizer method, as by this point we already know that D is regular on 2-edge centers, so is likely (though of course not guaranteed) to be highly symmetric. Testing on both random and named digraphs confirms the stabilizer method to be faster.

This new method has two distinct advantages: finding the number of 2-edges is now O(Vertices),  and the method may now return false before any orbit calculation occurs. It's also helpful to already have clues about the degree of symmetry of D and therefore know the stabilizer method is very likely to be much faster.

I also tested using _DigraphRemoveLoops_ versus removing the loops in place during the main for loop. At the moment, using _DigraphRemoveLoops_ turns out to be slightly faster. I think this is because it first checks _HasDigraphHasLoops_ and then _DigraphHasLoops_ so has the possibility of returning much faster.

frankiegillis:
Here is some evidence that this new method is faster than simply looping through all possible 2-edges
OldNew

wilfwilson:
This PR also names `IsTwoEdgeTransitive` to `Is2EdgeTransitive`. So either we should add a synonym, or this is a non-backwards-compatible breaking change that needs to be merged into the `dev-2.0` branch, as things stand.

frankiegillis:
This adds "IsTwoEdgeTransitive" as a synonym for Is2EdgeTransitive.

doc/prop.xml.save
james-d-mitchell, line 1: I think you've included this file by mistake.

frankiegillis:
I've removed the spurious files and updated the docs.

Issue #735, opened 2025-02-19
System for marking obsolete methods


mtorpey:
As mentioned in #730 we want to be able to mark something in the main branch as obsolete/deprecated, so it should print a warning that it will be removed soon.

We could do something like Semigroups with its obsolete.gi file, but we should think about how to do this in a nice way with synonyms.

mtorpey:
This should probably be a function called something like `DIGRAPHS_DeclareObsoleteSynonym` that installs a new global function with the desired name that prints a warning and then simply calls the existing thing.

mtorpey:
@devansh2605 is going to ty this!

devansh2605:
@mtorpey @james-d-mitchell This is also complete can we pls move this to under review?

PR #742, opened 2025-04-02
Add InstallDeprecatedMethod

devansh2605:
https://github.com/digraphs/Digraphs/issues/735

### DIFF
--- a/gap/obsolete.gd
+++ b/gap/obsolete.gd
@@ -1,0 +1,106 @@
+GetFunctionName := function(func)
+  local names, name;
+  names := Names(GLOBAL);
+  for name in names do
+    if IsBound(name) and Eval(name) = func then
+      return name;
+    fi;
+  od;
+  return func+ "cannot be found";
+end;
+
+DeclareDeprecatedSynonym := function(oldName, newFunc)
+  local newName, warningMsg;
+  
+  newName := GetFunctionName(newFunc);
+  warningMsg := Concatenation(oldName, " is deprecated. Instead use ", newName, ".");
+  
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
Thanks @devansh2605, can you please fix the linting, and then we can proceed to a proper review.

gap/obsolete.gd
```diff
@@ -0,0 +1,55 @@
...
+      Warning(warningMsg);
+      return func(a, b, c, d, e);
+    end;
+  elif args = 6 then
+    newMethod := function(a, b, c, d, e, f)
+      Warning(warningMsg);
+      return func(a, b, c, d, e, f);
+    end;
+  else
+    Error("Unsupported args: ", args);
```
james-d-mitchell, line 42: You could just do:

~~~
newMethod := function(args...)
  Warning(warningMsg);
   return CallFuncList(fun, args);
end;

james-d-mitchell: Ah this doesn't work, hmmm

devansh2605: @james-d-mitchell sir, should I update this to your suggestion or try an alternative route? Or is the current implementation correct?

devansh2605:
@james-d-mitchell I have fixed the linting as well, pls let me know if any other modifications are required.

mtorpey:
I've had a look at this – looks like you've put some work in, but at the moment there are quite a few syntax errors and it's not loading up.  Why don't we shelve this for now while you work on other issues? I'll leave the issue open.

PR #743, opened 2025-04-02
Use `gap-actions/run-pkg-tests`

Joseph-Edwards:
In #666, we added a custom action in place of https://github.com/gap-actions/run-pkg-tests that treats warnings as errors and makes running Valgrind easier. As of https://github.com/gap-actions/run-pkg-tests/pull/35, this is now integrated into the gap action.

This PR reuses the gap action.


Joseph-Edwards:
The Valgrind action hasn't been triggered, so I am testing it here:
https://github.com/Joseph-Edwards/Digraphs/actions/runs/14223354028/job/39856753182

james-d-mitchell:
Thanks @Joseph-Edwards is this ready? Why did the job not trigger?

Joseph-Edwards:
Yeah this is ready to go. It seems like the Valgrind job only gets triggered on releases or when the C code gets edited, so it's non-run was to be expected.

PR #749, opened 2025-04-07
Fix badges

BaseMax:


BaseMax:
cc https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/monitoring-workflows/adding-a-workflow-status-badge

PR #753, opened 2025-04-08
Update to Planarity v4.0.0.0

### LINKED ISSUE #741: planarity-4.x compatibility
There's a new version of planarity out:

* https://github.com/graph-algorithms/edge-addition-planarity-suite/releases/tag/Version_4.0.0.0

The headers have been moved around a bit, causing the digraphs build to fail when an external libplanarity (4.x) is used. Here's the lazy version of the patch:

```patch
--- a/src/planar.c
+++ b/src/planar.c
@@ -38,10 +38,7 @@
 #include "c/graphK33Search.h"
 #include "c/graphK4Search.h"
 #else
-#include "planarity/graph.h"
-#include "planarity/graphK23Search.h"
-#include "planarity/graphK33Search.h"
-#include "planarity/graphK4Search.h"
+#include "planarity/graphLib.h"
 #endif
 #if defined(__clang__)
 #pragma clang diagnostic pop
```

No other changes are required though, so it should be possible to support both versions by trying both sets of headers.

james-d-mitchell:
Resolves #741

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -45,28 +45,37 @@ endif
 
 
 ifdef WITH_INCLUDED_PLANARITY
-  KEXT_SOURCES += $(PLANARITY_SUITE_DIR)/c/graphDFSUtils.c
-  KEXT_SOURCES += $(PLANARITY_SUITE_DIR)/c/graphDrawPlanar.c
-  KEXT_SOURCES += $(PLANARITY_SUITE_DIR)/c/graphDrawPlanar_Extensions.c
-  KEXT_SOURCES += $(PLANARITY_SUITE_DIR)/c/graphEmbed.c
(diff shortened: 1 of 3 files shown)


james-d-mitchell:
This will fail in jobs with `--with-external-planarity` until:

https://github.com/conda-forge/planarity-feedstock/pull/21

or a successor is merged (i.e. until the latest version of libplanarity is available via conda).

james-d-mitchell:
Seems like maybe I introduced some bugs in this PR that are also causing the cygwin job to fail. Probably the best approach is to wait until my changes to libplanarity (as per 
https://github.com/graph-algorithms/edge-addition-planarity-suite/issues/147) are released

james-d-mitchell:
Superseded by #766

PR #755, opened 2025-04-08
Edge weights #6: RandomUniqueEdgeWeightedDigraph

mtorpey:
The next one of @RaiyanC's edge-weighted algorithms, adding randomly generated edge-weighted digraphs.

### DIFF
--- a/doc/weights.xml
+++ b/doc/weights.xml
@@ -271,3 +271,27 @@ gap> Sum(flow[1]);
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="RandomUniqueEdgeWeightedDigraph">
+<ManSection>
+  <Oper Name="RandomUniqueEdgeWeightedDigraph" Arg="[filt, ]n[, p]"/>
+  <Returns>An edge-weighted digraph.</Returns>
+  <Description>
+    This operation returns a random edge-weighted digraph based on the provided
+    arguments. <P/>
+
+    Its behaviour is the same as that of <Ref Oper="RandomDigraph"/> but the
+    returned digraph will additionally have the <Ref Attr="EdgeWeights"/>
+    attribute set populated with random unique weights from the set
+    <C>[1 .. m]</C> where <C>m</C> is the number of edges in the digraph.
+
+    <Log><![CDATA[
+gap> RandomUniqueEdgeWeightedDigraph(5);
+<immutable digraph with 5 vertices, 21 edges>
+gap> RandomUniqueEdgeWeightedDigraph(5, 1 / 2);
+<immutable digraph with 5 vertices, 14 edges>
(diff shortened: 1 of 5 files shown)


gap/weights.gi
```diff
@@ -780,3 +780,52 @@ function(digraph, start, destination)
 
   return flows;
 end);
+
+#############################################################################
+# 6. Random edge weighted digraphs
+#############################################################################
+
+DIGRAPHS_RandomUniqueWeights := function(digraph)
```
james-d-mitchell, line 788: shouldn't this function and the next one be bound globally?

gap/weights.gi
```diff
@@ -780,3 +780,52 @@ function(digraph, start, destination)
...
+  local outs, weightsSource;
+  outs := OutNeighbours(digraph);
+
+  # Unique weights are taken randomly from [1..nredges]
+  weightsSource := Shuffle([1 .. DigraphNrEdges(digraph)]);
+
+  # Assign the weights to the edges
+  return List(DigraphVertices(digraph),
+              u -> List(outs[u], _ -> Remove(weightsSource)));
```
james-d-mitchell, line 797: Interesting, does this work? If so, then nice!

james-d-mitchell (changes requested):
Looks good, mod the changes suggested.

gap/weights.gi
mtorpey: Yep, I was pleased with how concise I'd got that. `Remove` acts as a pop function, which allows for this one-liner.

doc/weights.xml
```diff
@@ -271,3 +271,27 @@ gap> Sum(flow[1]);
...
+<ManSection>
+  <Oper Name="RandomUniqueEdgeWeightedDigraph" Arg="[filt, ]n[, p]"/>
+  <Returns>An edge-weighted digraph.</Returns>
+  <Description>
+    This operation returns a random edge-weighted digraph based on the provided
+    arguments. <P/>
+
+    Its behaviour is the same as that of <Ref Oper="RandomDigraph"/> but the
+    returned digraph will additionally have the <Ref Attr="EdgeWeights"/>
+    attribute set populated with random unique weights from the set
```
james-d-mitchell, line 285:
```suggestion
    attribute populated with random unique weights from the set
```

gap/weights.gi
```diff
@@ -780,3 +780,50 @@ function(digraph, start, destination)
 
   return flows;
 end);
+
+#############################################################################
+# 6. Random edge weighted digraphs
+#############################################################################
+
+BindGlobal("DIGRAPHS_RandomEdgeWeightedDigraphFilt",
```
james-d-mitchell, line 788: I still think these two functions aren't necessary, and you're really just installing two methods in a more complicated way. Please look at `RandomDigraph` and implement this in the same way.

PR #761, opened 2025-05-07
Digraph distance attributes

weremczuk:
Following my comparison of features in Sage vs. GAP, I added functions computing the radius, periphery, and centre of a strongly connected digraph (DigraphRadius, DiagraphPeriphery, DigraphCentre).
Could be merged with DiagraphDiameter.

### DIFF
--- a/gap/attr.gi
+++ b/gap/attr.gi
@@ -3455,7 +3455,7 @@ InstallMethod(DigraphRadius, "for a digraph",
 [IsDigraph],
 D -> DigraphDistanceMetrics(D).Radius);
 
-InstallMethod(DigraphCentre, "for a digraph"
+InstallMethod(MyDigraphCentre, "for a digraph"
 [IsDigraph],
 D -> DigraphDistanceMetrics(D).Centre);
 
--- a/gap/attr.gi
+++ b/gap/attr.gi
@@ -3435,22 +3435,21 @@ D -> DIGRAPHS_IsMeetSemilatticeAndMeetTable(D)[2]);
 InstallMethod(DigraphDistanceMetrics, "for a digraph",
 [IsDigraph],
 function(G)
-  local ecc, v;
+  local ecc, v, n;
   if not IsDigraph(G) then
     Error("Input must be a digraph");
   elif not IsStronglyConnectedDigraph(G) then
     fail; # Only defined for strongly connected digraphs
   fi;
   ecc := [];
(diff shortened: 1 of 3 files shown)


gap/attr.gi
```diff
@@ -3431,3 +3431,42 @@ D -> DIGRAPHS_IsJoinSemilatticeAndJoinTable(D)[2]);
...
+  local ecc, c, u, v;
+  if not IsDigraph(G) then
+    Error("Input must be a digraph");
+  elif not IsStronglyConnectedDigraph(G) then
+    Error("Input digraph is not strongly connected; property undefined");
+  fi;
+  ecc := [];
+  for u in [1 .. DigraphNrVertices(G)] do
+    c := 0;
+    for v in [1 .. DigraphNrVertices(G)] do
```
james-d-mitchell, line 3447:
```suggestion
    for v in Vertices(G) do
```

gap/attr.gi
```diff
@@ -3431,3 +3431,42 @@ D -> DIGRAPHS_IsJoinSemilatticeAndJoinTable(D)[2]);
...
+      if u <> v then
+        c := Maximum(c, DigraphShortestDistance(G, u, v));
+      fi;
+    od;
+    Add(ecc, c);
+  od;
+  return rec(
```
james-d-mitchell, line 3454: I'm not 100% sure but I think the following is not particularly efficient because the computation of `Minimum(ecc)` and `Maximum(ecc)` will be repeated for each distinct value of `i` which is not such a good idea. Better to compute these values prior to this return statement.

james-d-mitchell:
This PR is also missing any documentation at all, which makes it not mergeable at the moment.

gap/attr.gi
```diff
@@ -3431,3 +3431,42 @@ D -> DIGRAPHS_IsJoinSemilatticeAndJoinTable(D)[2]);
...
 D -> DIGRAPHS_IsMeetSemilatticeAndMeetTable(D)[2]);
+
+InstallMethod(DigraphDistanceMetrics, "for a digraph",
+[IsDigraph],
+function(G)
+  local ecc, c, u, v;
+  if not IsDigraph(G) then
+    Error("Input must be a digraph");
+  elif not IsStronglyConnectedDigraph(G) then
+    Error("Input digraph is not strongly connected; property undefined");
```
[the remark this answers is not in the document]
wilfwilson: And please make a similar change for the error message a few lines above.

PR #766, opened 2025-07-03
build: support for EAPS 4

jengelh:
all I wanted to do is edit src/planar.c, but then found a vendored copy :-/
Anyway here that is.


james-d-mitchell:
Thanks for the PR @jengelh, if this works, then it will supersede #753. 

Seems like the CI has been disabled because it wasn't used for 60 days, I'll close and then re-open the PR so that all CI jobs are run.

james-d-mitchell:
I think a new version of EAPS is required to fix the reason that this fails to compile. We made some steps in this direction in:

https://github.com/graph-algorithms/edge-addition-planarity-suite/issues/147

but I didn't have time to get back to it. Maybe I'll have a go this week, time permitting

james-d-mitchell:
> >a new version of EAPS is required to fix the reason that this fails to compile
> 
> 
> 
> that's why the PR updates the bundled EAPS to 4 as well. where is the problem with that?

There's no problem with your pr, the problem is with EAPS which I don't think compiles with gcc-15, and the developers of EAPS need to make a new release to resolve this issue AFAIK (this was the case in April when I last looked at this in detail, and I don't think there was another release of EAPS in the mean time).

There're further issues with EAPS which are not related to digraphs or gcc-15 at all, for example some files are missing from the archive that prevent the conda package being updated to v4 also.

jengelh:
openSUSE uses gcc-15 and there have been no compile errors for EAPS 4.0.0.0.

jengelh:
Hadn't noticed because of

>Seems like the CI has been disabled (originally)

new commits are a-comin

james-d-mitchell:
Hi @john-boyer-phd, thanks for following up on this! @jengelh resolved this by:

1. putting all the includes from edge-addition-planarity-suite into a single header in Digraphs;
2. explicitly suppressing the warnings/errors in edge-addition-planarity-suite that were causing Digraphs to not compile, 
3. only including the file added in 1 elsewhere in the Digraphs kernel module. 

We've used this pattern in other projects when there are warnings emanating from included projects, this makes Digraphs independent of edge-addition-planarity-suite in this regard. Still be best if the warnings/errors could be resolved there too however. I hope this helps!

john-boyer-phd:
> Hi @john-boyer-phd, thanks for following up on this! @jengelh resolved this by:
> 
> 1. putting all the includes from edge-addition-planarity-suite into a single header in Digraphs;
> 2. explicitly suppressing the warnings/errors in edge-addition-planarity-suite that were causing Digraphs to not compile,
> 3. only including the file added in 1 elsewhere in the Digraphs kernel module.
> 
> We've used this pattern in other projects when there are warnings emanating from included projects, this makes Digraphs independent of edge-addition-planarity-suite in this regard. Still be best if the warnings/errors could be resolved there too however. I hope this helps!

Yes, it does, thank you @james-d-mitchell  
And yes, the warnings will still be resolved in EAPS. I did missingprototypes last night and will continue on until all are resolved.

PR #521, opened 2021-11-29
Implement DigraphIsKing and Digraph2Kings

hrj4:

PR #774, opened 2025-08-25
Remove from `.dirstamp` files

wilfwilson:
I noticed that #521 removed these two files. @james-d-mitchell says they should indeed not be present in the distro. But I figured it's best to have it as a separate PR because I like making PRs and I haven't made one for a while.


wilfwilson:
Thanks Wilf, beautiful.

PR #776, opened 2025-08-26
Improve some planarity-related manual entries

wilfwilson:
The documentation for `DualPlanarGraph` was a little non-standard, and also didn't explicitly state that it relies upon the edge-addition-planarity-suite. So I changed it.


james-d-mitchell (approved):
One super minor change, then happy to merge!

PR #781, opened 2025-08-27
Thank Leonard in the acknowledgements section of the manual

wilfwilson:
I noticed that Leonard isn't included as a contributor to the package, so I thought it would be appropriate to acknowledge him in the manual.


james-d-mitchell:
Great, thanks! @wilfwilson !

PR #782, opened 2025-08-27
Update for v1.10.1 release

wilfwilson:


wilfwilson:
The CI job to compile and upload the manual is taking forever (before it even gets to anything Digraphs-related) so I'm gonna just merge for now and be done with it.

PR #786, opened 2025-08-29
Add `GitHubUsername` to authors in the `PackageInfo.g` file

wilfwilson:
I don't know if these usernames are used anywhere, but `GitHubUsername` seems to be a used by @fingolfin in quite a lot of packages.


fingolfin:
See https://github.com/gap-system/gap/issues/4784 for background

PR #864, opened 2025-10-01
Added SwapDigraphs and DigraphsRemoveAllEdges functions

RheyaM:
Created SwapDigraphs to swap the contents of 2 digraphs, and added DigraphsRemoveAllEdges for removing all edges from a digraph


mtorpey (dismissed):
Great work, especially for your first pull request. I've suggested a load of changes (we always do this) so it'd be great if you could make those changes and we can get this merged. Sorry for the major changes I'm requesting to the core algorithm!

gap/attr.gi
```diff
@@ -2537,6 +2537,26 @@ function(D)
   return D;
 end);
 
+InstallMethodThatReturnsDigraph(DigraphRemoveAllEdges, "for a digraph",
+[IsDigraph],
+function(D)
+  local i, nbs, currentEdges, emptynbs;
+  nbs := OutNeighboursMutableCopy(D);
+  if IsImmutableDigraph(D) then
+    emptynbs := [];
```
james-d-mitchell, line 2546: This whole clause can be replaced with `NullDigraph(DigraphNrVertices(D))`

gap/attr.gi
```diff
@@ -2537,6 +2537,26 @@ function(D)
...
+    for i in [1 .. Length(nbs)] do
+      Add(emptynbs, []);
+    od;
+    return Digraph(IsImmutableDigraph, emptynbs);
+  else
+    for i in [1 .. Length(nbs)] do
```
james-d-mitchell, line 2552: I think this is probably not as efficient as it could be either

james-d-mitchell (changes requested):
Excellent start, but this Pr will need some changes before it can be merged, let's discuss tomorrow!

gap/oper.gi
```diff
@@ -989,6 +1006,47 @@ function(D1, D2, edge_function)
...
+"for a digraph and digraph",
+[IsDigraph, IsDigraph],
+function(D1, D2)
+  local nb1, nb2, i;
+
+  if IsMutableDigraph(D1) and IsMutableDigraph(D2) then
+    nb1 := OutNeighboursMutableCopy(D1);
+    nb2 := OutNeighboursMutableCopy(D2);
+    DigraphRemoveAllEdges(D1);
+    DigraphRemoveAllEdges(D2);
```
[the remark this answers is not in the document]
mtorpey: As shown here, I'd suggest using `OutNeighbours` instead of `OutNeighboursMutableCopy`.

gap/oper.gi
```diff
@@ -989,6 +989,17 @@ function(D1, D2, edge_function)
   return Digraph(edges);
 end);
 
+InstallMethod(SwapDigraphs,
+"for a digraph and digraph",
+[IsMutableDigraph, IsMutableDigraph],
+function(D1, D2)
+  local nb1, nb2;
+    nb1 := OutNeighboursMutableCopy(D1);
```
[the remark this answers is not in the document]
mtorpey: Indeed, I think `OutNeighbours` should be sufficient. Can talk today!

gap/attr.gd
```diff
@@ -89,7 +89,8 @@ DeclareAttributeThatReturnsDigraph("DigraphReverse", IsDigraph);
 DeclareAttributeThatReturnsDigraph("DigraphDual", IsDigraph);
 DeclareAttributeThatReturnsDigraph("ReducedDigraph", IsDigraph);
 DeclareAttributeThatReturnsDigraph("DigraphRemoveAllMultipleEdges", IsDigraph);
-DeclareAttributeThatReturnsDigraph("DigraphRemoveAllEdges", IsDigraph);
+DeclareAttributeThatReturnsDigraph("DigraphRemoveAllEdges", IsImmutableDigraph);
+DeclareAttributeThatReturnsDigraph("DigraphRemoveAllEdges", IsMutableDigraph);
```
mtorpey, line 93: Actually just one attribute is better here, with `IsDigraph`. Then the different methods are installed for more specific filters

PR #873, opened 2025-10-20
ViewString for edge weights

ap4108735:
Added a section to ViewString (line 643) which tests if the attribute EdgeWeights exists and updates the string appropriately

### DIFF
--- a/gap/digraph.gi
+++ b/gap/digraph.gi
@@ -681,7 +681,6 @@ function(D)
       fi;
     fi;
   fi;
-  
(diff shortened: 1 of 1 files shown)


gap/digraph.gi
```diff
@@ -681,6 +682,10 @@ function(D)
       fi;
     fi;
   fi;
+  
+  if HasEdgeWeights(D) = true then
```
james-d-mitchell: The = true is redundant here

Issue #867, opened 2025-10-01
Implement `DigraphMinimumCutSet`


reiniscirpons:
Let $D$ be an edge weighted digraph and $s, t$ be two vertices. A cut-set with source $s$ and target $t$ is, in some sense, a set of edges whose removal disconnects $s$ from $t$ in $D$ (this is precisely the case when $D$ is a symmetric edge-weighted digraph). See the Cuts definition from Wikipedia https://en.wikipedia.org/wiki/Max-flow_min-cut_theorem#Cuts for the full definition for a digraph. A minimum cut-set with source $s$ and target $t$ is a cut set whose sum of edge-weights is as small as possible.

It would be nice to have a function `DigraphMinimumCutSet` such that `DigraphMinimumCutSet(D, s, t)` returns the minimum cut set with source $s$ and target $t$ in the weighted digraph $D$. PR #584 contains some commented out code for for this, but I think we should be able to compute `DigraphMinimumCutSet(D, s, t)` from the flow computed by `DigraphMaximumFlow(D, s, t)` (in digraphs since PR #751) due to the max-flow min-cut theorem (see https://en.wikipedia.org/wiki/Max-flow_min-cut_theorem#Proof for a proof and the rest of the article for the statement of equivalence between the minimum cut and maximum flow).

frankiegillis:
I may look at this if no-one else is currently.

reiniscirpons:
Go for it! Let me know if you have any questions.

PR #879, opened 2025-10-31
Added `DigraphMinimumCutSet`

frankiegillis:
This adds the function `DigraphMinimumCutSet` using the `DigraphMaximumFlow` method and the max-flow min-cut theorem. See issue #867.

### DIFF
--- a/doc/weights.xml
+++ b/doc/weights.xml
@@ -285,7 +285,7 @@ gap> Sum(flow[1]);
     such that <A>s</A> is in <M>S</M> and <A>t</A> is in <M>T</M>. The <E>capacity</E>
     of an <M>s</M>-<M>t</M> cut is the sum of the weights of every edge whose source
     is in <M>S</M> and whose range is in <M>T</M>. The minimum <M>s</M>-<M>t</M> cut is
-    the <M>s</M>-<M>t</M> cut whose capacity is minimal.<P/>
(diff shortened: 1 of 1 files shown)


doc/weights.xml
```diff
@@ -272,6 +272,34 @@ gap> Sum(flow[1]);
 </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="DigraphMinimumCut">
+<ManSection>
+  <Attr Name="DigraphMinimumCut" Arg="digraph, s, t"/>
```
reiniscirpons: Same comment about name matching doc as above

gap/weights.gi
```diff
@@ -772,6 +772,54 @@ function(D, start, destination)
...
+function(D, s, t)
+  local weights, outs, vertices, flow, residuals, G, o, u, v, S, T;
+
+  # Extract important data
+  weights := EdgeWeights(D);
+  outs := OutNeighbours(D);
+  vertices := DigraphVertices(D);
+
+  # Check input
+  if not s in vertices then
```
reiniscirpons: Minor optimization: Because digraphs must be homogenous (all vertices are in the range `[1 .. DigraphNrVertices(D)]` with no holes), it might be better performance to check if `s >= 1` and `s <= DigraphNrVertices(D)`.

Issue #759, opened 2025-04-24
DotColoredDigraph documentation is broken


mtorpey:
I was looking at the documentation for `DotDigraph`, which includes variants such as `DotColoredDigraph`, and the examples included are problematic.  For some reason they're a `<Log>` instead of `<Example>`, so they're not executed by the automated tests – running them manually results in an error.

They're also weirdly formatted, with several graphs declared in a row without being used, they're extremely lengthy and verbose, and there are too many of them.

I might fix this soon if no one else does, since it's required for `DotColoredEdgeWeightedDigraph` (#746). But it occurs to me that it's quite a beginner-friendly task in case any newcomers want to try it.

coraaked:
I'd be keen to have a look at this!

PR #880, opened 2025-11-05
Fix dotdigraph doc #759

coraaked:
I have made some changes to the DotDigraph documentation page to include more relevant examples and to reduce verbosity. Such changes made include:

- Changing all instances of <Log> to <Example>
- Removing and merging duplicated and vacuous examples, such as graphs defined repeatedly and examples which I believe don't add much value.
- Adding examples of commands defined with no examples originally available.
- Making the coloring of edges and vertices more efficient
- Fixing minor typos, particularly in later examples.

If there are any issues or any way in which this can be improved, please do let me know.

### DIFF
--- a/doc/display.xml
+++ b/doc/display.xml
@@ -91,9 +91,9 @@
 
       This function was written by Attila Egri-Nagy and Manuel Delgado with some
       minor changes by J. D. Mitchell.<P/>
-      <Example><![CDATA[
+      <Log><![CDATA[
 gap> Splash(DotDigraph(RandomDigraph(4)));
-]]></Example>
+]]></Log>
     </Description>
   </ManSection>
 <#/GAPDoc>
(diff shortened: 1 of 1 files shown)


doc/display.xml
```diff
@@ -91,9 +91,9 @@
 
       This function was written by Attila Egri-Nagy and Manuel Delgado with some
       minor changes by J. D. Mitchell.<P/>
-      <Log><![CDATA[
+      <Example><![CDATA[
 gap> Splash(DotDigraph(RandomDigraph(4)));
-]]></Log>
+]]></Example>
```
mtorpey: Actually I think this one needs to stay as `Log` instead of `Example`. This is intended to open a PDF viewer in the desktop environment, so it's not suitable for automated testing (and it's causing the `dot: not found` error in the CI).

mtorpey (approved):
Looks like an improvement to me!

PR #884, opened 2025-11-21
Implementing DigraphEdgeConnectivity

RheyaM:
Implemented two methods for calculating the EdgeConnectivity of a Digraph, one using Spanning Trees and one using Dominating Sets, based on the Algorithms detailed in: https://www.cse.msu.edu/~cse835/Papers/Graph_connectivity_revised.pdf

Additional functions: DigraphDominatingSet() and DigraphGetNeighbourhood(), for getting a Dominating Set of a digraph and getting the neighbourhood of a subset of vertices in a digraph, respectively.


doc/oper.xml
```diff
@@ -1954,6 +1954,46 @@ rec( idom := [ 2, fail, 2, 2, 2 ], preorder := [ 2, 1, 3, 4, 5 ] )
...
 <#/GAPDoc>
 
+<#GAPDoc Label="DigraphDominatingSet">
+<ManSection>
+  <Oper Name="DigraphDominatingSet" Arg="digraph"/>
+  <Returns>A List</Returns>
+  <Description>
+      This creates a dominating set of <A>digraph</A>, which is a subset of the vertices in 
+      <A>digraph</A> such that the neighbourhood of this subset of vertices is equal to all 
+      the vertices in <A>digraph</A>. </P>
```
reiniscirpons:
```suggestion
      This function returns a <E>dominating set</E> of <A>digraph</A>, 
      which is a subset of vertices whose neighbourhood consists of all 
      vertices in <A>digraph</A>. <P/>
```
Some suggestions for rephrasing the sentence a bit. I think generally the first sentence of a function description should concisely describe what the function does in terms of what does the function expect of its arguments and what does the function return. Saying it "creates" something is not bad, but its maybe slightly unclear whether this is created internally and stored somewhere or returned by the function directly etc. Also the new paragraph attribute should be `<P/>`.

I think you should also clarify what a neighbourhood of a set of vertices is, this is not immediately clear. It also depends for a directed graph - you have the neighbourhood of vertices reachable by out-edges and a different neighbourhood reachable by in-edges. As a rule of thumb, when writing, imagine you are writing the doc to explain it to yourself in the past, before you ever undertook this project.

gap/oper.gi
```diff
@@ -2522,6 +2522,45 @@ function(D, root)
   return result;
 end);
 
+# For calculating a dominating set for a digraph
+# Algorithm 7 in :
+# https://www.cse.msu.edu/~cse835/Papers/Graph_connectivity_revised.pdf
+InstallMethod(DigraphDominatingSet, "for a digraph",
```
reiniscirpons: Does the method work for all digraphs or only symmetric digraphs? If its only intended for symmetric digraphs, then this should be explicitly checked for in the function and mentioned in the documentation.

If its meant to work for all digraphs, it matters whether the dominating set is with respect to in-neighbours or out-neighbours, so the function name should be updated (I think the currently implemented function returns the out-dominating set), say `DigraphOutDominatingSet`, and a dual function for in-neighbours should be implemented.

Alternatively, the function could also take the symmetric closure of the digraph, this way it would be a dominating set with respect the notion of undirected adjacency (this might be called a weak dominating set for a directed graph? Not sure).

reiniscirpons (changes requested):
Thanks for the implementation! I had some suggestions for the documentation and for improving the performance of the methods. Mainly more details in the docs and reworking some loops to implement more efficient DFS.

Some other things:
* The documentation is not currently linked in the manual. You should edit the correct `doc/z-chapX.xml` file to add this. For example, to get the `DigraphDominatingSet` documentation to show up you should add the `<#Include Label="DigraphDominatingSet">` tag in the correct place of `doc/z-chap4.xml`. You can test this by seeing if `?DigraphDominatingSet` shows up the docs after running `DigraphsMakeDoc()` in `gap`. Same applies for the other functions.
* Functions need to be more careful about symmetric vs non symmetric digraphs.
* You should re-run the benchmarks after making the performance based changes to the implementation. I think currently the benchmarks are skewed by the time it takes to find the dominating set/spanning tree, but after the reimplementation this time should become negligible, and it will be more about how many vertices need to be checked. So maybe hold-off on removing the methods for now.

Afterwards should be good to merge!

gap/weights.gi
```diff
@@ -772,6 +772,211 @@ function(D, start, destination)
...
+    sum, a, b, st, added, NeighboursV, Edges, notadded,
+    max, NextVertex, notAddedNeighbours, non_leaf;
+
+  if DigraphNrVertices(digraph) = 1 then
+    return 0;
+  fi;
+
+  if DigraphNrStronglyConnectedComponents(digraph) > 1 then
+    return 0;
+  fi;
```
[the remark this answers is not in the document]
reiniscirpons: Hi Rheya, nevermind, I misread the notes, my previous comment was completely wrong, namely it says:

> Similarly, the edge–connectivity lambda(G) of a digraph G is
> the least cardinality |S| of an arc set S such that G – S is no longer strongly connected or is trivial. 

I previously thought that Chapter 2 was exclusively relating to undirected graphs, but after double checking it looks like I had the wrong idea about it. Ill edit my comments to remove these suggestions.

My only hesitation is that Algorithms 4.-7. seem to be written exclusively for undirected graphs, and I'm not sure if the proofs of correctness of these in the literature apply for both directed and undirected graphs.

Looking at the following publication by Esfahanian and Hakirn:

> Esfahanian, A.H. and Louis Hakimi, S. (1984), On computing the connectivities of graphs and digraphs. Networks, 14: 355-366. https://doi.org/10.1002/net.3230140211

they separate the computation of edge-connectivity for graphs and digraphs into two sections and seem to say about the spanning tree-based approach in particular that

> It is important to note that Lemma 7 does not necessarily hold for vertices in R .
> This is why the methods described so far may not work for digraphs.
> But there are other spanning trees!

And then they describe a different method for digraphs. So it might be that the spanning tree algorithm does not work as written for digraphs. It might be simplest to restrict the method to symmetric graphs for now. If it turns out that it also words for non-symmetric ones, we can always expand it in the future.

I don't think you need to implement arc connectivity, it seems to be quite different from the method for edge-connectivity.

mtorpey:
@RheyaM, I wonder if you could merge or rebase to resolve the merge conflicts with the current main branch. Then could you indicate whether you think you've resolved @reiniscirpons's comments? We can then review this again.

reiniscirpons:
Thanks again for the PR @RheyaM ! I took some time to finish off some minor polish so we can merge into main.

@mtorpey @james-d-mitchell Could either of you give this a review when you have the time please?

doc/attr.xml
```diff
@@ -2702,6 +2702,47 @@ gap> Length(M);
...
+      out-dominating set</E> of <A>digraph</A> with respect to the ordering
+      <M>1 &lt; 2 &lt; ... &lt; n</M> where <M>n</M> is the number of vertices in
+      <A>digraph</A>. An <E>out-dominating set</E> is a subset <M>S</M> of
+      vertices of <A>digraph</A> such that every vertex of <A>digraph</A> is
+      either in <M>S</M> or an out-neighbour of a vertex in <M>S</M>. Such a
+      set is <E>greedy</E>, if it is obtained by starting with an empty set
+      <M>S</M> and repeatedly adding to <M>S</M> the least vertex that is not
+      in <M>S</M> and not and out-neighbour of <M>S</M>.
+
+      See also <Ref Oper="IsDigraphOutDominatingSet"/> for a further
```
james-d-mitchell:
```suggestion
      See also <Ref Oper="IsDigraphOutDominatingSet"/> for further
```

doc/attr.xml
```diff
@@ -2702,6 +2702,47 @@ gap> Length(M);
...
+      information on dominating sets, and <Ref
+      Oper="DigraphGreedyOutDominatingSet"/> for a version of this function
+      that finds a greedy out-dominating set with respect to an arbitrary
+      ordering on vertices.
+
+      Similarly, <C>DigraphGreedyInDominatingSet</C> returns the unique <E>greedy
+      in-dominating set</E> of <A>digraph</A> with respect to the ordering
+      <M>1 &lt; 2 &lt; ... &lt; n</M> where <M>n</M> is the number of vertices in
+      <A>digraph</A>. This is equivalently the greedy out-dominating set of
+      the <Ref Attr="DigraphDual"/> of <A>digraph</A>.
```
james-d-mitchell:
```suggestion
      the <Ref Attr="DigraphReverse"/> of <A>digraph</A>.
```
The dual is the graph where every non-edge becomes an edge and vice versa, do you mean the reverse?

doc/oper.xml
```diff
@@ -1954,6 +1954,92 @@ rec( idom := [ 2, fail, 2, 2, 2 ], preorder := [ 2, 1, 3, 4, 5 ] )
...
+      that the list <A>order</A> induced on the vertices in
+      <A>digraph</A>. An <E>out-dominating set</E> is a subset <M>S</M> of
+      vertices of <A>digraph</A> such that every vertex of <A>digraph</A> is
+      either in <M>S</M> or an out-neighbour of a vertex in <M>S</M>. Such a
+      set is <E>greedy</E>, if it is obtained by starting with an empty set
+      <M>S</M> and repeatedly adding to <M>S</M> the least (with respect to the
+      ordering <A>order</A>) vertex that is not in <M>S</M> and not and
+      out-neighbour of <M>S</M>.
+
+      See also <Ref Oper="IsDigraphOutDominatingSet"/> for a further
```
james-d-mitchell:
```suggestion
      See also <Ref Oper="IsDigraphOutDominatingSet"/> for further
```

doc/oper.xml
```diff
@@ -1954,6 +1954,92 @@ rec( idom := [ 2, fail, 2, 2, 2 ], preorder := [ 2, 1, 3, 4, 5 ] )
...
+      information on dominating sets, and <Ref
+      Attr="DigraphGreedyOutDominatingSet"/> for the attribute version of this
+      function that finds a greedy out-dominating set with respect to the
+      ordering <M>1 &lt; 2 &lt; ... &lt; n </M> on the vertices of <A>digraph</A>.
+
+      Similarly, <C>DigraphGreedyInDominatingSet</C> returns the unique <E>greedy
+      in-dominating set</E> of <A>digraph</A> with respect to the ordering
+      that the list <A>order</A> induces on the vertices in
+      <A>digraph</A>. This is equivalently the greedy out-dominating set of
+      the <Ref Attr="DigraphDual"/> of <A>digraph</A>.
```
james-d-mitchell:
```suggestion
      the <Ref Attr="DigraphReverse"/> of <A>digraph</A>.
```

Same comment as above.

doc/weights.xml
```diff
@@ -330,6 +330,7 @@ gap> DigraphMinimumCutSet(g, 1, 3);
 </ManSection>
 <#/GAPDoc>
 
+
```
james-d-mitchell:
```suggestion
```

No change required in this file.

gap/attr.gi
```diff
@@ -3504,6 +3504,149 @@ function(D)
...
+# This function computes the greedy dominating set for the subdigraph
+# of a digraph induced by a set of vertices. The neighbour_fun function
+# determines if in or out-edges are used. Pass OutNeighboursOfVertex
+# for out-edges and InNeighboursOfVertex for in-edges.
+#
+# In other words, we find a subset S of the vertices in the parameter
+# vertices such that every vertex in vertices is in S or adjacent
+# to a vertex in S.
+#
+# This is done in a greedy manner by including every vertex in
```
james-d-mitchell: This is probably not important, but can you possibly reflow this comment, the line breaks look weird to me.

gap/attr.gi
```diff
@@ -3504,6 +3504,149 @@ function(D)
...
+  if not IsSymmetricDigraph(digraph) then
+    ErrorNoReturn("the argument <digraph> must be a symmetric digraph,");
+  fi;
+
+  if DigraphNrVertices(digraph) = 1 or
+      DigraphNrConnectedComponents(digraph) > 1 then
+    return 0;
+  fi;
+
+  EdgeD := UnitEdgeWeightedDigraph(DigraphImmutableCopyIfMutable(digraph));
```
james-d-mitchell: Move this line down to just before `if Length(D) > 1 then` no point in doing this here, again a minor point.

gap/attr.gi
```diff
@@ -3504,6 +3504,149 @@ function(D)
...
+      sum := Minimum(Sum(a), Sum(b));
+      if (sum < min or min = -1) then
+        min := sum;
+      fi;
+
+    od;
+  fi;
+
+  return Minimum(min,
+    Minimum(Minimum(OutDegrees(EdgeD)),
```
james-d-mitchell: `Minimum` is associative, no? So this could just be `Minimum(min, Minimum(OutDegrees(EdgeD)), Minimum(InDegrees(EdgeD))`, right?

gap/oper.gi
```diff
@@ -2522,6 +2522,66 @@ function(D, root)
...
+  [IsDigraph, IsList],
+  {digraph, vertices} ->
+    DIGRAPHS_IsDominatingSet(digraph, vertices, InNeighboursOfVertex));
+
+InstallMethod(DigraphGreedyOutDominatingSet,
+  "for a digraph and a list of vertices",
+  [IsDigraph, IsList],
+  function(digraph, vertex_order)
+    if not Length(vertex_order) = DigraphNrVertices(digraph) or
+       not Set(vertex_order) = DigraphVertices(digraph) then
```
james-d-mitchell:
```suggestion
    if Length(vertex_order) <> DigraphNrVertices(digraph) or
       Set(vertex_order) <> DigraphVertices(digraph) then
```

gap/oper.gi
```diff
@@ -2522,6 +2522,66 @@ function(D, root)
...
+    return Set(DIGRAPHS_GreedyDominatingSet(
+      digraph, vertex_order, OutNeighboursOfVertex));
+  end);
+
+InstallMethod(DigraphGreedyInDominatingSet,
+  "for a digraph and a list of vertices",
+  [IsDigraph, IsList],
+  function(digraph, vertex_order)
+    if not Length(vertex_order) = DigraphNrVertices(digraph) or
+       not Set(vertex_order) = DigraphVertices(digraph) then
```
james-d-mitchell: Same comment as above

james-d-mitchell (changes requested):
Some superficial changes, then I'm happy to merge this

PR #889, opened 2025-12-02
Hamiltonian path finding function and associated tests

ap4108735:
I have added a function to find the more general notion of a Hamiltonian path, to be used alongside the existing Hamiltonian cycle function. I have also added an associated testing function. At the moment, for large cases the function is not very efficient so I will continue to look into some possible fixes for that.

### DIFF
--- a/tst/extreme/attr.tst
+++ b/tst/extreme/attr.tst
@@ -43,20 +43,20 @@ gap> circs := DigraphAllSimpleCircuits(gr);;
 gap> Length(circs);
 1291792
 
-#  HamiltonianPath and IsHamiltonianDigraph
+#  HamiltonianCycle and IsHamiltonianDigraph
 gap> g := CompleteDigraph(20);
 <immutable complete digraph with 20 vertices>
-gap> HamiltonianPath(g);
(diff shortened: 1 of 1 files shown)


gap/oper.gi
```diff
@@ -1596,6 +1586,40 @@ function(D, edges)
   return true;
 end);
 
+InstallMethod(TestHamiltonianPath, "for a digraph", [IsDigraph],
```
james-d-mitchell: This would be more consistent with elsewhere in the Digraphs package if it was:

```suggestion
InstallMethod(IsHamiltonianPath, "for a digraph", [IsDigraph, IsList],
```

i.e. returns `true` if the 2nd argument is a hamiltonian path in the 1st argument.

configure~
james-d-mitchell: This was committed by accident and should removed.

digraph.dot
james-d-mitchell: Remove, unless required for something, but I can't see how it could be.

my_digraph.dot
james-d-mitchell: Remove

PR #899, opened 2026-02-18
Implementation of DigraphInsertEdge and DigraphReduceEdge

linuskuehnle:
Implemented DigraphInsertEdge and DigraphReduceEdge.
These functions already exist in the Simplicial Surfaces package, which I am currently working on with Meike Weiß at RWTH Aachen University.
For reference, the corresponding functions there are implemented as EdgeInsertion and EdgeReduction.

Both DigraphInsertEdge and DigraphReduceEdge check whether the given digraph D is symmetric, and modifications to D preserve this symmetry.
Should D therefore retain the symmetric tag by explicitly calling SetIsSymmetricDigraph after modification?
If so, is an additional call to SetDigraphSymmetricClosureAttr required?

### DIFF
--- a/doc/oper.xml
+++ b/doc/oper.xml
@@ -2037,7 +2037,7 @@ true
   <Description>
     If <A>edge</A> is a pair of vertices of <A>digraph</A>, or <A>src</A> and
     <A>ran</A> are vertices of <A>digraph</A>, where <A>ran</A> &lt;&gt; <A>src</A>, 
-    then then this operation merges the two vertices of the edge given into one. 
+    then this operation merges the two vertices of the edge given into one. 
     Edges incident to <A>src</A> and <A>ran</A> will now be incident to <C>v</C>, 
     the new vertex, with their direction preserved. <P/>
     
@@ -2083,13 +2083,13 @@ gap> DigraphVertexLabels(D);
   <Returns>A digraph.</Returns>
   <Description>
     If <A>edge1</A> and <A>edge2</A> are distinct pairs of vertices of <A>digraph</A>,
-    then then this operation inserts a new edge between <A>edge1</A> and
+    then this operation inserts a new edge between <A>edge1</A> and
     <A>edge2</A>.
     <A>edge1</A> and <A>edge2</A> are split up into two new edges each, which are
     all adjacent to the inserted edge. Hence, the vertices of the inserted edge both
-    have a degree of 3.
+    have a degree of 3.<P/>
 
-    The opposite operation is <Ref Oper="DigraphReduceEdge"/>.
+    The opposite operation is <Ref Oper="DigraphReduceEdge"/>.<P/>
 
     A new digraph constructed from <A>digraph</A> is returned,
     unless <A>digraph</A> belongs to <Ref Filt="IsMutableDigraph"/>;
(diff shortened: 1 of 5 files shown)


doc/oper.xml
```diff
@@ -2076,6 +2076,84 @@ gap> DigraphVertexLabels(D);
...
+<ManSection>
+    <Oper Name="DigraphInsertEdge" Arg="digraph, edge1, edge2"
+        Label="for a symmeric digraph and two lists of positive integers"/>
+  <Returns>A digraph.</Returns>
+  <Description>
+    If <A>edge1</A> and <A>edge2</A> are distinct pairs of vertices of <A>digraph</A>,
+    then then this operation inserts a new edge between <A>edge1</A> and
+    <A>edge2</A>.
+    <A>edge1</A> and <A>edge2</A> are split up into two new edges each, which are
+    all adjacent to the inserted edge. Hence, the vertices of the inserted edge both
```
james-d-mitchell: What does it mean for an edge to be adjacent to an edge?

james-d-mitchell:
Thanks for the contribution @linuskuehnle ! I'll try to review your PR and answer your questions early next week.

gap/oper.gi
```diff
@@ -603,6 +603,183 @@ function(D, edge)
   return DigraphContractEdge(D, edge[1], edge[2]);
 end);
 
+DIGRAPHS_CheckInsertEdgeDigraph := function(D, edge1, edge2)
```
james-d-mitchell: This function could be a number of lines shorter if the conditions were combined with `elif` rather than being separate `if` conditions.

james-d-mitchell (changes requested):
Generally this looks pretty good, thanks for the contribution @linuskuehnle ! If you could please address the comments, then I'll be happy to merge this.

doc/oper.xml
```diff
@@ -2076,6 +2076,84 @@ gap> DigraphVertexLabels(D);
...
+    have a degree of 3.
+
+    The opposite operation is <Ref Oper="DigraphReduceEdge"/>.
+
+    A new digraph constructed from <A>digraph</A> is returned,
+    unless <A>digraph</A> belongs to <Ref Filt="IsMutableDigraph"/>;
+    in this case changes are made directly to <A>digraph</A>, which is then returned. 
+    The <A>digraph</A> must not belong to <Ref Filt="IsMultiDigraph"/>. <P/>
```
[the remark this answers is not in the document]
linuskuehnle: I copied this section from the documentation of DigraphContractEdge. I can mention that an error is thrown in that case, should i then adjust it for DigraphContractEdge as well?

gap/oper.gd
```diff
@@ -38,6 +38,10 @@ DeclareOperation("DigraphClosure", [IsDigraph, IsPosInt]);
 DeclareOperation("DigraphContractEdge", [IsDigraph, IsPosInt, IsPosInt]);
 DeclareOperation("DigraphContractEdge", [IsDigraph, IsDenseList]);
 
+DeclareOperation("DigraphInsertEdge", [IsDigraph, IsDenseList, IsDenseList]);
```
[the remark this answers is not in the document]
linuskuehnle: I chose to implement it with a rather generic operation name since you would probably change it anyways. However i also don't really have an idea for a more specific name, hence yes i am open to suggestions

doc/oper.xml
linuskuehnle: By adjacent edges I mean edges that share a common vertex. The operation subdivides edge1 and edge2 by placing a new vertex on each, and then connects those two new vertices with a newly inserted edge. So the four 'split' edge halves share endpoints with the newly inserted edge. Hope i explained it properly now.
Should i rephrase the documentation here?

PR #905, opened 2026-03-11
Implementing Colour Refinement

RheyaM:
Implementation of DigraphColourRefinement. Takes a digraph and returns a colouring (A list where value at i<sup>th</sup> position is the colour of the i<sup>th</sup> node), such that all nodes with the same colour have the same configuration of coloured neighbours. That is, if two nodes have the same colour, they will have the same number of neighbours of colour x, and colour y, etc. DigraphColourRefinement considers out and in neighbours separately.

If two digraphs produce two different colouring under Colour Refinement, we can know that they are not isomorphic.

Now includes optimisation step


gap/oper.gi
```diff
@@ -2732,3 +2732,154 @@ function(D, n)
...
+
+  c_min := 1;
+  c_max := 1;
+
+  # Queue of colours
+  Q := [1];
+
+  # Initial colouring
+  # vertices -> colour
+  C := List([1 .. DigraphNrVertices(D)], x -> 1);
```
mtorpey: Better would be `ListWithIdenticalEntries(DigraphNrVertices(D), 1)`

gap/oper.gi
```diff
@@ -2732,3 +2732,154 @@ function(D, n)
...
+  # All vertices initialised to 1
+  # colour -> vertices labelled as such
+  P := [];
+  P[1] := [1 .. DigraphNrVertices(D)];
+
+  while not IsEmpty(Q) do
+
+    # Pop colour off Q
+    q := Q[1];
+    Remove(Q, 1);
```
mtorpey: `Remove` acts as a pop function, so we can actually combine this to
`q := Remove(Q, 1);`

gap/oper.gi
```diff
@@ -2732,3 +2732,154 @@ function(D, n)
...
+
+    # For each v (vertices) in D:
+    # Get the neighbours of v that are in the colour class q
+    Out := rec();
+    In := rec();
```
mtorpey: Records that use `[1..n]` as the fields are likely to be slower than a simple list. Why use `Out.(3)` when you could use `Out[3]`?

I suggest setting both `Out` and `In` to `EmptyPlist(DigraphNrVertices(D))` and then writing to them with `Out[v]` and `In[v]`

gap/oper.gi
```diff
@@ -2732,3 +2732,154 @@ function(D, n)
...
+    for v in DigraphVertices(D) do
+      Out.(v) := Intersection(OutNeighbours(D)[v], P[q - c_min + 1]);
+      In.(v) := Intersection(InNeighbours(D)[v], P[q - c_min + 1]);
```
mtorpey: `OutNeighbours(D)` and `InNeighbours(D)` are called in a double-nested loop here, which is potentially a lot of calls for something that gives the same answer every time.

I suggest reading these into local variable at the beginning of the function, just to avoid these calls.

mtorpey: This might also be worth doing with `DigraphVertices` and `DigraphNrVertices`.

gap/oper.gi
```diff
@@ -2732,3 +2732,154 @@ function(D, n)
...
+    od;
+
+    # CD:
+    # [colour of v, number of q coloured out-neighbours of v,
+    # number of q coloured in-neighbours of v, v]
+    CD := [];
+    for v in DigraphVertices(D) do
+      Add(CD, [C[v], Length(Out.(v)), Length(In.(v)), v]);
+    od;
```
mtorpey: Since `CD` will have a known length, we can give this as a hint to the interpreter by using `CD := EmptyPlist(DigraphNrVertices(D))` instead of `CD := []`.

Alternatively, do the map-style shorthand thing with `CD := List(DigraphVertices(D), v -> ...)`

gap/oper.gi
```diff
@@ -2732,3 +2732,154 @@ function(D, n)
...
+
+    Sort(CD);
+
+    Sets := [];
+    currentPair := [];
+    colour_cell := [];
```
mtorpey: Are you using camelCase or snake_case? Be consistent.

gap/oper.gi
```diff
@@ -2732,3 +2732,154 @@ function(D, n)
...
+    # [colour of v, number of q coloured out-neighbours of v,
+    # number of q coloured in-neighbours of v, v]
+    CD := [];
+    for v in DigraphVertices(D) do
+      Add(CD, [C[v], Length(Out.(v)), Length(In.(v)), v]);
+    od;
+
+    Sort(CD);
+
+    Sets := [];
```
mtorpey: Use lower-case for local variables, unless they're symbols like `S` or `D`.

gap/oper.gi
```diff
@@ -2732,3 +2732,154 @@ function(D, n)
...
+    recolour := false;
+
+    j := 0;
+
+    for pair in CD do
+      current := [pair[1], pair[2], pair[3]];
+
+      # If different colour reached:
+      if currentPair <> [] and current[1] <> currentPair[1] then
+        Add(colour_cell, cell);
```
mtorpey: Should this be `AddSet`? I notice `colour_cell` is being added to something called `Sets`, so perhaps you intend it to be a set (i.e. a duplicate-free, sorted list, which can be searched more easily). If not, then maybe `Sets` is the wrong name.

mtorpey: (Several more instances of `Add` below)

gap/oper.gi
```diff
@@ -2732,3 +2732,154 @@ function(D, n)
...
+        od;
+      od;
+    fi;
+
+  od;
+
+  # Normalising C to start at 1
+  for i in [1 .. Length(C)] do
+    C[i] := C[i] - (c_min - 1);
+  od;
```
mtorpey: You can subtract a scalar from a vector in GAP! So all this is just `C := C - (c_min - 1)`

tst/standard/oper.tst
```diff
@@ -3324,6 +3324,32 @@ gap> DigraphEdges(D);
...
+gap> D := Digraph([[2], [3], [1]]);;
+gap> DigraphColourRefinement(D);
+[ 1, 1, 1 ]
+gap> D := Digraph([[2, 4], [5], [2, 4], [5], [1, 3]]);;
+gap> DigraphColourRefinement(D);
+[ 2, 1, 2, 1, 3 ]
+gap> D := Digraph([[4], [1, 3], [4], [5], [1, 3]]);;
+gap> DigraphColourRefinement(D);
+[ 2, 3, 2, 1, 4 ]
+
```
mtorpey, line 3358: Perhaps a couple more tests, for things like an empty digraph, digraph with loops, and any other edge cases you can think of.

mtorpey (dismissed):
A few suggestions for changes here! Great work, and this stuff shouldn't take too much longer.

gap/oper.gi
```diff
@@ -2732,3 +2732,154 @@ function(D, n)
...
+
+        colour_cell := Sets[i];
+
+        # Determine the largest cell for that colour
+        largest := 1;
+        for j in [1 .. Length(colour_cell)] do
+          if Length(colour_cell[j]) > Length(colour_cell[largest]) then
+            largest := j;
+          fi;
+        od;
```
mtorpey: I think this might be replacable with `largest := PositionMaximum(colour_cell, Length)`. Could you try this?

mtorpey:
Hi @RheyaM! Well done on the quick fixes.

Once you've tried that PositionMaximum fix above, it's just about resolving that "canonical" thing. If you can have a quick think about that and then leave a comment here summarising to what extent it's canonical, and what "canonical" means, I think we can merge this.

Joseph-Edwards (commented):
I've had quite a thorough look at this now, and it seems to me like the colouring that get's returned is canonical; as far as I can tell, we don't use any information of the vertex labels when deciding which cells/colours to refine with respect to. If we could mention this in the documentation (and also do what Michael suggested in his comment) then I would be happy for this to be merged.

Nice work @RheyaM!

doc/oper.xml
```diff
@@ -2637,6 +2637,10 @@ true]]>
     <C>DigraphColourRefinement</C> returns the colouring as a list where the value at the ith position is the 
     colour of node i. The time complexity of this algorithm is <M>O(n^2 log n)</M>.
     <P/>
+    Because the labels of the vertices are not used in any capacity during the refinement process (I.e. to determine which cells 
+    to refine), the colouring produced is canonical. This means that for two isomorphic digraphs, they would produce the same colouring.
```
mtorpey: Sorry to keep fussing, but one more thing that should be made clear here.

The **colouring** is canonical, but my understanding is that the **output of the function** is not canonical (because it's a list that uses the ordering of vertices). This should be made clear, because as written it sounds like isomorphic digraphs have the same output.

Have I got this right

doc/oper.xml
```diff
@@ -2637,6 +2637,10 @@ true]]>
     <C>DigraphColourRefinement</C> returns the colouring as a list where the value at the ith position is the 
     colour of node i. The time complexity of this algorithm is <M>O(n^2 log n)</M>.
     <P/>
+    Because the labels of the vertices are not used in any capacity during the refinement process (I.e. to determine which cells
```
mtorpey: Lower case `i.e.`

mtorpey (changes requested):
Sorry, nearly done

doc/oper.xml
RheyaM: I added an explanation for this, but I'm not sure if it comes across clearly so I included two isomorphic digraphs in the examples and their 'identical' colourings (I just swapped vertices 1 and 8, and the output swaps the colours of those vertices as well). Is that okay/does it make sense?

mtorpey (approved):
This looks great to me. Well done on all the hard work!

PR #909, opened 2026-03-20
Add Nix flake for deterministic Nix builds (attempt 2)

ThatOtherAndrew:
[Nix](https://nixos.org/) is a functional programming language used for deterministic builds, by pinning environment "inputs" (e.g. `gcc`) to a specific version. This allows for a universal and deterministic `nix build` command which replaces the need for autogen, configure, and make.

This is a recreation of a previous PR (#906) which was messed up due to accidentally committing multiple features to the main branch. I have learned from my mistake and one sleepless night later my changes are properly branched now. oops!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,7 @@ main.xml
 Makefile
 manual.lab
 missing
+result
 src/pkgconfig.h.in
 tags
 tst/out/
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1772963539,
+        "narHash": "sha256-9jVDGZnvCckTGdYT53d/EfznygLskyLQXYwJLKMPsZs=",
+        "owner": "nixos",
+        "repo": "nixpkgs",
+        "rev": "9dcb002ca1690658be4a04645215baea8b95f31d",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nixos",
+        "ref": "nixos-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "nixpkgs": "nixpkgs"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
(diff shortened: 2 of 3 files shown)


reiniscirpons (commented):
Hi Andromeda, thanks for the PR! I'm not a nix user myself, but think its a good effort to try and package Digraphs for simpler installation.

Some comments:

1. It seems the flake only lists gap itself as a dependency, but Digraphs depends on other gap packages such as `datastructures` and `io` (see the `Dependencies` record in `PackageInfo.g` (https://github.com/digraphs/Digraphs/blob/main/PackageInfo.g#L560) for more info). The `gap` dependency may or may not package them, and may or may not package the right versions of these packages, so its probably a good idea to explicitly set them as dependencies.

This of course introduces some headache, since as far as I know neither the `datastructures` nor `io` are available as nix packages. Of course, in theory we could make flakes available for these packages too, but now these would also need to be maintained etc, which can quickly get out of hand. I think this should be a community rather than single author effort to make it maintainable, but at the moment nix is not used much in the community as far as I can tell.

2. Gap has its own package manager: the `PackageManager` package (documentation: https://gap-packages.github.io/PackageManager/doc/chap0_mj.html , github: https://github.com/gap-packages/PackageManager) by @mtorpey . It solves quite a lot of headaches related to building and installing gap packages and their dependencies, and if I understand correctly it should be able to build Digraphs too (including a development version). So maybe it would be possible to simplify the script by having a separate flake for `PackageManager`, and then all this flake would do is call the PackageManager `InstallPackage` command from gap?

3. Before merging this PR we would probably want to add some CI job for testing the nix builds. Otherwise it is a very high probability that this file will just go out-of-sync or break as we update the package. At the same time, I'm not sure if any of the current maintainers of digraphs would be able to fix any nix-related issues, since we are not nix users ourselves.

Overall I think its a good effort and probably quite useful for other nix users wanting to install digraphs, but I'm not sure we will merge this PR in the near future mainly due to a lack of nix support in the gap ecosystem and lack of nix know-how among the digraphs maintainers. So don't feel like you need to spend much more time fixing any of the issues above.

If you are interested in gap packaging in general, it might be a good idea to talk to Michael at one of the meetings about the PackageManager package. Thanks again!

Issue #714, opened 2024-11-20
Implement functions for detecting whether a digraph is inbetween the cover relation and relation for a class of relations


reiniscirpons:
Computing all the edges of a partial order or lattice via `DigraphReflexiveTransitiveClosure` can be very time and memory intensive, especially for larger graphs. At the same time, many algorithms for lattices can be run quite efficiently even if some transitive edges are missing.

We should implement the functions
* `IsDigraphInbetweenCoverAndPreorder` - to check if a given digraph contains the cover relation of some preorder digraph, and is in turn contained in a preorder digraph.
* `IsDigraphInbetweenCoverAndPartialOrder` - the equivalent functionality for partial order
* `IsDigraphInbetweenCoverAndLattice` - for lattices.

One naive way to implement this would be to compute the reflexive transitive closure and then apply the appropriate filter, i.e. `IsLatticeDigraph(DigraphReflexiveTransitiveClosure(D));` would compute `IsDigraphInbetweenCoverAndLattice`. However as mentioned above, doing so is memory intensive on large graphs, see also #636 . So the idea would be to implement `IsDigraphInbetweenCoverAndLattice` ideally without adding any extra edges to the given digraph.

See also #664 and #632 for related, but not quite same problems.

reiniscirpons:
@zljlzljlz this is a writeup of what we spoke about earlier. I think the name I initially suggested for the function wasn't all that good and we should instead call the functions `IsDigraphInbetweenCoverAndX` for each of the classes `X`.

As you mentioned, `IsDigraphInbetweenCoverAndPreorder` should just always return `true` since this will always be the case. For `IsDigraphInbetweenCoverAndPartialOrder(D)` I spoke to @james-d-mitchell  briefly and we thought that maybe this would just involve checking if `D` is acyclic. For lattices we weren't really sure how to proceed but we can have a brainstorm next week!

mtorpey:
This is what @zljlzljlz is working on, right?

PR #910, opened 2026-03-20
Implement IsJoinIrreducible and IsMeetIrreducible

### LINKED ISSUE #400: Implement IsJoin/MeetIrreducible for vertices in lattice digraphs
...and for other kinds of digraphs where the notion makes sense, such as maybe join/meet semilattice digraphs, and possibly even partial order digraphs in general.

And correspondingly, there could be an attribute `JoinIrreducibleVertices`, etc.

This is just an idea!

There's a very slight possibility that this could ultimately be used by the Semigroups package, where there exist functions such as `IsJoinIrreducible` and `JoinIrreducibleDClasses`.

The notion of join/meet primality might also be good to have https://en.wikipedia.org/wiki/Lattice_(order)#Important_lattice-theoretic_notions

m31-galaxy:
Adds two new operations to section 10 (Operations for vertices) of `oper.gi`/`oper.gd`:

- `IsJoinIrreducible(D, v)` — returns `true` if vertex `v` is not the join of any two distinct smaller nodes in `D`
- `IsMeetIrreducible(D, v)` — returns `true` if vertex `v` is not the meet of any two distinct larger nodes in `D`

Both were originally implemented by building the full O(N²) join/meet table, but then optimised to use `DigraphReflexiveTransitiveReduction`, a function producing a Hasse diagram which allows for a computational shortcut (see SageMath docs below):

https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/posets/hasse_diagram.html#sage.combinat.posets.hasse_diagram.HasseDiagram.find_nontrivial_congruence

Tests are merged into `tst/standard/oper.tst`, covering N5, a chain, B2, M3, and error handling (two test case sets written by myself and the rest with LLM assistance because I do not think my mathematical knowledge of the underlying concepts is strong enough to come up with reasonably comprehensive/diverse test cases).

---

Closes #400


james-d-mitchell (dismissed):
Looks good to me! If you also resolve #714, then this PR (or the code if merged first) should be updated to check for that rather than `IsPartialOrderDigraph`. Otherwise, this looks good to me!

Issue #47, opened 2017-03-02
Improve the documentation/behaviour of the digraph encoding formats


wilfwilson:
I don't think it is sufficiently easy to find out which of the four main digraph encoding formats apply to which kinds of digraphs. I'm talking about `(di)sparse6` and `(di)graph6`. For instance, I naively did the following, expecting it to work:
```
gap> gr := Digraph([[2, 2], []]);
<multidigraph with 2 vertices, 2 edges>
gap> Digraph6String(gr);
"+AG"
gap> DigraphFromDigraph6String(last) = gr;
false
```
I got a string from `Digraph6String(gr)`, so why didn't recreating the digraph from the string work? I did `?Digraph6String` and got the following documentation, which covers all four of the formats:
```
These  four functions return a highly compressed string fully describing the
digraph digraph.

Graph6  and  Digraph6  are  formats  best  used  on  small, dense graphs, if
applicable. For larger, sparse graphs use Sparse6 and Disparse6 (this latter
also preserves multiple edges).

See WriteDigraphs (9.2-5).
```
*[Technical note: shouldn't these things be attributes, and described as such, since they don't change?]*
*[Also, haven't we decided at some point that the names `Graph6` and `Digraph6` and `Sparse6` and `Disparse6` are written without capital letters, when they appear as words in the middle of a sentence?]*

From reading this, I would interpret the phrase "fully describing the digraph" to mean that each format can completely represent any type of digraph; I got a string representing my digraph `gr` above, so I would expect it to "fully describe" `gr`. However the phrase "(this latter also preserves multiple edges)", which comes later, seems confusing in this context, as it implies that some of the formats don't support all digraphs.

There is the final line, which is a direction to look at `WriteDigraphs`, but it is not clear why this is relevant or what one would expect to find there. I certainly wouldn't expect that that would be the place to find a more detailed description of the formats.

Similarly, doing `?DigraphFromDigraph6String` gives me the following:
```
If  str  is  a  string  encoding  a  graph  in  Graph6, Digraph6, Sparse6 or
DiSparse6  format, then the corresponging function returns a digraph. In the
case  of  either  Graph6  or  Sparse6, formats which do not support directed
edges,  this  will  be a digraph such that for every edge, the edge going in
the opposite direction is also present.
```
[...]

PR #912, opened 2026-03-25
Digraph decode/encode

anujthak:
Addressing the latter part of #47 , makes it so that it is now impossible to encode a digraph into an improper format(ex: a graph with multiple edges into Digraph6) and also recommends alternatives for a proper encoding in the error message.

### DIFF
--- a/gap/io.gi
+++ b/gap/io.gi
@@ -2283,15 +2283,6 @@ function(D)
   # matrix, and appends a '&' to the start.  The old '+' format can be read by
   # DigraphFromDigraph6String, but can no longer be written by this function.
 
-  if IsMultiDigraph(D) then
-    if IsSymmetricDigraph(D) then
-      ErrorNoReturn("the argument <D> must not have multiple edges ",
-                  "consider encoding in Sparse6 or Disparse6, ");
-    fi;
-    ErrorNoReturn("the argument <D> must not have multiple edges ",
-                  "consider encoding in Disparse6, ");
-  fi;
-
   list := [];
   adj := OutNeighbours(D);
   n := Length(DigraphVertices(D));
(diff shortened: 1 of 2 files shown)


gap/io.gi
```diff
@@ -2222,10 +2222,15 @@ InstallMethod(Graph6String, "for a digraph by out-neighbours",
...
-                  "with no loops or multiple edges,");
+  if IsMultiDigraph(D) then
+    ErrorNoReturn("the argument <D> must not have multiple edges; ",
+      "consider encoding in Disparse6 or Digraph6");
+  elif not IsSymmetricDigraph(D) then
+    ErrorNoReturn("the argument <D> must be a symmetric digraph; ",
+      "consider encoding in Sparse6 or Disparse6");
+  elif DigraphHasLoops(D) then
+    ErrorNoReturn("the argument <D> must not have loops; ",
+      "consider encoding in Sparse6 or Disparse6");
```
mtorpey: I think all three of these "consider" lines are wrong. Revise this with Wilf's summary from his issue:

> Graph6: can encode any symmetric digraph without loops or multiple edges
> Sparse6: can encode any symmetric digraph
> Digraph6: can encode any digraph without multiple edges
> Disparse6: can encode any digraph

tst/standard/io.tst
```diff
@@ -283,8 +283,8 @@ true
 gap> gr[3] := Digraph([[1, 2], [1, 2]]);
 <immutable digraph with 2 vertices, 4 edges>
 gap> WriteDigraphs(filename, Digraph([[2], []]), Graph6String);
-Error, the argument <D> must be a symmetric digraph with no loops or multiple \
-edges,
+Error, the argument <D> must be a symmetric digraph; consider encoding in Spar\
+se6 or Disparse6
 gap> OnBreak := oldOnBreak;;
```
mtorpey, line 288: The fact these break across lines is probably an argument for putting `\n` after the semicolon. Could you do this in all cases?

mtorpey (changes requested):
This has the right basic form, but needs to be corrected a bit. See my comments below.

mtorpey (approved):
This is now looking good to me!

PR #919, opened 2026-04-22
attr.gi code coverage

mtorpey:
These PR fixes the four untested sections of code from attr.gi, some by adding tests, and some by amending the code.

The four commits are logically separate, so it might be better not to squash.

Code coverage of attr.gi should now be 100%.


james-d-mitchell (approved):
Nice, thanks @mtorpey

PR #920, opened 2026-04-22
Implement IsBetweenCoverAndLattice

m31-galaxy:
Prompted by conversation in #714, I believe this is a functional implementation of the `IsBetweenCoverAndLattice` function. My implementation is based on the three-step outline given by @Joseph-Edwards and also implements `DIGRAPHS_MeetJoinTableBetweenCover` (as a slightly modified/generalised version of `DIGRAPHS_MeetJoinTable`) in order to make this possible. I've thrown in some tests as well (due to my limited understanding of the maths I'm unsure how comprehensive they are though!) and they do appear to pass when building and running locally. (hopefully CI/CD can validate if they still work!)

### DIFF
--- a/gap/prop.gi
+++ b/gap/prop.gi
@@ -140,23 +140,36 @@ D -> DIGRAPHS_IsMeetSemilatticeAndMeetTable(D)[1]);
 InstallMethod(IsBetweenCoverAndLattice, "for a digraph",
 [IsDigraph],
 function(D)
-  local order, hasse;
+  local copy, order, hasse, neighbours, table;
 
   # 1. Topologically sort the nodes in D.
+  copy := DigraphRemoveLoops(DigraphMutableCopyIfMutable(D)); # protect from nasty mutable side effects i think
   order := DigraphTopologicalSort(copy);
   if order = fail then
     return false;
   fi;
 
   # 2. Iterate through pairs of nodes of D in topological order and construct a table of their meets.
-  hasse := DigraphTransitiveReduction(DigraphMutableCopyIfMutable(copy));
-  in := InNeighbours(hasse);
-  # something along the lines of DigraphMeetTable()
-  out := OutNeighbours(hasse);
-  # something along the lines of DigraphJoinTable()
+  hasse := DigraphTransitiveReduction(copy);
+  neighbours := InNeighbours(hasse);
+  table := DIGRAPHS_MeetJoinTableBetweenCover(
+    DigraphNrVertices(copy),
+    Reversed(order),
+    neighbours,
+    false
+  );
+  if table = fail then
+    return false;
+  fi;
 
-  # TODO: step 3
-  return true;
+  neighbours := OutNeighbours(hasse);
+  table := DIGRAPHS_MeetJoinTableBetweenCover(
+    DigraphNrVertices(copy),
+    order,
+    neighbours,
+    true
+  );
+  return table <> fail;
 end);
 
 InstallImmediateMethod(IsStronglyConnectedDigraph,
--- a/gap/prop.gi
+++ b/gap/prop.gi
@@ -137,6 +137,28 @@ InstallMethod(IsMeetSemilatticeDigraph, "for a digraph",
 [IsDigraph],
 D -> DIGRAPHS_IsMeetSemilatticeAndMeetTable(D)[1]);
 
+InstallMethod(IsBetweenCoverAndLattice, "for a digraph",
+[IsDigraph],
+function(D)
+  local order, hasse;
+
+  # 1. Topologically sort the nodes in D.
+  order := DigraphTopologicalSort(copy);
+  if order = fail then
+    return false;
+  fi;
+
+  # 2. Iterate through pairs of nodes of D in topological order and construct a table of their meets.
+  hasse := DigraphTransitiveReduction(DigraphMutableCopyIfMutable(copy));
+  in := InNeighbours(hasse);
+  # something along the lines of DigraphMeetTable()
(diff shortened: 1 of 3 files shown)


m31-galaxy:
gaplint should be satisfied now :>

gap/prop.gi
```diff
@@ -91,6 +91,60 @@ function(D, P, U, join)
...
+          q := z;
+        fi;
+      od;
+      for z in T do
+        # the below conditions are the only part that
+        # differs from MeetJoinTable above
+        if join and tab[q, z] <> z then
+          return fail;
+        elif not join and tab[z, q] <> z then
+          return fail;
```
reiniscirpons, line 137: Given that this is the only difference, would it be feasible to modify `DIGRAPHS_MeetJoinTable` to have both functionalities e.g. by adding an extra  boolean parameter `check_edges`, which makes the `DIGRAPHS_MeetJoinTable` behave the same as before if set to `true` and otherwise make it behave like `DIGRAPHS_MeetJoinTableBetweenCover`?

gap/prop.gi
```diff
@@ -137,6 +191,41 @@ InstallMethod(IsMeetSemilatticeDigraph, "for a digraph",
 [IsDigraph],
 D -> DIGRAPHS_IsMeetSemilatticeAndMeetTable(D)[1]);
 
+InstallMethod(IsBetweenCoverAndLattice, "for a digraph",
+[IsDigraph],
+function(D)
+  local copy, order, hasse, neighbours, table;
+
```
reiniscirpons:
```suggestion
 if IsMultiDigraph(D) then
    ErrorNoReturn("the argument must not be a multidigraph,");
  fi;
```

gap/prop.gi
```diff
@@ -137,6 +191,41 @@ InstallMethod(IsMeetSemilatticeDigraph, "for a digraph",
...
+  # 1. Topologically sort the nodes in D.
+  copy := DigraphRemoveLoops(DigraphMutableCopyIfMutable(D));
+  # ^ protect from nasty mutable side effects i think
```
reiniscirpons:
```suggestion
```

This is correct, but not a very useful comment.

reiniscirpons (changes requested):
Thanks for the commit! I have suggested some changes. One larger change I would like is for you to modify `DIGRAPHS_MeetJoinTable` to have the behavior of `DIGRAPHS_MeetJoinTableBetweenCover` depending on a switch.

One other thing: The function is currently missing documentation. To add this you should modify the `doc/prop.xml` file and add an entry to this file for the `IsBetweenCoverAndLattice` function, see the doc for `IsMeetSemilatticeDigraph` function doc for an example:

https://github.com/digraphs/Digraphs/blob/313228bfcdf5e16f8c45f6678240e1f34e8d4b38/doc/prop.xml#L1295-L1349

You then want to include the function doc in the manual, the `doc/z-chap5.xml` file in the orders section after `IsMeetSemilatticeDigraph` is probably the right place for this:

https://github.com/digraphs/Digraphs/blob/313228bfcdf5e16f8c45f6678240e1f34e8d4b38/doc/z-chap5.xml#L41-L51

Otherwise looks good in the first instance!

gap/prop.gi
```diff
@@ -137,6 +191,41 @@ InstallMethod(IsMeetSemilatticeDigraph, "for a digraph",
...
+    return false;
+  fi;
+
+  neighbours := OutNeighbours(hasse);
+  table := DIGRAPHS_MeetJoinTableBetweenCover(
+    DigraphNrVertices(copy),
+    order,
+    neighbours,
+    true);
+  return table <> fail;
```
reiniscirpons, line 227: Given the similarity of this line to the previous one, would it might make more sense to have two functions `IsBetweenCoverAndMeetSemilattice` and `IsBetweenCoverAndJoinSemilattice` which to this check with the meet and join table respectively and then make `IsBetweenCoverAndLattice` to simply return `true` when both `IsBetweenCoverAndMeetSemilattice` and `IsBetweenCoverAndJoinSemilattice`  are true, like we do for `IsLatticeDigraph`.

gap/prop.gi
```diff
@@ -137,6 +191,41 @@ InstallMethod(IsMeetSemilatticeDigraph, "for a digraph",
...
+  neighbours := InNeighbours(hasse);
+  table := DIGRAPHS_MeetJoinTableBetweenCover(
+    DigraphNrVertices(copy),
+    Reversed(order),
+    neighbours,
+    false);
+  if table = fail then
+    return false;
+  fi;
+
```
reiniscirpons, line 220:
```suggestion
  if IsImmutableDigraph(D) then
    SetDigraphMeetTable(D, tab);
  fi;
```
We can save the result as the meet table here, since we already did all that work to compute it. The same should be done for the join table below.

reiniscirpons (changes requested):
A few more things I noticed. We should discuss next meeting, but overall looks good.

gap/prop.gi
m31-galaxy: On it! Would doing a text search for "DIGRAPHS_MeetJoinTable" be sufficient to ensure that all places affected by the function signature breaking change are covered?

reiniscirpons: Yes I think so! It would also be nice if you could add a comment describing what each of the input parameters `D, P, U, join` and the new one you add is expected to be, and how using them affects the output. The `DIGRAPHS_` prefix functions are for internal use only, so they don't get an official documentation entry, but it might be useful for the next person who touches the code.

Issue #546, opened 2022-05-09
Map graphs


mtorpey:
Do we have anything much to do with [map graphs](https://en.wikipedia.org/wiki/Map_graph)?  We can view a map (e.g. a map of the United States) as a graph where the different regions (e.g. US states) are the vertices, and an edge lies between two vertices whenever there is a border between two states.

Most of these are planar graphs, but we can have things like the [Four Corners](https://en.wikipedia.org/wiki/Four_Corners) in the US, where four states meet at a point, and this makes map graphs a superset of the planar graphs.  I think the mathematical definitions don't allow exclaves (disconnected segments of regions, like Russia's Kaliningrad oblast) which tend to confuse things, but I'm not sure of that.

I don't think we have anything to do with these, but it might be nice in future, perhaps as a student project.  I'd like to see:
- An algorithm to determine `IsMapGraph`;
- Special `ChromaticNumber`/`DigraphColoring` methods (the [maximum must be 4](https://en.wikipedia.org/wiki/Four_color_theorem));
- A visual alternative to `DotDigraph`, to draw it as a map instead of a graph.

and I'm sure there's other great stuff to think about.

mtorpey:
Looks like Anuj is going to take a look at this.

PR #922, opened 2026-04-23
Map Graph Recognition

anujthak:
Adding an algorithm in response to #546 that recognizes whether or not an inputted graph is a map graph.

### DIFF
--- a/doc/planar.xml
+++ b/doc/planar.xml
@@ -464,3 +464,39 @@ true
 </ManSection>
 
 <#/GAPDoc>
+
+<#GAPDoc Label="IsMapGraph">
+<ManSection>
+  <Attr Name="IsMapGraph" Arg="D"/>
+  <Returns><K>true</K> or <K>false</K>.</Returns>
+  <Description>
+    A map graph is a graph whose vertices can be represented as disjoint 
+    regions in the plane, where two vertices are adjacent if the 
+    corresponding regions share at least one point on their boundaries. 
+    This is a generalized notion of planar graphs by allowing any number 
+    of regions to meet at a single point rather than just along 
+    borders of non-zero length.
+    <P/>
+    A graph <A>D</A> is a map graph if and only if it is the 
+    <E>half-square</E> of some planar bipartite graph <M>H</M>. That is, 
+    <A>D</A> is the graph formed by one part of the bipartition of <M>H</M>, 
+    where two vertices in <A>D</A> are connected if they are at distance 2 in 
+    <M>H</M>.
+
+    Note that every planar graph is a map graph, but map graphs may contain 
+    arbitrarily large cliques making them a strictly larger class than planar graphs.
+
+<Example><![CDATA[
+gap> D := CompleteDigraph(5);
+<immutable complete digraph with 5 vertices>
+gap> IsPlanarDigraph(D);
+false
+gap> IsMapGraph(D);
+true
+gap> D := CompleteBipartiteDigraph(3, 7);;
+gap> IsMapGraph(D);
+false
+]]></Example>
+  </Description>
+</ManSection>
+<#/GAPDoc>
--- a/doc/planar.xml
+++ b/doc/planar.xml
@@ -479,12 +479,15 @@ true
     <P/>
     A graph <A>D</A> is a map graph if and only if it is the 
     <E>half-square</E> of some planar bipartite graph <M>H</M>. That is, 
-    <A>D</A> is the graph formed by one part of the bipartition of <M>H</M>, 
+    <A>D</A> is the graph formed by one part of the bipartitions of <M>H</M>, 
     where two vertices in <A>D</A> are connected if they are at distance 2 in 
-    <M>H</M>.
+    <M>H</M>.<P/>
 
     Note that every planar graph is a map graph, but map graphs may contain 
-    arbitrarily large cliques making them a strictly larger class than planar graphs.
+    arbitrarily large cliques making them a strictly larger class than planar graphs.<P/>
+
+    The runtime of this algorithm is at worst O(2^n * poly(n)) due to the exhaustive
+    clique-based backtracking search.
 
 <Example><![CDATA[
 gap> D := CompleteDigraph(5);
--- a/gap/planar.gi
+++ b/gap/planar.gi
@@ -114,7 +114,7 @@ function(D)
     return DigraphByEdges(dualEdges);
 end);
 
-# A graph is a map graph if we can find a "witness" that is planar 
+# A graph is a map graph if we can find a "witness" that is planar
 # and bipartite that the original graph is a half-square of
 
 # for all vertices v find all ways to cover its edges using a set of cliques
@@ -296,7 +296,7 @@ end);
 
 BindGlobal("DIGRAPHS_IsMapGraphSearch",
 function(G, n)
-  local allCovers, coverCounts, indices, assignment, H, 
+  local allCovers, coverCounts, indices, assignment, H,
         halfsq, i, v, targetEdges;
 
   targetEdges := Set(DigraphEdges(G));
--- a/gap/planar.gi
+++ b/gap/planar.gi
@@ -114,13 +114,14 @@ function(D)
     return DigraphByEdges(dualEdges);
 end);
 
-# A graph is a map graph if we can find a "witness" that is planar and bipartite that the original graph is a half-square of
+# A graph is a map graph if we can find a "witness" that is planar 
+# and bipartite that the original graph is a half-square of
 
 # for all vertices v find all ways to cover its edges using a set of cliques
 BindGlobal("DIGRAPHS_NbrCliqueCovers",
 function(D, v)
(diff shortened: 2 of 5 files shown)


doc/planar.xml
```diff
@@ -464,3 +464,39 @@ true
...
+    A map graph is a graph whose vertices can be represented as disjoint 
+    regions in the plane, where two vertices are adjacent if the 
+    corresponding regions share at least one point on their boundaries. 
+    This is a generalized notion of planar graphs by allowing any number 
+    of regions to meet at a single point rather than just along 
+    borders of non-zero length.
+    <P/>
+    A graph <A>D</A> is a map graph if and only if it is the 
+    <E>half-square</E> of some planar bipartite graph <M>H</M>. That is, 
+    <A>D</A> is the graph formed by one part of the bipartition of <M>H</M>,
```
james-d-mitchell:
```suggestion
    <A>D</A> is the graph formed by one part of the bipartitions of <M>H</M>, 
```

gap/planar.gi
```diff
@@ -114,6 +114,231 @@ function(D)
...
+  for v in [1 .. Length(outNeighb)] do
+    Sort(outNeighb[v]);
+  od;
+
+  return DigraphImmutableCopy(Digraph(outNeighb));
+end);
+
+# reconstruct graph from the planar bipartite graph
+# for the inputted H we have U = [1...n] and V is the rest(all junctions)
+BindGlobal("DIGRAPHS_HalfSquare",
```
james-d-mitchell, line 268: I think this should be an attribute (or maybe an operation), for a bipartite graph, that has independent documentation and tests. It should probably take which set of vertices to use (instead of `[1 .. n]`), and check if that set of vertices is one of the two sets used in the definition of the bipartite graph.

gap/planar.gi
```diff
@@ -114,6 +114,231 @@ function(D)
...
+          if v <= n and v <> u then
+            adj[u][v] := true;
+          fi;
+        od;
+      fi;
+    od;
+  od;
+
+  outgoing := List([1 .. n], u -> ListBlist([1 .. n], adj[u]));
+  return DigraphImmutableCopy(Digraph(outgoing));
```
james-d-mitchell:
```suggestion
  return Digraph(outgoing);
```
This is immutable by default and there's no sense in taking an additional copy here.

gap/planar.gi
```diff
@@ -114,6 +114,231 @@ function(D)
...
+        for v in OutNeighboursOfVertex(H, w) do
+          if v <= n and v <> u then
+            adj[u][v] := true;
+          fi;
+        od;
+      fi;
+    od;
+  od;
+
+  outgoing := List([1 .. n], u -> ListBlist([1 .. n], adj[u]));
```
james-d-mitchell: I also don't really see any advantage of using blists here, rather than just constructing `outgoing` directly. Also we'd usually refer to this as the `out-neighbours` of `H`.

gap/planar.gi
```diff
@@ -151,3 +376,31 @@ function(D)
...
+    ErrorNoReturn("expected a digraph with no multiple edges");
+  fi;
+
+  G := MaximalSymmetricSubdigraphWithoutLoops(D);
+
+  n := DigraphNrVertices(G);
+  m := DigraphNrAdjacenciesWithoutLoops(G);
+
+  if n < 5 or m < 9 then
+    return true;
```
james-d-mitchell, line 378: be good to combine these `if`s into a single statement using `elif`, and possibly providing a reference for why these conditions imply the answers they do.

doc/planar.xml
```diff
@@ -464,3 +464,39 @@ true
...
+    corresponding regions share at least one point on their boundaries. 
+    This is a generalized notion of planar graphs by allowing any number 
+    of regions to meet at a single point rather than just along 
+    borders of non-zero length.
+    <P/>
+    A graph <A>D</A> is a map graph if and only if it is the 
+    <E>half-square</E> of some planar bipartite graph <M>H</M>. That is, 
+    <A>D</A> is the graph formed by one part of the bipartition of <M>H</M>, 
+    where two vertices in <A>D</A> are connected if they are at distance 2 in 
+    <M>H</M>.
```
james-d-mitchell:
```suggestion
    <M>H</M>.<P/>
```

doc/planar.xml
```diff
@@ -464,3 +464,39 @@ true
...
+
+    Note that every planar graph is a map graph, but map graphs may contain 
+    arbitrarily large cliques making them a strictly larger class than planar graphs.
```
james-d-mitchell: Be good to indicate something about the complexity of the implementation, it looks fairly high (unavoidably) so be good to indicate this.

gap/planar.gi
```diff
@@ -114,6 +114,231 @@ function(D)
...
+        fi;
+      od;
+    od;
+  od;
+
+  for v in [1 .. Length(outNeighb)] do
+    Sort(outNeighb[v]);
+  od;
+
+  return DigraphImmutableCopy(Digraph(outNeighb));
```
james-d-mitchell: This is also an unnecessary copy

gap/planar.gi
```diff
@@ -114,6 +114,231 @@ function(D)
...
+
+# reconstruct graph from the planar bipartite graph
+# for the inputted H we have U = [1...n] and V is the rest(all junctions)
+BindGlobal("DIGRAPHS_HalfSquare",
+function(H, n)
+  local adj, u, w, v, outgoing;
+
+  adj := List([1 .. n], i -> BlistList([1 .. n], []));
+
+  for u in [1 .. n] do
```
james-d-mitchell: You could replace this whole loop with `DigraphDistanceSet`

gap/planar.gi
```diff
@@ -114,6 +114,231 @@ function(D)
...
+
+      if not IsBound(witnessMap.(key)) then
+        witnessMap.(key) := nextWitness;
+        Add(outNeighb, []);
+        nextWitness := nextWitness + 1;
+      fi;
+      w := witnessMap.(key);
+
+      for u in canon do
+        if not w in outNeighb[u] then
```
james-d-mitchell: Unless there's a reason to have the elements in `outNeighb[u]` in the order they are added here, then you'd be better to do `AddSet(outNeighb[u], w);` instead of this `if`-condition. The reason being that `w in outNeighb[u]` is a linear search if `outNeighb[u]` is not sorted (or it doesn't know its sorted), whereas `AddSet(outNeighb[u], w)` does a binary search, with log complexity.

gap/planar.gi
```diff
@@ -114,6 +114,231 @@ function(D)
...
+          Add(outNeighb[u], w);
+        fi;
+        if not u in outNeighb[w] then
```
james-d-mitchell: same coment here

gap/planar.gi
```diff
@@ -114,6 +114,231 @@ function(D)
...
+  outNeighb := List([1 .. n], i -> []);
+
+  for v in [1 .. n] do
+    for clique in covers[v] do
+      canon := ShallowCopy(clique);
+      if not v in canon then
+        Add(canon, v);
+      fi;
+      Sort(canon);
+      key := String(canon);
```
james-d-mitchell: I don't think this will work in general, how is the key "123" interpreted? It can be generated by `[1, 23]`, and `[123]`, and there's no way to tell the difference. I think you maybe want to use a `HashMap` here instead of a record.

gap/planar.gi
```diff
@@ -114,6 +114,231 @@ function(D)
...
+  fi;
+
+  maxCliques := List(DigraphMaximalCliques(sub),
+                     cl -> List(cl, idx -> nbrs[idx]));
+
+  mapping := [];
+  for i in [1 .. Length(nbrs)] do
+    mapping[nbrs[i]] := i;
+  od;
+
```
james-d-mitchell, line 178: Local functions that depend on the state of "global" variables, like `subEdges` here, smell bad, and should probably be avoided. Can you please refactor this so that `EdgesCoveredBy` is a global function and that it's arguments are explicitly given?

gap/planar.gi
```diff
@@ -114,6 +114,231 @@ function(D)
...
+  if IsEmpty(nbrs) then
+    return [[]];
+  fi;
+  sub := InducedSubdigraph(D, nbrs);
+
+  # gets the edges in the new graph
+  subEdges := [];
+  for i in DigraphVertices(sub) do
+    for e in OutNeighboursOfVertex(sub, i) do
+      if e > i then
```
james-d-mitchell, line 161: I'm struggling to understand why `e > i` is being checked here? Be good to have some more comments to understand what's being done here.

anujthak: Intent is so that we don't double count edges([1,2] and [2,1]) ordering doesn't matter but edges are undirected(as a cant neighbor b without b also neighboring a)

james-d-mitchell: Gotcha, I'd still like to understand what algorithm your implementing but we can discuss it tomorrow, looks like there's great stuff in here!

PR #925, opened 2026-04-28
CI: use configflags instead of CONFIGFLAGS

fingolfin:
See https://github.com/gap-actions/build-pkg/pull/32


james-d-mitchell:
Thanks @fingolfin

PR #928, opened 2026-05-01
grahom.gi test coverage

mtorpey:
Nothing too controversial here, but this brings test coverage of grahom to 100%.

In case anyone's wondering: yes, this is more fun than marking.

### DIFF
--- a/gap/grahom.gi
+++ b/gap/grahom.gi
@@ -521,8 +521,8 @@ InstallMethod(DigraphsRespectsColouring,
 [IsDigraph, IsDigraph, IsTransformation, IsList, IsList],
 function(src, ran, x, cols1, cols2)
   if Maximum(OnTuples(DigraphVertices(src), x)) > DigraphNrVertices(ran) then
-    ErrorNoReturn("the third argument <x> must map the vertices of the first ",
-                  "argument <src> into the vertices of the second argument ",
+    ErrorNoReturn("the 3rd argument <x> must map the vertices of the 1st ",
+                  "argument <src> into the vertices of the 2nd argument ",
                   "<ran>,");
   fi;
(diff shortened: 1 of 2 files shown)


gap/grahom.gi
```diff
@@ -521,9 +521,9 @@ InstallMethod(DigraphsRespectsColouring,
 [IsDigraph, IsDigraph, IsTransformation, IsList, IsList],
 function(src, ran, x, cols1, cols2)
   if Maximum(OnTuples(DigraphVertices(src), x)) > DigraphNrVertices(ran) then
-    ErrorNoReturn("the third argument <x> must map the vertices of the first ",
-                  "argument <src> into the vertices of the second argument ",
-                  "<ran>,");
+    ErrorNoReturn("the third argument <x> must map the vertices of the\n",
```
james-d-mitchell: I have the memory that we're not supposed to put new line characters in error messages, maybe I'm misremembering, are there other instances where we're doing this? O/w looks great to me!

mtorpey: Whoops! Looks like we don't do it anywhere else (except a recent PR from Anuj that I merged).

Are we happy for error messages to just be very long? If so I'll simply let this run on.

james-d-mitchell: I think someone once told me (maybe Markus?) that we should leave the error formatting to GAP and not do it ourselves. So yeah I think we're okay with it

mtorpey:
@james-d-mitchell – changed that error message back, hopefully all okay now.

mtorpey:
And fixed.

Curiously, this was caused by the error message wrapping over the word `<src>` so that the `>` was the first character on the second line. This means that the expected output in the `.tst` file was treated as run-on input and caused a syntax warning, even though it was generated by `rewriteToFile`!

james-d-mitchell (approved):
Nice thanks @mtorpey

PR #937, opened 2026-06-17
build: support EAPS 5

jengelh:
They renamed a few functions as far as gap-digraphs is concerned, but other than that, the EAPS API is the same.


jengelh:
The Github CI/testsuite only fails because there is that one compile test that tries to use --with-external-planarity and (ofc) an external EAPS 4.

james-d-mitchell:
Thanks @jengelh much appreciated!

PR #945, opened 2026-09-08
Build the manual from the current checkout

james-d-mitchell:
Manual generation can select another installed copy of Digraphs, leaving the checkout without the HTML files required by CI. Read the checkout's metadata and XML files and pass its directory to AutoDoc so the generated manual belongs to the checkout.

Verified a documentation build in a temporary checkout while GAP selected another installed copy by package name. The checkout's HTML, index, and version checks passed, as did `git diff --check`. The workflow and `use-latex` setting are unchanged.

Addresses the [failed manual job](https://github.com/digraphs/Digraphs/actions/runs/34203125254/job/101986306050).

AI disclosure: OpenAI Codex diagnosed the failure, implemented and verified the fix, and prepared this commit and PR. Codex is the commit author.


wilfwilson:
Decent.

